PR #104, opened 2026-04-25
fix(nip59): authenticate daemon via seal identity, not rumor sender

grunch:
The previous check compared the active Mostro pubkey against the rumor sender, which is self-asserted whenever the inner signature is absent — the common case for protocol responses. An attacker could seal a wrap with their own key and set the rumor pubkey to the configured Mostro key, bypassing the check.

Authenticate against `identity` (the seal signer, verified inside `unwrap_message`) and reject unsigned rumors whose `sender` diverges from the verified `identity`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced protocol response authentication to use more secure verification methods.
  * Added stricter validation for message integrity to reject improperly signed requests.


grunch:
@coderabbitai review

PR #164, opened 2026-06-26
fix: Mostro node selection now takes effect, persists, and refreshes the order book

### LINKED ISSUE #158: Mostro node change does nothing — pubkey does not persist, does not rehydrate, does not refresh subscriptions
**Milestone:** 1 — Bug fixes (correctness)
**Type:** bug / protocol
**Priority:** high
**Blocks:** Multi-Mostro node management (M5).

**Symptom**
In the node selector the user pastes the pubkey of the desired node and **nothing changes** — the order book continues to show data from the previous node.

**Diagnosis (verified): three defects that compound**
The selector calls `setMostroPubkey(pubkey)` (`mostro_node_selector.dart:93`) — never `setMostroNode` (which has no caller in the UI). `set_mostro_pubkey` (`settings.rs:190`) does **one thing**: it writes `static ACTIVE_MOSTRO_PUBKEY` in memory (`config.rs:42`). From there:

1. **Subscriptions are not refreshed (TODO #93 — visible symptom).** `subscribe_order_and_dm_feeds` (`relay_pool.rs:158`) reads `active_mostro_pubkey()` exactly **once** at startup and pins the filter with `.author(old_pubkey)`. Changing the pubkey does not re-run the subscription → the order book (38383) and the Mostro replies (kind 14) keep filtering by the **old** node. You see nothing from the new node.
2. **Half-changed (confusing).** Outgoing actions (create/take order, `orders.rs:490/688/815...`) **do** read `active_mostro_pubkey()` fresh → they target the new node. But the inbound side stays pinned to the old node → looks broken in a half-changed way.
3. **Does not persist + does not rehydrate.** `set_mostro_pubkey` does **not** write to DB (only `set_mostro_node` would, and nobody calls it). And **nothing at startup** loads the persisted node into `ACTIVE_MOSTRO_PUBKEY` (initialized to `None` → falls back to the compiled default; `main.dart` only calls `initDb`). Any change is lost on relaunch.

**Fix (three parts)**
- **Refresh subscriptions** when the node changes: unsubscribe the old author-pinned filters and re-run `subscribe_order_and_dm_feeds` + the orders/DM subs with the new pubkey *(TODO #93)*.
- **Persist** the selection: the selector should go through a path that writes to DB (`set_mostro_node`), or `set_mostro_pubkey` should also persist.
- **Rehydrate at startup**: in init, load the persisted active node (`get_active_mostro_node`) into `ACTIVE_MOSTRO_PUBKEY` so the change survives relaunch.

This bug must be fixed before Multi-Mostro node management (M5) becomes usable.

Catrya:
fix #158 
Selecting a Mostro node in Settings did nothing: the order book kept showing the previous node. Three compounding defects:

1. **No refresh** — the inbound subscriptions (Kind 38383 order book + Kind 14 Mostro replies) were author-pinned to thribed.
2. **Half-changed** — outgoing actions readthey targeted the new node), while the inboun.
3. **No persistence / rehydration** — the smemory override; it was never written to the DB nor loaded at startup, so any change was lost on relaunch.

## Approach

Single source of truth = the active node's **pubkey** (identity); node metadata (kind 0 / 38385) is a separate concern defe One activation entry, `set_active_mostro_node(p updates the override, and re-targets the live subscriptions. The switch is purely local (no daemon messages), matching v1 behavior.

Notable design points:
- Re-subscribe with **stable subscription IDs** so a switch replaces the old author-pinned filters in place (no leakedrt); the long-lived loop resolves the active node per event.
- nostr-sdk suppresses already-seen events ream, so returning to a previously-visited node would show an empty book. The switch therefore **refetches** the node's currenraw message channel, not subject to that dedup) and feeds them through the same ingestion path as live events.

## Out of scope (later / M5)
Relay auto-discovery (kind 10002 sync), full session recovery, and guarding a switch made mid-trade. No relay state is changed on switch.

### DIFF
--- a/lib/features/settings/widgets/mostro_node_selector.dart
+++ b/lib/features/settings/widgets/mostro_node_selector.dart
@@ -63,13 +63,13 @@ class _MostroNodeSelectorState extends ConsumerState<MostroNodeSelector> {
     final previous = ref.read(mostroPubkeyProvider);
     ref.read(mostroPubkeyProvider.notifier).state = _defaultMostroPubkey;
     try {
-      await settings_api.setMostroPubkey(pubkey: null);
+      await settings_api.setActiveMostroNode(pubkey: _defaultMostroPubkey);
       if (!mounted) return;
       _controller.clear();
       setState(() => _errorText = null);
       Navigator.of(context).pop();
     } catch (e) {
-      debugPrint('[MostroNodeSelector] setMostroPubkey(null) failed: $e');
+      debugPrint('[MostroNodeSelector] setActiveMostroNode(default) failed: $e');
       ref.read(mostroPubkeyProvider.notifier).state = previous;
       if (!mounted) return;
       setState(() => _errorText = 'Failed to reset node');
@@ -90,11 +90,11 @@ class _MostroNodeSelectorState extends ConsumerState<MostroNodeSelector> {
     final previous = ref.read(mostroPubkeyProvider);
     ref.read(mostroPubkeyProvider.notifier).state = pubkey;
     try {
-      await settings_api.setMostroPubkey(pubkey: pubkey);
+      await settings_api.setActiveMostroNode(pubkey: pubkey);
       if (!mounted) return;
       Navigator.of(context).pop();
     } catch (e) {
-      debugPrint('[MostroNodeSelector] setMostroPubkey failed: $e');
+      debugPrint('[MostroNodeSelector] setActiveMostroNode failed: $e');
       ref.read(mostroPubkeyProvider.notifier).state = previous;
       if (!mounted) return;
       setState(() => _errorText = 'Invalid pubkey or bridge error');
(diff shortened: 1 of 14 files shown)


Catrya:
@coderabbitai review

Catrya:
The entire IndexedDB backend is currently stubbed: all 29 `Storage` trait methods
(orders, trades, identity, trade keys, relays, queued messages, etc.) return
`Ok(())` / `Ok(None)` without persisting. So on web, *nothing* survives a reload
today — node selection is just one of many. The new `save/get_active_mostro_pubkey`
methods follow that exact pattern (warn + `Ok`), and the runtime warning already
surfaces the limitation.

Implementing only these two methods wouldn't make web actually work: identity,
trade keys, trades, and relays would still not persist, so the app can't function
across a web reload regardless. Making it real means implementing the IndexedDB
backend wholesale, which is well outside the scope of this node-selection bugfix.

Native (SQLite) persistence — the target platform for this change — works and is
covered by a round-trip test here. I'd prefer to track web persistence as its own
effort rather than implement it piecemeal illow-up
issue for "Implement the IndexedDB Storage  that's
useful.

Issue #196, opened 2026-07-18
About screen: display the active node's bond policy


AndreaDiazCorreia:
**Part of #145.**

## Goal
Show the connected node's bond policy on the About screen.

## Depends on
The bond-policy discovery issue in this epic (kind 38385 parsing).

## Parity reference (v1)
`MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` section 2.

## Tasks
- Render whether the node enforces bonds, on which side (`apply_to`), and the parameters (pct / base / claim window) when `enabled`.
- Handle the three states cleanly: `unsupported` (legacy daemon) vs `disabled` vs `enabled`.
- Localize across en/es/fr/de/it.

## Acceptance criteria
- The About screen reflects the parsed policy for the active node and updates on instance switch.

PR #198, opened 2026-07-18
feat: parse anti-abuse bond policy from kind-38385 info event

### LINKED ISSUE #190: Parse and validate bond policy from kind 38385 info events
**Part of #145.**

## Goal
Parse the node's anti-abuse bond policy from its kind-38385 info event and expose it to the app, so the UI can warn users up front and the About screen can display it.

## Current state (v2)
Kind 38385 is parsed today in `lib/features/about/models/mostro_instance.dart:75-112`, but it carries no bond fields.

## Parity reference (v1)
`MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` section 2. v1 model: `BondPolicy` with three states — `unsupported` (tag absent, legacy daemon), `disabled` (`bond_enabled="false"`), `enabled` (`bond_enabled="true"`).

## Tasks
- Add the three-state `BondPolicy` model (not a boolean).
- Parse and range-validate the 7 tags: `bond_enabled`, `bond_apply_to` (`take|make|both`), `bond_slash_on_waiting_timeout`, `bond_amount_pct` (`[0.0, 1.0]`), `bond_base_amount_sats` (`>= 0`), `bond_slash_node_share_pct` (`[0.0, 1.0]`), `bond_payout_claim_window_days` (`> 0`, default 15).
- Treat empty/whitespace `bond_enabled=""` as missing (`unsupported`), not `disabled`.
- Sanitize malformed tag arrays defensively: a corrupt payload must never masquerade as an intentional policy (fall back to `unsupported`); out-of-range or unparseable parameters yield `null`.

## Placement note
Per the architecture golden rule (protocol logic belongs in Rust), confirm whether this parsing should move to Rust or extend the existing Dart parser before implementing.

## Acceptance criteria
- Unit tests cover the three states, each tag's range validation, and malformed/empty inputs (mirror v1 `mostro_instance_test.dart`).
- The bond amount is never computed client-side for charging (the daemon sends the exact bolt11); pct/base are used only for up-front warnings.

AndreaDiazCorreia:
Closes #190. Part of the anti-abuse bond epic #145.

## What

Parses and validates the anti-abuse bond policy advertised by a Mostro node in its kind-38385 info event, exposing it on `MostroInstance` for the rest of the app to consume. Parsing only — no UI (the About-screen display is tracked in #196).

## Changes

- Add a three-state `BondPolicy` enum (`unsupported` / `disabled` / `enabled`) so the app can tell "feature off" apart from a legacy daemon, plus a `BondApplyTo` enum (`take` / `make` / `both`).
- Add seven bond fields to `MostroInstance` (`bondPolicy` defaults to `unsupported`; the six parameters are nullable and meaningful only when the policy is `enabled`).
- Parse the bond tags in `MostroInstance.fromTags` with defensive sanitization:
  - Empty or whitespace-only `bond_enabled=""` is treated as missing (`unsupported`), not `disabled`, preserving the three-state semantics.
  - Malformed `bond_enabled` values fall back to `unsupported`.
  - Range validation: `bond_amount_pct` and `bond_slash_node_share_pct` in `[0.0, 1.0]`, `bond_base_amount_sats` >= 0, `bond_payout_claim_window_days` > 0; out-of-range or unparseable values yield `null`.
- The bond amount is never computed client-side for charging — the daemon always sends the exact bolt11; the pct/base fields exist only for up-front UI warnings.

## Tests

- 18 unit tests covering the three policy states, case-insensitive parsing, empty / whitespace / malformed / value-less tags, per-tag parsing, and range validation.
- `flutter analyze` clean; existing non-bond parsing is unaffected.

## Parity reference

Client-side behavior mirrors v1 (`MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` section 2).

## Summary by CodeRabbit

* **New Features**
  * Added support for Mostro instance “bond” configuration.
  * Bonds now support three states: enabled, disabled, or unsupported, with an application scope (take/make/both) and configurable parameters.
* **Bug Fixes**
  * Improved tag parsing robustness: whitespace and casing are handled, malformed values are ignored, and numeric inputs are validated (including range limits where applicable).
  * Bond parameters are only populated when the bond policy is enabled.
* **Tests**
  * Added coverage for bond policy resolution, parameter parsing/validation, and ensuring existing non-bond tag behavior remains unchanged.

### DIFF
--- a/lib/features/about/models/mostro_instance.dart
+++ b/lib/features/about/models/mostro_instance.dart
@@ -1,5 +1,19 @@
 import 'package:flutter/foundation.dart';
 
+/// Anti-abuse bond policy advertised by a Mostro daemon via its kind-38385
+/// info event.
+///
+/// Three states must be distinguished:
+/// - [unsupported]: the `bond_enabled` tag is absent (legacy daemon that
+///   predates the anti-abuse bond feature) or carries a malformed value.
+/// - [disabled]: `bond_enabled="false"` — the operator left the feature off.
+/// - [enabled]: `bond_enabled="true"` — the bond is active and the remaining
+///   bond tags are meaningful.
+enum BondPolicy { unsupported, disabled, enabled }
+
+/// Which side of a trade a bond applies to (`bond_apply_to` tag).
+enum BondApplyTo { take, make, both }
+
 /// Dart model parsed from a Nostr Kind 38385 (Mostro instance status) event.
 ///
 /// All fields are derived from individual tags — the event `content` is empty
@@ -29,6 +43,13 @@ class MostroInstance {
     this.lndNetworks,
     this.lndUris,
     this.maxOrdersPerResponse,
+    this.bondPolicy = BondPolicy.unsupported,
+    this.bondApplyTo,
+    this.bondSlashOnWaitingTimeout,
+    this.bondAmountPct,
+    this.bondBaseAmountSats,
+    this.bondSlashNodeSharePct,
+    this.bondPayoutClaimWindowDays,
   });
 
   /// Mostro daemon pubkey (from `d` tag).
@@ -66,6 +87,26 @@ class MostroInstance {
   final String? lndNetworks;
   final String? lndUris;
 
+  // ── Anti-abuse bond ─────────────────────────────────────────────────────────
+
+  /// Bond policy state; defaults to [BondPolicy.unsupported]. See [BondPolicy].
+  final BondPolicy bondPolicy;
+
+  // The following six fields carry the bond parameters and are only meaningful
+  // when [bondPolicy] is [BondPolicy.enabled]. Each is `null` when the tag is
+  // absent or when its value fails validation.
+
+  final BondApplyTo? bondApplyTo;
+  final bool? bondSlashOnWaitingTimeout;
+  /// Bond fraction of the order amount, constrained to `[0.0, 1.0]`.
+  final double? bondAmountPct;
+  /// Minimum bond floor in sats, `>= 0`.
+  final int? bondBaseAmountSats;
+  /// Node's share of a slashed bond, constrained to `[0.0, 1.0]`.
+  final double? bondSlashNodeSharePct;
+  /// Days to claim a payout before forfeit, `> 0`.
+  final int? bondPayoutClaimWindowDays;
+
   // ── Factory ─────────────────────────────────────────────────────────────────
 
   /// Parse a Kind 38385 event's tag list into a [MostroInstance].
@@ -80,6 +121,73 @@ class MostroInstance {
         }
       }
       return null;
+    }
+
+    // Like [get] but trims the value and treats empty/whitespace-only as
+    // missing, so an empty tag (e.g. `bond_enabled=""`) is not misparsed as a
+    // legitimate value — preserving the three-state [BondPolicy] semantics.
+    String? getOptional(String name) {
+      final raw = get(name);
+      if (raw == null) return null;
+      final value = raw.trim();
+      return value.isEmpty ? null : value;
+    }
+
+    BondPolicy parseBondPolicy() {
+      switch (getOptional('bond_enabled')?.toLowerCase()) {
+        case 'true':
+          return BondPolicy.enabled;
+        case 'false':
+          return BondPolicy.disabled;
+        default:
+          return BondPolicy.unsupported;
+      }
+    }
+
+    BondApplyTo? parseBondApplyTo() {
+      switch (getOptional('bond_apply_to')?.toLowerCase()) {
+        case 'take':
+          return BondApplyTo.take;
+        case 'make':
+          return BondApplyTo.make;
+        case 'both':
+          return BondApplyTo.both;
+        default:
+          return null;
+      }
+    }
+
+    // Parses a boolean tag, yielding `null` for anything other than
+    // `"true"`/`"false"` so malformed data is not collapsed into a valid state.
+    bool? parseBool(String name) {
+      switch (getOptional(name)?.toLowerCase()) {
+        case 'true':
+          return true;
+        case 'false':
+          return false;
+        default:
+          return null;
+      }
+    }
+
+    // Parses a fraction constrained to `[0.0, 1.0]`; out-of-range or
+    // unparseable values yield `null`.
+    double? parseFraction(String name) {
+      final value = double.tryParse(getOptional(name) ?? '');
+      if (value == null) return null;
+      return (value >= 0.0 && value <= 1.0) ? value : null;
+    }
+
+    int? parseNonNegativeInt(String name) {
+      final value = int.tryParse(getOptional(name) ?? '');
+      if (value == null) return null;
+      return value >= 0 ? value : null;
+    }
+
+    int? parsePositiveInt(String name) {
+      final value = int.tryParse(getOptional(name) ?? '');
+      if (value == null) return null;
+      return value > 0 ? value : null;
     }
 
     return MostroInstance(
@@ -108,6 +216,14 @@ class MostroInstance {
       lndUris: get('lnd_uris'),
       maxOrdersPerResponse:
           int.tryParse(get('max_orders_per_response') ?? ''),
+      bondPolicy: parseBondPolicy(),
+      bondApplyTo: parseBondApplyTo(),
+      bondSlashOnWaitingTimeout: parseBool('bond_slash_on_waiting_timeout'),
+      bondAmountPct: parseFraction('bond_amount_pct'),
+      bondBaseAmountSats: parseNonNegativeInt('bond_base_amount_sats'),
+      bondSlashNodeSharePct: parseFraction('bond_slash_node_share_pct'),
+      bondPayoutClaimWindowDays:
+          parsePositiveInt('bond_payout_claim_window_days'),
     );
   }
 
--- a/lib/features/about/models/mostro_instance.dart
+++ b/lib/features/about/models/mostro_instance.dart
@@ -100,7 +100,8 @@ class MostroInstance {
   final BondApplyTo? bondApplyTo;
   final bool? bondSlashOnWaitingTimeout;
 
-  /// Bond fraction of the order amount, constrained to `[0.0, 1.0]`.
+  /// Bond fraction of the order amount (0.01 = 1%). Non-negative and finite;
+  /// not capped at 1.0 (the daemon does not constrain its upper bound).
   final double? bondAmountPct;
 
   /// Minimum bond floor in sats, `>= 0`.
@@ -172,10 +173,14 @@ class MostroInstance {
       }
     }
 
-    double? parseFraction(String name) {
+    // Rejects NaN/Infinity and negatives. [max] caps the upper bound only when
+    // the protocol constrains it: the daemon validates `slash_node_share_pct`
+    // to `<= 1.0`, but `amount_pct` is an unbounded fraction (0.01 = 1%).
+    double? parseFraction(String name, {double? max}) {
       final value = double.tryParse(getOptional(name) ?? '');
-      if (value == null) return null;
-      return (value >= 0.0 && value <= 1.0) ? value : null;
+      if (value == null || !value.isFinite || value < 0.0) return null;
+      if (max != null && value > max) return null;
+      return value;
     }
 
     int? parseNonNegativeInt(String name) {
@@ -229,7 +234,9 @@ class MostroInstance {
       bondBaseAmountSats:
           isEnabled ? parseNonNegativeInt('bond_base_amount_sats') : null,
       bondSlashNodeSharePct:
-          isEnabled ? parseFraction('bond_slash_node_share_pct') : null,
+          isEnabled
+              ? parseFraction('bond_slash_node_share_pct', max: 1.0)
+              : null,
       bondPayoutClaimWindowDays:
           isEnabled ? parsePositiveInt('bond_payout_claim_window_days') : null,
     );
--- a/lib/features/about/models/mostro_instance.dart
+++ b/lib/features/about/models/mostro_instance.dart
@@ -61,8 +61,10 @@ class MostroInstance {
   final String? commitHash;
   final int? maxOrderAmount;
   final int? minOrderAmount;
+
   /// Pending order lifetime in hours.
   final int? expirationHours;
+
   /// Fee as a fraction, e.g. 0.006 = 0.6 %.
   final double? fee;
   final String? fiatCurrenciesAccepted;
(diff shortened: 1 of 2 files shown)


lib/features/about/models/mostro_instance.dart
```diff
@@ -82,6 +128,73 @@ class MostroInstance {
...
+          return true;
+        case 'false':
+          return false;
+        default:
+          return null;
+      }
+    }
+
+    double? parseFraction(String name) {
+      final value = double.tryParse(getOptional(name) ?? '');
```
grunch, line 180: **MEDIUM — is `bond_amount_pct` a fraction or a percent on the wire?**

The `[0.0, 1.0]` guard encodes an assumption I can't verify from this repo. The PR description cites `MostroP2P/mobile docs/architecture/ANTI_ABUSE_BOND.md` §2, but that file isn't reachable here and nothing in this repo pins the unit.

If the daemon publishes a **percent** (`bond_amount_pct="5"` for 5%) rather than a fraction, this guard silently discards *every* real value:

```
bond_amount_pct="5"  →  5.0 > 1.0  →  null
```

The failure mode is the bad one. Per the PR description, these fields exist "only for up-front UI warnings" — so a unit mismatch means the warning never renders, the user takes an order believing no bond applies, and the daemon then charges one. Silent-null in the direction of *understating* a financial obligation.

A sanity check that a real Mostro node's published `bond_amount_pct` survives this parser would settle it. If the fraction reading is confirmed, please link the spec section in the doc comment so the next reader doesn't have to re-derive it.

Secondary point on the same guard: `1.0` is accepted as valid, i.e. a bond equal to 100% of the order amount. Harmless to parse, but if the protocol caps this lower, the ceiling could be tightened here.

lib/features/about/models/mostro_instance.dart
```diff
@@ -82,6 +128,73 @@ class MostroInstance {
...
+
+    int? parsePositiveInt(String name) {
+      final value = int.tryParse(getOptional(name) ?? '');
+      if (value == null) return null;
+      return value > 0 ? value : null;
+    }
+
+    // Parameters are gated on an enabled policy so a disabled or malformed
+    // event never exposes live bond values (consumers key off nullability).
+    final bondPolicy = parseBondPolicy();
```
grunch, line 200: **MEDIUM — `enabled` with all-`null` parameters is reachable and undocumented.**

Confirmed on the branch:

```
bond_enabled=true, no other bond tags
  → bondPolicy=enabled, bondApplyTo=null, bondAmountPct=null, … all six null
```

The field comment above states the parameters are "non-null only when `bondPolicy` is `enabled`" — true, but the converse doesn't hold, and that's the direction a consumer will actually rely on. A UI writing `if (bondPolicy == BondPolicy.enabled) showBondWarning(instance.bondAmountPct!)` compiles fine and crashes against a real node that omitted the tag.

Worse, the nullability is **lossy about why**. These two events are indistinguishable downstream:

- operator didn't publish `bond_amount_pct` at all
- operator published `bond_amount_pct="1.5"` and this parser rejected it as out of range

The first is "amount unknown, ask the daemon." The second is "this node is advertising a malformed policy" — arguably something the user should see before committing funds. Right now both render as a blank field.

Two options, either fine:

1. **Document it** — amend the field comment to state explicitly that an `enabled` policy does *not* guarantee any parameter is non-null, and that `null` means "absent or rejected." Cheapest fix, unblocks #196.
2. **Preserve the distinction** — surface rejected-value information so the About screen (#196) can distinguish "not advertised" from "advertised but invalid."

Given #196 is where this gets consumed, option 1 now plus a note on that issue seems proportionate. Whichever you pick, a test pinning the all-null-when-enabled behavior would make the contract explicit.

lib/features/about/models/mostro_instance.dart
```diff
@@ -1,5 +1,19 @@
 import 'package:flutter/foundation.dart';
 
+/// Anti-abuse bond policy advertised by a Mostro daemon via its kind-38385
+/// info event.
+///
+/// Three states must be distinguished:
+/// - [unsupported]: the `bond_enabled` tag is absent (legacy daemon that
```
grunch, line 7: **MINOR — `unsupported` carries two different meanings.**

The doc is admirably honest that this state covers both "tag absent (legacy daemon)" and "malformed value." But those warrant different UI:

- **absent** → legacy daemon, bonds genuinely don't exist, no warning needed
- **malformed** (`bond_enabled="yes"`) → a daemon that *knows about* bonds and published something this client can't read. Treating it as "no bonds here" is the optimistic reading, and it's optimistic about whether the user's funds are at risk.

Not arguing for a fourth enum case necessarily — `unsupported` already means "we don't know," and a UI can reasonably warn on it. But if #196 renders `unsupported` as "this node does not support bonds" (rather than "unknown"), the malformed case becomes an incorrect claim to the user. Worth deciding now while the enum is still cheap to change, and worth a sentence in the doc noting the UI implication.

lib/features/about/models/mostro_instance.dart
```diff
@@ -82,6 +128,73 @@ class MostroInstance {
...
+    }
+
+    double? parseFraction(String name) {
+      final value = double.tryParse(getOptional(name) ?? '');
+      if (value == null) return null;
+      return (value >= 0.0 && value <= 1.0) ? value : null;
+    }
+
+    int? parseNonNegativeInt(String name) {
+      final value = int.tryParse(getOptional(name) ?? '');
```
grunch, line 187: **MINOR — decimal-formatted integers are silently dropped.**

Confirmed on the branch:

```
bond_base_amount_sats="1000.0"  →  int.tryParse → null  →  field is null
```

A sats amount is conceptually an integer, so `"1000"` is the expected form — but JSON/serde round-trips and float-typed config on the daemon side both commonly emit `1000.0`. Losing the value entirely (rather than reading `1000`) is a strict-parsing choice that costs a user-visible field.

If tolerance is wanted, in both this helper and `parsePositiveInt`:

```dart
final raw = getOptional(name);
final value = int.tryParse(raw ?? '') ??
    double.tryParse(raw ?? '')?.toInt();
```

(`double.tryParse` handles `NaN`/`Infinity` → `.toInt()` would throw, so guard with `isFinite` if you take this route.)

If strictness is deliberate — reasonable, given these are financial parameters and a malformed one should degrade to "unknown" rather than a guess — a one-line comment saying so would stop a future contributor from "fixing" it.

grunch (commented):
## Review — `feat: parse anti-abuse bond policy from kind-38385 info event`

**Verdict: no blockers.** Clean, well-scoped parsing PR with genuinely good defensive handling. I checked out the branch and ran the suite (`flutter test test/features/about/models/mostro_instance_test.dart` → **20 passed**) and probed the parsers empirically rather than reasoning from the source alone. Findings below are mostly hardening and test-coverage; one open question about wire units is worth answering before this lands.

### What I verified by execution

| Input (with `bond_enabled=true`) | Result | Assessment |
|---|---|---|
| `bond_amount_pct="NaN"` | `null` | ✅ correct, but incidental — see comment |
| `bond_amount_pct="Infinity"` | `null` | ✅ same |
| `bond_amount_pct=" 0.05 "` | `0.05` | ✅ |
| `bond_apply_to="BOTH"` | `BondApplyTo.both` | ✅ untested though |
| `bond_base_amount_sats="1000.0"` | `null` | ⚠️ see comment |
| `bond_slash_node_share_pct="-0.5"` | `null` | ✅ untested though |
| `bond_payout_claim_window_days="9999…9"` (i64 overflow) | `null` | ✅ |
| duplicate `bond_enabled` `false` then `true` | `disabled` | ✅ first-wins, undocumented |
| `bond_enabled=true`, no parameter tags | policy `enabled`, all six `null` | ⚠️ see comment |

### Summary of findings

| # | Severity | Finding |
|---|---|---|
| 1 | **MEDIUM** | `bond_amount_pct` unit (fraction vs. percent) is unverified against the protocol spec — a wrong assumption fails silently in the direction that surprises users with funds |
| 2 | **MEDIUM** | `bondPolicy == enabled` with all-`null` parameters is a reachable, undocumented state; "absent tag" and "value rejected as out-of-range" are indistinguishable to the UI |
| 3 | **MINOR** | `unsupported` conflates "legacy daemon" with "bond-aware daemon publishing garbage" |
| 4 | **MINOR** | `parseNonNegativeInt` / `parsePositiveInt` reject decimal-formatted integers (`"1000.0"` → `null`) |
| 5 | **MINOR** | `getOptional` sanitization applies only to bond tags; the other ~20 fields still use raw `get` |
| 6 | **NIT** | Test helpers `tagsWith` / `enabledTagsWith` are public top-level, unlike `_enabledBondTags` |
| 7 | **NIT** | PR body says "18 unit tests"; the suite has 20 |
| — | **NOTE** | `@immutable` model still has no `==` / `hashCode` — pre-existing, but this PR takes it from 21 → 28 fields, and it's consumed via a Riverpod `FutureProvider` |

### Test coverage gaps

The suite is good — three-state semantics, whitespace, value-less tags, and the gating invariant are all locked in. Missing cases, roughly in priority order:

1. **`NaN` / `Infinity` rejection for both fraction fields.** This currently works only because IEEE comparisons with `NaN` are all false. A future refactor to `value.clamp(0.0, 1.0)` would silently start returning `NaN`, and nothing would fail. This is the highest-value missing test.
2. **`bond_apply_to` case-insensitivity** — the parser calls `.toLowerCase()`, but only `bond_enabled` casing is tested.
3. **Negative `bond_slash_node_share_pct`** — only the upper bound (`2.0`) is covered; the lower bound isn't.
4. **Enabled policy with bond parameter tags entirely absent** — the inverse of the two existing gating tests, and the state discussed in finding #2.
5. **Duplicate tags** — first-wins is real behavior (confirmed above) but unspecified; a relay-merged or malicious event could carry two `bond_enabled` tags.

### Architecture note (non-blocking)

Per `CLAUDE.md`'s golden rule, protocol logic belongs in Rust; `rust/src/api/nostr.rs:167` fetches the kind-38385 tags and hands Dart a raw `List<List<String>>`, so this PR correctly follows the pattern already established by `MostroInstance`. Worth flagging for the epic though: the range validation added here (`[0.0, 1.0]`, `> 0`) is protocol semantics, not UI concern. If bond values ever get consumed by Rust-side logic, the validation will need to exist in both places or move. Not something to fix in this PR.

---

_Parsing-only scope is respected, no UI leakage, existing behavior untouched, and the reformatting of the `int.tryParse` calls is `dart format` noise rather than a behavior change. Nice work on the three-state enum — the `unsupported` vs `disabled` distinction is the right call and the reasoning is documented where a reader will find it._

lib/features/about/models/mostro_instance.dart
```diff
@@ -82,6 +128,73 @@ class MostroInstance {
       return null;
     }
 
+    // Treats empty/whitespace-only as missing, so an empty `bond_enabled=""`
+    // stays `unsupported` rather than being read as `disabled`.
+    String? getOptional(String name) {
```
grunch, line 134: **MINOR — sanitization is inconsistent across the model, and partly redundant where it's applied.**

Two observations on `getOptional`:

**It's redundant for the numeric parsers.** Dart's `int.tryParse` and `double.tryParse` already tolerate surrounding whitespace — I confirmed `bond_amount_pct=" 0.05 "` parses to `0.05`. The trim only does real work for the four string-switch helpers (`parseBondPolicy`, `parseBondApplyTo`, `parseBool`), where it is genuinely necessary and correctly reasoned about in the comment above.

**It's absent everywhere else.** The other ~20 fields still call raw `get`, so `mostro_version="  "` yields a whitespace string rather than `null`, and `fiat_currencies_accepted=""` yields `""` rather than `null` — both of which a UI has to defend against separately. Since `getOptional` is strictly safer and the numeric parsers are unaffected by it, making it the single accessor for the whole factory would make the model uniformly clean:

```dart
String? get(String name) {
  for (final tag in tags) {
    if (tag.isNotEmpty && tag[0] == name) {
      final value = tag.length > 1 ? tag[1].trim() : null;
      return (value == null || value.isEmpty) ? null : value;
    }
  }
  return null;
}
```

That touches non-bond fields, so it may belong in a follow-up rather than here — your call on scope. Flagging because two accessors with near-identical names and subtly different semantics is the kind of thing that gets picked wrong later.

While here: `get` returns the **first** matching tag and ignores any duplicates (I confirmed `bond_enabled=false` followed by `bond_enabled=true` resolves to `disabled`). First-wins is a fine policy, but it's currently implicit — worth a comment, since a relay-merged or adversarially-constructed event is exactly where duplicate tags show up.

test/features/about/models/mostro_instance_test.dart
```diff
@@ -0,0 +1,290 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mostro/features/about/models/mostro_instance.dart';
+
+/// Builds the tag list for a kind-38385 event, always including the `d`
+/// (pubkey) tag and merging any extra tags passed in.
+List<List<String>> tagsWith(Map<String, String> extra) {
```
grunch: **NIT — helper visibility is inconsistent.**

`tagsWith` and `enabledTagsWith` are public top-level functions, while `_enabledBondTags` just below is correctly private. Prefixing both with `_` matches the existing convention in this file and keeps them out of scope if this test file is ever `part`-ed or imported.

The helpers themselves read well — `enabledTagsWith` delegating to `tagsWith` with a spread is a nice touch, and the doc comment explaining *why* it exists (the gating invariant) is more than most test helpers get.

test/features/about/models/mostro_instance_test.dart
```diff
@@ -0,0 +1,290 @@
...
+      expect(
+        MostroInstance.fromTags(
+          enabledTagsWith({'bond_slash_on_waiting_timeout': 'maybe'}),
+        ).bondSlashOnWaitingTimeout,
+        isNull,
+      );
+    });
+  });
+
+  group('MostroInstance.fromTags — bond parameter validation', () {
```
grunch, line 175: **Test coverage — the highest-value missing case is `NaN` / `Infinity`.**

I verified both are currently rejected:

```
bond_amount_pct="NaN"       → null  ✅
bond_amount_pct="Infinity"  → null  ✅
```

But this works *incidentally*, not by design: `double.tryParse` happily returns `double.nan`, and `NaN >= 0.0` and `NaN <= 1.0` are both false, so the range check rejects it as a side effect of IEEE-754 semantics. Nothing in the code says "we handle NaN." Rewrite `parseFraction` as `value.clamp(0.0, 1.0)` — a plausible future simplification — and `NaN` starts flowing through into `bondAmountPct` with zero test failures.

This is exactly what a regression test is for:

```dart
test('bond_amount_pct rejects NaN and Infinity', () {
  for (final bogus in ['NaN', 'Infinity', '-Infinity']) {
    expect(
      MostroInstance.fromTags(
        enabledTagsWith({'bond_amount_pct': bogus}),
      ).bondAmountPct,
      isNull,
      reason: 'rejects $bogus',
    );
  }
});
```

Three smaller gaps in this group and the one above:

- **`bond_apply_to` casing** — the parser lowercases and I confirmed `"BOTH"` → `BondApplyTo.both`, but only `bond_enabled` casing is tested. Same for `bond_slash_on_waiting_timeout`.
- **Negative `bond_slash_node_share_pct`** — only `2.0` (upper bound) is covered here, whereas `bond_amount_pct` correctly tests both directions. Asymmetric; `-0.5` → `null` is confirmed working but unlocked.
- **Duplicate tags** — first-wins is real behavior worth pinning, since a merged or malicious event is where it would matter.

The existing tests in this group are well-constructed, incidentally — testing both `0.0` and `1.0` boundaries and distinguishing `0` (valid) from `-1` (invalid) for `bond_base_amount_sats` is precisely the right instinct.

test/features/about/models/mostro_instance_test.dart
```diff
@@ -0,0 +1,290 @@
...
+      expect(
+        MostroInstance.fromTags(
+          enabledTagsWith({'bond_payout_claim_window_days': '-5'}),
+        ).bondPayoutClaimWindowDays,
+        isNull,
+      );
+    });
+  });
+
+  group('MostroInstance.fromTags — parameters gated on enabled policy', () {
```
grunch, line 246: **Test coverage — the inverse gating case is missing.**

This group correctly proves that `disabled` and `unsupported` nodes expose no parameters *even when the tags are present* — that's the security-relevant direction and it's well covered.

The complement isn't tested: an **enabled** node whose parameter tags are **absent**. I confirmed it yields policy `enabled` with all six parameters `null`, which is the ambiguous state raised on the factory. Pinning it here documents the contract at the point a consumer would look for it:

```dart
test('enabled node with no parameter tags exposes null parameters', () {
  final instance = MostroInstance.fromTags(
    tagsWith({'bond_enabled': 'true'}),
  );

  expect(instance.bondPolicy, BondPolicy.enabled);
  expectNoBondParameters(instance);
});
```

The `expectNoBondParameters` helper is already right here and reusable as-is.

PR #199, opened 2026-07-20
feat(trades): surface cancel/dispute/release as visible buttons

### LINKED ISSUE #134: Trade Detail: surface CONTACT / CANCEL / OPEN-DISPUTE as visible buttons per role + status
**Milestone:** 6 — UX / minor parity
**Type:** ux / bug
**Priority:** high

In Active and FiatSent states the destructive / safety actions (Cancel, Open dispute) are hidden inside the ⋮ overflow menu, and Contact is only reachable via the chat chip in the header. This makes it hard for users to:
- Stop a trade that's going wrong (Cancel buried).
- Escalate a disagreement (Open dispute buried).
- Reach the counterpart promptly (Contact not a primary action).

**Scope**
- Promote CONTACT, CANCEL, OPEN-DISPUTE to first-class buttons visible by role + status:
  - **Buyer · Active:** Fiat Sent (primary) · Cancel · Open dispute · Contact
  - **Buyer · FiatSent:** Cancel · Open dispute · Contact
  - **Seller · Active:** Cancel · Open dispute · Contact
  - **Seller · FiatSent:** Release (primary) · Cancel · Open dispute · Contact
- Keep cooperative-cancel UX (separate issue) layered on top once the buttons are visible.
- Use the existing button system (`MostroReactiveButton`) so loading + result states stay consistent.

Originally surfaced via T067; this is the UX half (the backend wiring is already correct — see the polling→push enhancement issue).

BraCR10:
## What
- Promotes Cancel, Open dispute, and Release (role/status-dependent) from the
  trade detail overflow menu (⋮) to first-class, always-visible buttons below
  the primary action.
- Adds the l10n labels for these actions across all 5 locales (en/es/it/fr/de).
- Reuses the existing `MostroReactiveButton` so loading/result states stay
  consistent with the primary CTA.
- Updates docs
- Creates new tests
 

## Why
Closes #134. In Active/FiatSent trades, Cancel and Open dispute were hidden in the overflow menu, making it hard for users to escalate a disputed trade or back out of a bad one quickly. 
The Contact chat chip in the header is untouched by this PR. It stays exactly where it was. Instead of four same weight buttons, the fix is a secondary action row for Cancel, Dispute, and Release only, keeping the existing chat chip for Contact. This PR implements that reduced scope.

## Behavior by role + status
| Role + status | Buttons shown |
|---|---|
| Buyer · Active | Fiat Sent (primary) · Cancel · Open dispute |
| Buyer · FiatSent | Cancel · Open dispute |
| Seller · Active | Cancel · Open dispute |
| Seller · FiatSent | Release (primary) · Cancel · Open dispute |

## Test plan
- [x] `flutter analyze` — no issues
- [x] `flutter test` — 46 tests pass, including new
      `trade_detail_screen_test.dart` (role/status button matrix) and
      `mostro_reactive_button_test.dart`
- [X] Manually verified on device

## Release Notes

* **New Features**
  * Updated Trade Detail screen action layout (primary CTAs and secondary outlined destructive actions) for cancel/dispute/release and marking fiat as sent.
  * Added an app bar “Share order” option with a coming-soon message.
  * Expanded localized trade action labels and added action-failure announcement text for accessibility.
* **Bug Fixes**
  * Trade action failures now show the correct snackbar and then propagate for consistent upstream handling; user-aborted actions no longer surface as failures.
* **Tests**
  * Added/extended widget tests for trade action behavior, share-order UI, and button cooldown/state transitions.


BraCR10:
Suggestion, not blocking this PR.

This PR adds 5 files just from running flutter gen-l10n: app_localizations.dart plus the 4 locale variants. Every future ARB change will keep touching these same generated files and adding review noise.

Worth considering the same treatment for lib/l10n/app_localizations*.dart, generated by flutter gen-l10n from the arb files. The arb source files would stay tracked as they are the actual input.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -166,7 +163,10 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
         ],
       ),
     );
-    if (confirmed != true || !mounted) return;
+    if (!mounted) return;
+    if (confirmed != true) {
+      throw StateError('Cancel declined by user');
```
grunch: **🔴 CRITICAL — declining the dialog disables the button for 4 seconds**

This fixes the false success checkmark, but trades it for a worse bug. When the user taps **Cancel order** and answers **No**:

1. `_cancelOrder` throws `StateError`.
2. `MostroReactiveButton._handlePress` catches it, calls `widget.onError?.call(e)` — which is **`null`** for every secondary-row button — and sets `_state = _ButtonState.error`.
3. `onPressed: _state == _ButtonState.idle ? _handlePress : null` → the button is **disabled for 4 full seconds**, with no visual change and no message, because the error styling was removed in this same PR.

So the user declines a dialog — a completely normal, non-error interaction — and the button silently stops responding. Identical for **Release** at line 222, in both the fiat-sent primary CTA and the disputed secondary row.

Secondary concern: using an exception for user intent is a landmine. `_cancelOrder` / `_releaseOrder` are only reached via `MostroReactiveButton` today, but any future direct or `unawaited(...)` call turns "user said No" into an unhandled async `StateError`.

**Suggested fix** — model "aborted" as an outcome, not a failure:

```dart
// mostro_reactive_button.dart
/// Thrown by an onPressed handler when the user aborts before any work
/// starts (e.g. declines a confirmation dialog). Not a failure.
class MostroActionAborted implements Exception {
  const MostroActionAborted();
}

// _handlePress
} on MostroActionAborted {
  // No success checkmark, no cooldown — straight back to idle.
  if (mounted) setState(() => _state = _ButtonState.idle);
} catch (e) {
  ...
}
```

Then throw `const MostroActionAborted()` here and at line 222. Either way, please add a test: *decline → no checkmark AND button immediately re-enabled*.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -209,14 +210,17 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
 
-  /// Confirm and release the sats (seller). Shared between the primary CTA
-  /// in the fiat-sent state and the overflow menu in the disputed state.
+  /// Shared by the fiat-sent primary CTA and the disputed secondary row.
   Future<void> _releaseOrder() async {
     final confirmed = await showReleaseConfirmationDialog(context);
-    if (confirmed != true || !mounted) return;
+    if (!mounted) return;
+    if (confirmed != true) {
+      throw StateError('Release declined by user');
```
grunch: Same blocking issue as line 168 — see that comment for the full analysis and suggested fix.

Worth calling out that this one is reachable from **two** places: the fiat-sent primary CTA (`:855`) and the disputed secondary row. Declining `showReleaseConfirmationDialog` freezes whichever button was tapped for 4s.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -542,49 +584,53 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
+      if (canDispute)
+        destructiveButton(
+          label: l10n.openDisputeButton,
+          onPressed: _openDispute,
+        ),
+    ];
+
+    return [
+      const SizedBox(height: AppSpacing.sm),
+      Row(
```
grunch, line 650: **🟠 MAJOR — this row will overflow on long locales / small screens**

Two `Expanded` `OutlinedButton`s split the available width. On a 320–360dp screen minus `AppSpacing.lg` padding, each button gets roughly 140–160dp. German `"Bestellung stornieren"` (21 chars) and `"Streitfall eröffnen"` (19 chars) at the default 14sp label will not fit — expect a `RenderFlex overflowed` banner or clipped text. French (`"Annuler la commande"`) is nearly as long.

The 6 new widget tests all run at the default 800×600 surface in English, so they structurally cannot catch this.

Please either let the label wrap:

```dart
child: Text(
  label,
  maxLines: 2,
  textAlign: TextAlign.center,
  softWrap: true,
),
```

or add a golden/layout test at 320dp with `Locale('de')`. A quick manual check on a small device in German would also be enough to dismiss this.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -209,14 +210,17 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
       ScaffoldMessenger.of(context).showSnackBar(
         SnackBar(content: Text(AppLocalizations.of(context).openDisputeFailed)),
       );
+      rethrow;
```
grunch, line 246: **🟡 MEDIUM — two competing error-reporting conventions now coexist**

After this PR:

- `markFiatSentButton` (`:840`) reports failures via the widget's `onError` callback.
- `_cancelOrder` / `_openDispute` / `_releaseOrder` show their own SnackBar and `rethrow`, passing no `onError`.

Both designs are reasonable; having both in the same `switch` is not — the next contributor has to guess which one applies. Given these handlers already need `mounted` and `l10n` in scope anyway, the SnackBar-in-handler style is the better one to standardise on. If you go that way, `onError` becomes dead API on `MostroReactiveButton` after migrating fiat-sent, and can be dropped.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -355,7 +360,7 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
           isBuyer
               ? 'Time for the seller to confirm receipt'
               : 'Time to confirm receipt and release',
-          'If something looks wrong, open a dispute from the ⋮ menu.',
+          'If something looks wrong, open a dispute using the button below.',
```
grunch: **🟡 MEDIUM — hardcoded English inside the code this PR localizes**

This line is edited by this PR (`⋮ menu` → `button below`), but stays a raw English literal while the PR's stated goal is *"Adds the l10n labels for these actions across all 5 locales"*. Non-English users will see the old `⋮` wording replaced by… still-English text.

Please add an ARB key for it alongside the six new ones.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -523,12 +533,44 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
 
-  // ── Overflow menu (collapsed secondary/destructive actions) ──────────────
+  // ── Overflow menu (share order) ───────────────────────────────────────────
 
-  Widget _buildOverflowMenu(
-      TradeStatus status, bool isBuyer, AppColors? colors) {
-    final red = colors?.destructiveRed ?? const Color(0xFFD84D4D);
+  /// Unconditional `⋮` menu — sharing an order is always a valid action,
+  /// unlike the status-gated Cancel/Dispute/Release row below.
+  Widget _buildOverflowMenu(BuildContext context) {
```
grunch: **🔵 MINOR — `context` parameter shadows `State.context`**

`_buildOverflowMenu` is a method on the `State`, so it already has `context` in scope; the parameter shadows it for no benefit. Every other `_build*` method here takes domain args only.

Also in this method: `itemBuilder: (ctx) =>` declares `ctx` and then ignores it, closing over the outer `context` instead, and `AppLocalizations.of(context)` is called twice (once in `onSelected`, once in `itemBuilder`). Suggest dropping the parameter and hoisting `l10n` once.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -523,12 +533,44 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
+        ScaffoldMessenger.of(context).showSnackBar(
+          SnackBar(
+            content: Text(l10n.comingSoonMessage),
+            duration: const Duration(seconds: 2),
+          ),
+        );
+      },
+      itemBuilder: (ctx) => [
+        PopupMenuItem(
+          value: 0,
```
grunch: **🔵 MINOR — magic value in a typed popup menu**

`PopupMenuButton<int>` with `value: 0` and `onSelected: (_)` works with a single item, but it discards the type safety the deleted `_MenuAction` enum provided. When a second item lands, `onSelected` ignoring its argument becomes a silent bug.

```dart
enum _OverflowAction { shareOrder }
```

matches the pattern this PR just removed and costs three lines.

lib/shared/widgets/mostro_reactive_button.dart
```diff
@@ -93,13 +111,7 @@ class _MostroReactiveButtonState extends State<MostroReactiveButton> {
...
           child: const Icon(Icons.check, size: 20),
         );
-      case _ButtonState.error:
-        return Semantics(
-          label: 'Error',
-          liveRegion: true,
-          child: const Icon(Icons.error_outline, size: 20),
-        );
-      case _ButtonState.idle:
+      case _ButtonState.idle || _ButtonState.error:
```
grunch: **🟡 MEDIUM — accessibility regression + misleading enum name**

Two things here:

1. This removes `Semantics(label: 'Error', liveRegion: true, child: Icon(Icons.error_outline))`. The failure signal is now SnackBar-only — the button itself announces nothing and just goes silently unresponsive for 4s. If the visual error state is intentionally gone, consider keeping a `Semantics(liveRegion: true)` announcement so the state change stays perceivable to screen-reader users. (Note the old label was hardcoded English, so restoring it needs an ARB key.)

2. `_ButtonState.error` now renders identically to `idle` — the value no longer means "error", it means "cooling down after a failure". Renaming it to `cooldown` would make `case _ButtonState.idle || _ButtonState.cooldown:` self-explanatory instead of looking like a bug.

lib/shared/widgets/mostro_reactive_button.dart
```diff
@@ -47,7 +49,7 @@ class _MostroReactiveButtonState extends State<MostroReactiveButton> {
       if (!mounted) return;
       setState(() => _state = _ButtonState.error);
 
-      await Future.delayed(const Duration(milliseconds: 2000));
+      await Future.delayed(const Duration(seconds: 4));
```
grunch: **🔵 MINOR — magic duration**

The commit message explains this is 4s to match the SnackBar's default duration, but that rationale is lost in the code. Please extract it:

```dart
/// Matches SnackBar's default display duration, so the button re-enables
/// as the failure message disappears.
static const _kErrorCooldown = Duration(seconds: 4);
```

Also note the success path just above still uses a bare `Duration(milliseconds: 1500)` — worth giving the same treatment for symmetry.

lib/l10n/app_en.arb
```diff
@@ -445,6 +445,18 @@
   "@fiatSentFailed": {"description": "Snackbar shown when the fiat-sent action fails"},
   "releaseFailed": "Failed to release. Please try again.",
   "@releaseFailed": {"description": "Snackbar shown when the release-sats action fails"},
+  "cancelTradeButton": "Cancel order",
```
grunch: **🟡 MEDIUM — duplicate key with diverging translations**

`cancelTradeButton` (this line) and the existing `cancelOrderButton` (`:474`) both resolve to **"Cancel order"** in English, but diverge in the other locales:

| key | en | de | fr |
|---|---|---|---|
| `cancelTradeButton` (new) | Cancel order | Bestellung stornieren | Annuler la commande |
| `cancelOrderButton` (existing) | Cancel order | Angebot stornieren | — |

So a German user sees two different labels for what reads as the same action in two screens. Please either reuse `cancelOrderButton`, or make the translations agree.

Minor naming point on the same line: the key is `cancelTradeButton` but the value says "order" — the surrounding keys (`cancelTradeDialogTitle`: *"Cancel trade?"*) use "trade". Worth picking one noun for the whole flow.

grunch (changes requested):
## Summary

The layout change itself is good — promoting Cancel/Dispute/Release out of the `⋮` menu is the right call for #134, the spec/tasks updates are thorough, and the new widget tests are well-structured (the comments explaining why `pumpAndSettle()` is unsafe against the countdown `Timer.periodic` are genuinely useful).

However, the last commit (`be9a31d`, *"throw when cancel or release confirmation is declined"*) trades one bug for a worse one, and it ships with no test coverage. Requesting changes on that plus a layout risk.

### Blocking

1. **Declining a confirmation dialog leaves the button dead for 4s.** Throwing `StateError` for "user said No" routes a normal interaction through the failure path, which disables the button for the full error cooldown with no visual feedback. Details inline at `trade_detail_screen.dart:168`.
2. **Secondary row will overflow on long locales / small screens.** Two `Expanded` buttons with German/French labels won't fit at ~140dp each. Details inline at `trade_detail_screen.dart:625`.

### Non-blocking findings

| # | Severity | Finding |
|---|---|---|
| 3 | MEDIUM | Duplicate l10n key: `cancelTradeButton` and `cancelOrderButton` both mean "Cancel order" in `en` but diverge in `de` (`Bestellung stornieren` vs `Angebot stornieren`) |
| 4 | MEDIUM | Two competing error-reporting conventions now coexist in the same `switch` (`onError` vs SnackBar-in-handler + `rethrow`) |
| 5 | MEDIUM | Accessibility regression: the `Semantics(label: 'Error', liveRegion: true)` announcement was removed |
| 6 | MEDIUM | Test gaps — see below |
| 7 | MINOR | `label: 'View dispute'` (`trade_detail_screen.dart:861`) and `label: 'Pay hold invoice'` (`:826`) are still hardcoded English, sitting directly beside the labels this PR just localized. Outside the diff hunks so I couldn't anchor inline, but they read as an oversight given the PR's stated l10n goal |
| 8 | MINOR | Scope creep: restoring a `⋮` menu whose only entry is a non-functional "Share order" adds always-visible dead UI plus 5 translations for an unimplemented contract. Consider splitting it out or gating it |
| 9 | MINOR | Undocumented visual change: the release CTA lost `icon: Icons.lock_open`. Intentional? Not mentioned in the PR body or the spec updates |
| 10 | MINOR | Locale register: the screen title is fr `DÉTAILS DE L'ORDRE` but the new buttons use `commande`. The fr file already mixes *ordre* (trading) and *commande* (retail); new keys should follow *ordre*, matching `navOrderBook` / `orderDetailsTitle` |

### Test coverage

- **The entire last commit is untested.** No test covers "decline the dialog" for either cancel or release — which is exactly where the blocking bug lives.
- **No test covers the three `rethrow`s** added to `_cancelOrder` / `_openDispute` / `_releaseOrder`.
- `trade_detail_screen_test.dart:227` invokes `popupButton.onSelected!(0)` directly. The reasoning is documented, but it means the `PopupMenuItem(value: 0)` → `onSelected` wiring is never exercised; a wrong `value` would still pass. `tester.tap(find.text('Share order'))` followed by `await tester.pump(const Duration(seconds: 1))` is usually stable enough.
- No layout/locale test — the 6 new tests all run at the default 800×600 surface in English, so they structurally cannot catch finding #2.

Nice work overall — the gating matrix and the role/status test table are solid. Happy to re-review once the decline path is reworked.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Fixed it properly: added a `MostroActionAborted` marker exception that `MostroReactiveButton` catches separately and returns straight to idle, no checkmark, no cooldown. Added a test for it too, decline shows no checkmark and the button re-enables immediately.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Same fix covers this one, `MostroActionAborted` applies to both cancel and release now.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Wrapped the label instead (maxLines: 2, softWrap: true) so it can never force a single line. Added a layout test at 320dp with Locale('de') to actually prove it stops overflowing.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Agreed, standardized on the SnackBar plus rethrow pattern. Moved markFiatSentButton off onError into its own _markFiatSent method, same shape as cancel, dispute, and release now.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Both wired to l10n now. View dispute reuses a key that already existed in all 5 locales but was never actually used (viewDisputeButtonLabel), just fixed the casing to match what was shown. Pay hold invoice is a new key across all 5 locales.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Dropped the parameter, it uses the State's own context now. Also hoisted l10n once instead of calling AppLocalizations.of(context) twice.

lib/features/trades/screens/trade_detail_screen.dart
BraCR10: Brought back a typed enum, _OverflowAction, instead of the bare int. Matches the _MenuAction pattern this PR had removed.

lib/shared/widgets/mostro_reactive_button.dart
BraCR10: Both fixed. Added a Semantics(liveRegion: true) announcement scoped to the cooldown state, localized. Renamed _ButtonState.error to cooldown too, you're right that it renders the same as idle now and the old name was misleading.

lib/shared/widgets/mostro_reactive_button.dart
BraCR10: Extracted both the success display duration and the error cooldown into named constants, kept your comment about matching the SnackBar's duration on the cooldown one.

lib/l10n/app_en.arb
BraCR10: Reworded cancelTradeButton to Cancel trade across all 5 locales instead of reusing cancelOrderButton. They're actually different flows, cancelling an order listing versus cancelling an active trade, so I didn't want to blur that distinction. Matches cancelTradeDialogTitle's existing wording. Also caught the same issue on shareOrderButton in French, it was using the retail register (commande) instead of ordre like the rest of the screen.

BraCR10:
Addressed everything from the review. @grunch 

Both blocking issues are fixed, the decline path no longer freezes the button  and the secondary row wraps instead of overflowing on long locales.

All non-blocking findings are resolved too.

Ready for another look whenever you have time.

AndreaDiazCorreia (commented):
Hi @BraCR10, could you resolve the merge conflicts? I think several of them are probably related to the i18n changes

BraCR10:
> Hi @BraCR10, could you resolve the merge conflicts? I think several of them are probably related to the i18n changes

HI, Yes I'll resolve them ASAP

AndreaDiazCorreia (approved):
tACK

grunch (approved):

PR #209, opened 2026-07-23
fix(web): convert PlatformInt64 fields before using them as int

grunch:
## What

`flutter build web` failed to compile. This fixes the four failing sites and removes the
scattered duplication that caused the bug.

## Why

Rust `u64`/`i64` fields cross the flutter_rust_bridge boundary as `PlatformInt64`: a plain
`int` on native, but a `BigInt` on web (dart2js cannot hold a 64-bit integer in `int`). Four
sites used these values where an `int` is expected, so the code compiled on native and broke
on web:

```
log_report_screen.dart:127  The argument type 'BigInt' can't be assigned to the parameter type 'int'.
log_report_screen.dart:209  The argument type 'BigInt' can't be assigned to the parameter type 'int'.
trade_detail_screen.dart:118 The operator '-' isn't defined for the type 'Object'.
trade_detail_screen.dart:235 The argument type 'Object' can't be assigned to the parameter type 'int'.
```

Two of them already tried to guard for web with `raw is BigInt ? raw.toInt() : raw`, but the
`else` branch is not cast. The ternary's static type is then the `int`/`BigInt` least upper
bound — `Object` — which is precisely the `Object` in the errors above.

## How

- New `lib/shared/utils/platform_int64.dart` with `platformInt64ToInt(dynamic)` and
  `intToPlatformInt64(int)`, each documenting the native-vs-web split. `platformInt64ToInt`
  casts the `else` branch so the result is always `int`.
- Route the four failing sites through `platformInt64ToInt`.
- Replace three pre-existing private copies of the same conversion
  (`home_order_providers`, `trades_providers`, `identity_service`) with the shared helpers, so
  the logic lives in one place and the bug cannot recur in an un-fixed copy.

## Not in scope

The `flutter build web` output also lists "Wasm dry run findings" (`dart:html` in
`flutter_secure_storage_web`, `flutter_keyboard_visibility_web`, and an FRB web util). Those are
informational warnings for a `--wasm` build, not part of this failure — the standard dart2js/JS
build is what was broken, and it now compiles. Left for a separate discussion if wasm output is
wanted.

## Test plan

- [x] `flutter build web` → `✓ Built build/web`
- [x] `flutter analyze` → No issues found
- [x] `flutter test` → 121 pass (120 existing + new helper unit test covering int and BigInt inputs)
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when handling timestamps across web and native platforms.
  * Ensured timestamps display and share correctly in logs, orders, trades, identity details, and disputes.

* **Tests**
  * Added coverage for platform-specific timestamp conversion, including integer and large-integer values.

### DIFF
--- a/lib/core/services/identity_service.dart
+++ b/lib/core/services/identity_service.dart
@@ -1,13 +1,8 @@
 import 'package:flutter/foundation.dart';
 import 'package:flutter_secure_storage/flutter_secure_storage.dart';
 
+import 'package:mostro/shared/utils/platform_int64.dart';
 import 'package:mostro/src/rust/api/identity.dart' as identity_api;
-
-/// Returns [v] as the correct `PlatformInt64` type for the current platform:
-/// `int` on native, `BigInt` on web (dart2js). Returning `dynamic` lets
-/// callers pass it directly to bridge parameters typed as `PlatformInt64?`
-/// without a static-type error on either platform.
-dynamic _toInt64(int v) => kIsWeb ? BigInt.from(v) : v;
 
 /// Secure-storage keys.
 const _kMnemonic = 'mostro_identity_mnemonic';
@@ -185,7 +180,7 @@ class IdentityService {
       words: words,
       tradeKeyIndex: tradeKeyIndex,
       privacyMode: privacyMode,
-      createdAt: createdAt > 0 ? _toInt64(createdAt ~/ 1000) : null,
+      createdAt: createdAt > 0 ? intToPlatformInt64(createdAt ~/ 1000) : null,
     );
 
     debugPrint('[identity] identity loaded — pubkey=${info.publicKey}');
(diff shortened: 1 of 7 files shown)


grunch:
@codex review

grunch:
Addressed the review:

- **Trailing commas on `DateTime.fromMillisecondsSinceEpoch` (home_order_providers.dart)** — done in 05095c7. Applied by hand rather than `dart format`, which under the newer "tall" formatter (Dart 3.11) reflows the entire file (65 insertions/56 deletions of unrelated churn); CI does not enforce formatting, so the manual trailing commas keep the diff minimal.
- **Chrome-targeted test for `intToPlatformInt64` (platform_int64_test.dart)** — skipped. The Flutter CI job runs `flutter test` on the VM, not `--platform chrome`, so a browser-only test would not execute in CI (false confidence) and would fail for contributors without Chrome. The direction that caused the original build break — `BigInt → int` via `platformInt64ToInt` — is already covered by the VM test.

CI green (Flutter + Rust).

PR #211, opened 2026-07-23
fix(web): route wall-clock time through rt so wasm does not panic

### LINKED ISSUE #210: Web: SystemTime::now() panics on wasm — client shows a blank page
## Summary

The app cannot run on web. After the Rust core is compiled to WASM
(`flutter_rust_bridge_codegen build-web`) and served with the required COOP/COEP
headers, the page loads but stays blank: the WASM aborts on the first call that
reads wall-clock time.

```
panicked at .../std/src/sys/time/unsupported.rs:35:9:
time not implemented on this platform

<std::time::SystemTime>::now
rust::api::identity::unix_now
rust::api::identity::create_identity
```

## Root cause

`std::time::SystemTime::now()` is **not implemented on `wasm32-unknown-unknown`** and
panics at runtime. `rust/src/rt.rs` already abstracts the monotonic clock and timers
per platform (`Instant`, `sleep`, `timeout` via `wasmtimer` on wasm), but wall-clock
time was never abstracted — call sites use `std::time::SystemTime::now()` directly.

Identity creation runs on first launch, so the very first thing the app does on web
panics, which is why the screen is blank.

This is invisible to CI: `cargo check --target wasm32-unknown-unknown` (ci.yml)
type-checks the wasm build but never runs it, and the panic is at runtime.

## Affected call sites

Each has its own private `unix_now()` (or an inline `SystemTime::now()`), all with the
same `SystemTime::now().duration_since(UNIX_EPOCH)` shape:

| File | Line |
| --- | --- |
| `rust/src/api/identity.rs` | 368–371 |
| `rust/src/api/disputes.rs` | 103–105 |
| `rust/src/api/reputation.rs` | 112–114 |
| `rust/src/api/orders.rs` | 667 |
| `rust/src/api/logging.rs` | 113–114 |
| `rust/src/mostro/session.rs` | 80, 152 |
| `rust/src/nwc/client.rs` | 349–351 |
| `rust/src/queue/outbox.rs` | 174–176 |

(`Duration` and `Instant` uses are fine — those are already routed through
`crate::rt::time` or are platform-safe.)

## Proposed fix

`rust/src/rt.rs` is the right home; it is missing the wall-clock half of its own clock
abstraction.

1. Add a wasm-safe wall clock to `crate::rt::time`, resolved per platform exactly like
   the existing `Instant`:
   - native: `std::time::SystemTime::now()`
   - wasm: a browser-backed clock. The crate already depends on `wasmtimer`, which
     exposes `wasmtimer::std::SystemTime` — to be confirmed during implementation;
     otherwise `web-time` is the de-facto standard drop-in.
   Expose a single `pub fn unix_now() -> i64` so callers do not re-implement the
   `duration_since(UNIX_EPOCH)` dance.
2. Replace the ~8 private `unix_now()` copies (and the inline `SystemTime::now()` sites)
   with `crate::rt::time::unix_now()`. This also de-duplicates logic that is currently
   copy-pasted eight times.

## Preventing regressions

CI only type-checks wasm; it never executes it, so a runtime panic like this slips
through. Consider a follow-up: a headless wasm smoke test (e.g. `wasm-pack test
--headless`) that exercises identity creation, or at minimum a lint/grep guard that
fails when `std::time::SystemTime::now()` appears outside `rt.rs`.

## Acceptance criteria

- [ ] `crate::rt::time::unix_now()` (or equivalent) returns correct Unix time on native
      and on wasm without panicking.
- [ ] No `std::time::SystemTime::now()` remains outside `rt.rs`.
- [ ] The app runs on web: `build-web` + `flutter run -d chrome` (with COOP/COEP headers)
      reaches the UI and creates an identity without a WASM abort.
- [ ] Native behavior unchanged; `cargo test` and `cargo clippy` clean.

## Context

Found while trying to run the client on web. The web run flow itself
(`build-web` → `flutter run -d chrome` with COOP/COEP headers, plus the wasm-pack /
nightly+rust-src prerequisites) is also undocumented; that is being tracked separately
as a docs change.

grunch:
Closes #210

## What

The web client rendered a blank page. The Rust core panicked at runtime the moment it
read wall-clock time, which happens on first launch during identity creation:

```
panicked at .../std/src/sys/time/unsupported.rs:35:9:
time not implemented on this platform
  <std::time::SystemTime>::now
  rust::api::identity::unix_now
  rust::api::identity::create_identity
```

## Why

`std::time::SystemTime::now()` is unimplemented on `wasm32-unknown-unknown` and aborts.
`rust/src/rt.rs` already abstracts the monotonic clock and timers per platform
(`Instant`, `sleep`, `timeout` via `wasmtimer` on wasm) — but the wall clock was never
abstracted, and 12 call sites read `SystemTime::now()` directly.

CI never caught it: `cargo check --target wasm32` type-checks the wasm build but never
runs it, and the panic is at runtime.

## How

- **`rt.rs`**: add `SystemTime`/`UNIX_EPOCH` to `crate::rt::time`, resolved per platform
  (native `std`, wasm `wasmtimer::std` — already a dependency, no new crate), plus a
  single `rt::unix_now() -> i64`.
- **12 call sites** routed through it:
  - seven private `unix_now()` copies (`identity`, `disputes`, `reputation`, `messages`,
    `nwc`, `outbox`, `relay_pool`) now `use crate::rt::unix_now`;
  - five inline reads (`orders` ×2, `logging`, `session` ×2, `blossom`) call
    `crate::rt::unix_now()` directly.
- Removes eight duplicated copies of the same `duration_since(UNIX_EPOCH)` logic.

No `std::time::SystemTime::now()` remains outside `rt.rs`.

## Docs

Adds a **Running on Web** section to the README — the flow was entirely undocumented and
is what made this hard to hit: nightly + `rust-src` prerequisites, `build-web` to compile
the Rust core to WASM, `flutter run -d chrome` with COOP/COEP headers, the exact console
error each missing step produces, and a note that production hosting needs the same
headers.

## Test plan

- [x] `cargo test` — 110 pass
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo check --target wasm32-unknown-unknown` — clean
- [x] `flutter_rust_bridge_codegen build-web` — rebuilds `web/pkg` successfully
- [x] Web reaches the UI: `build-web` + `flutter run -d chrome` (COOP/COEP) no longer
      aborts on identity creation — blank page resolved
- [ ] CI green

## Follow-up (out of scope, noted in the issue)

CI only type-checks wasm; a runtime panic like this slips through. A headless
`wasm-pack test` smoke test (or a grep guard against `SystemTime::now()` outside `rt.rs`)
would prevent regressions.

## Summary by CodeRabbit

* **Documentation**
  * Expanded web setup instructions with platform-specific run commands.
  * Added guidance for building Rust components to WebAssembly and configuring cross-origin isolation for development and production.

* **Improvements**
  * Improved timestamp consistency across messaging, orders, sessions, logging, identity, reputation, networking, and uploads.
  * Added more reliable time support for web builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,13 +310,14 @@ compiles the Rust core for you; on web it does **not** — you compile the Rust 
 WebAssembly yourself, and the page must be served with cross-origin isolation headers.
 Skip either step and the app builds but shows a **blank page**.
 
-**One-time prerequisites** — in addition to the table above. `build-web` compiles the
-Rust core with `-Z build-std`, which requires a nightly toolchain with `rust-src`:
+**One-time prerequisites** — in addition to the table above. The wasm build compiles
+the Rust core with `-Z build-std`, which requires a nightly toolchain with `rust-src`
+(and the wasm target on that same nightly toolchain):
 
 ```bash
 rustup toolchain install nightly
 rustup component add rust-src --toolchain nightly
-rustup target add wasm32-unknown-unknown
+rustup target add wasm32-unknown-unknown --toolchain nightly
 cargo install wasm-pack
 ```
 
@@ -324,11 +325,17 @@ cargo install wasm-pack
 `rust_bg.wasm`). Re-run it after any change under `rust/src/`:
 
 ```bash
-flutter_rust_bridge_codegen build-web
+./scripts/build-web.sh              # add --release for an optimized build
 ```
 
-If you skip this, the console shows `Refused to execute script from '.../pkg/rust.js'
-because its MIME type ('text/html') is not executable` — the WASM was never built.
+Do **not** run `flutter_rust_bridge_codegen build-web` directly: on current nightly it
+silently produces WASM without shared memory, which crashes at runtime with
+`DataCloneError: ... #<Memory> could not be cloned` when FRB spawns its worker pool.
+The script adds the required linker flags and verifies the output (see issue #212).
+
+If you skip this step entirely, the console shows `Refused to execute script from
+'.../pkg/rust.js' because its MIME type ('text/html') is not executable` — the WASM was
+never built.
 
 **2. Run in Chrome with cross-origin isolation headers.** The Rust core uses
 `SharedArrayBuffer`, which browsers enable only under COOP/COEP:
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ Make sure the following tools are installed on your system:
 | Rust toolchain | 1.94+ stable | `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \| sh` |
 | WASM target | — | `rustup target add wasm32-unknown-unknown` |
 | wasm-pack | latest | `cargo install wasm-pack` |
+| Rust nightly + `rust-src` (web only) | — | `rustup toolchain install nightly && rustup component add rust-src --toolchain nightly` |
 | flutter_rust_bridge CLI | 2.11.1 | `cargo install flutter_rust_bridge_codegen --version 2.11.1 --locked` |
 | Xcode (macOS / iOS only) | 15+ | Mac App Store |
 | Android Studio / NDK (Android only) | latest | [developer.android.com](https://developer.android.com/studio) |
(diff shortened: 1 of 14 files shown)


grunch:
@coderabbitai review

grunch:
@codex review

grunch:
Addressed the CodeRabbit finding plus one addition, both in 5cdf71f:

- **`rustup target add wasm32-unknown-unknown` now uses `--toolchain nightly`** in the Running on Web prerequisites — valid finding: the wasm build runs under `RUSTUP_TOOLCHAIN=nightly`, so the target must live on that toolchain. (The plain `rustup target add` in the general prerequisites table is left as-is on purpose: the stable toolchain needs the target for `cargo check --target wasm32-unknown-unknown` in CI and the pre-commit hook.)

- **New `scripts/build-web.sh`**, now referenced by the README instead of the bare `flutter_rust_bridge_codegen build-web`. During verification of this PR we found that the bare command produces wasm **without shared memory** on current nightly (rustc no longer auto-adds the shared-memory linker flags for `+atomics`), which loads fine and then crashes FRB's worker pool at runtime (`DataCloneError: #<Memory> could not be cloned`). The script pins the working linker flags and parses the emitted wasm to verify the memory is actually shared, failing loudly on regression. Full analysis in #212, which will reuse this script for the GitHub Pages deploy.

Verified locally: the app now boots on Chrome end to end (identity creation, relay connection, order book).

Issue #191, opened 2026-07-18
Phase 5: maker anti-abuse bond on order creation


AndreaDiazCorreia:
**Part of #145.**

## Goal
When `apply_to in {make, both}`, require the maker to pay a bond before the order is published to Nostr.

## Parity reference (v1)
`MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` section 5; v1 PR #608.

## Design (v1) — key points to preserve
- The maker side reuses `pay-bond-invoice` (no dedicated `waiting-maker-bond` status on the client).
- The uncommitted order stays ephemeral: its session lives in memory only (a `bondPending`-style flag), so an abandoned order never persists and never appears in My Trades (no published NIP-33 event).
- Cancel abandons locally (the daemon rejects an explicit `cancel` during `WaitingMakerBond`); the hold invoice expires server-side.
- The create timeout must be disarmed when the bond invoice arrives, and the notifier must stay alive so the post-bond `new-order` ack still confirms.

## Tasks
- Handle `pay-bond-invoice` on the create request path; show the maker-specific pay-bond screen (copy warns the user to keep the screen open and pay, or the order is not created).
- Implement the ephemeral in-memory session and skip persistence while bond-pending; persist only on the confirming ack.
- Implement local cancel/abandon for the maker bond.

## Notes
Depends conceptually on the shared pay-bond handling. Current v2 maps `WaitingMakerBond` to `None` (`rust/src/api/orders.rs:2069-2070`).

## Acceptance criteria
- An abandoned maker bond leaves nothing on disk and never reappears after restart.
- Paying the bond publishes the order and confirms via the create ack.
- Tests mirror v1 `order_state_maker_bond_test.dart` and `maker_bond_timeout_test.dart`.

PR #213, opened 2026-07-23
feat(bond): recognize pay-bond-invoice and add taker bond payment flow

### LINKED ISSUE #208: Phase 1 (taker): recognize pay-bond-invoice and retire the BondRequired rejection
**Part of #145.**

## Goal
Enable taking orders on a bond-requiring node. When a taker takes such an order, the daemon replies with `pay-bond-invoice` (a Lightning hold invoice for the taker's anti-abuse bond) instead of progressing the trade. Today the client rejects that reply with a stable `BondRequired` marker; this issue replaces the rejection with the real taker bond flow. This is the entry point for the taker experience and a prerequisite for exercising the payout (#192/#193) and slash (#194, done) flows end-to-end against a take.

## Parity reference (v1)
- v1 PR #592 (taker pay-bond flow).
- `MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` (Flow 1).
- Design mocks: PR #183 (`docs/design/145-*.png`).

## Current state (verified)
- `classify_take_reply` maps `Action::PayBondInvoice` to a stable `BondRequired` rejection (`rust/src/api/orders.rs:267-271`), unit-tested at `rust/src/api/orders.rs:3187-3190`.
- Dart surfaces that marker as `l10n.bondRequired` (`lib/features/order/screens/take_order_screen.dart:166`).
- `WaitingTakerBond | WaitingMakerBond` have no local `OrderStatus` and return `None` (`rust/src/api/orders.rs:2124`).
- `mostro-core 0.13.1` already ships `Action::PayBondInvoice`, `WaitingTakerBond`/`WaitingMakerBond`; no crate upgrade needed.

## Tasks
- Retire the `BondRequired` rejection branch in `classify_take_reply`; update/replace its unit test.
- Recognize `Action::PayBondInvoice` and its bond hold-invoice payload (confirm the exact payload type against `mostro-core 0.13.1`), returning a take reply that carries the bond bolt11 to Dart.
- Add a local `OrderStatus::WaitingTakerBond` representation (remove the `None` mapping for it) so a taken order can sit in the waiting-bond state; keep the no-wildcard match exhaustive.
- Surface the bond hold invoice to the taker for payment (pay-from-wallet / external), following the seller hold-invoice UX pattern.
- On bond payment, let the trade progress into the normal take flow.
- Localize any new user-facing strings in all 5 locales.

## Acceptance criteria
- Taking an order on a bond-requiring node no longer shows a `BondRequired` error; the taker is shown the bond hold invoice to pay.
- After the taker pays the bond, the order advances to the normal take flow and reflects `WaitingTakerBond` while pending.
- `cargo test` / `clippy` and `flutter analyze` / `test` green.

## Notes
- Maker-side bond on order creation is tracked separately in #191; this issue is taker-only.

AndreaDiazCorreia:
Closes #208. Part of #145 (anti-abuse bond).

## Summary

Phase 1 (taker) of the anti-abuse bond feature. When a taker takes an order on a bond-requiring node, the daemon replies with `pay-bond-invoice` (a Lightning hold invoice for the taker's anti-abuse bond) instead of progressing the trade. Previously the client rejected that reply with a stable `BondRequired` marker; this PR replaces the rejection with the real taker bond flow: the bond hold invoice is shown to the taker, and once it is paid the trade advances into the normal take flow.

## What changed

**Rust (`rust/src/api/`)**
- Retired the `BondRequired` rejection branch in `classify_take_reply`.
- Added `OrderStatus::WaitingTakerBond` and mapped it in `map_core_status` / `status_for_action` (maker-side `WaitingMakerBond` stays unmapped — that is #191).
- Added a dedicated `bond_invoice` field to `TradeInfo` and `DaemonReply::TakeAccepted`, kept distinct from `hold_invoice` (see "Verified against the daemon" below).
- Added an `Action::AddInvoice` dispatcher arm so the buyer's trade advances correctly after the bond is paid.

**Dart (`lib/`)**
- New `PayBondInvoiceScreen` (route `/pay_bond_invoice/:orderId`) reusing the QR / NWC / external-wallet payment pattern of the seller hold-invoice screen. It watches the order status and forwards the taker to the next step once the bond is paid: add-invoice (buyer), pay-invoice (seller), or trade detail.
- `take_order_screen` routes to the bond screen when the take reply comes back as `WaitingTakerBond`, and the `BondRequired` error branch was removed.
- Added `tradeBondInfoProvider` and the `waitingTakerBond` cases to the exhaustive `OrderStatus` switches.

**l10n**: new bond-screen strings across all 5 locales; retired `bondRequired`.

## Verified against the daemon

Cross-checked against `mostro` (daemon), `mostro-cli`, and v1 `mobile`, which surfaced two issues fixed here:

1. The daemon stamps the `pay-bond-invoice` payload's `SmallOrder` with the wire-mapped `Pending` status (`WaitingTakerBond` is never exposed on the NIP-69 wire), so `classify_take_reply` now derives the bond status from the action, not the payload. Without this the client would never route to the bond screen.
2. On bond payment the daemon both republishes the Kind 38383 event to `WaitingPayment` and sends the `pay-invoice` gift wrap. Reusing the `hold_invoice` slot for the bond raced these two: the pay-invoice screen could latch the stale bond bolt11. The dedicated `bond_invoice` field removes the collision (`hold_invoice` stays null until the escrow arrives).

## Testing

- `cargo test` (110 passing) and `cargo clippy` clean.
- `flutter analyze` clean and `flutter test` (120 passing).
- `flutter_rust_bridge` bindings regenerated and verified in sync.

## Out of scope / follow-ups

- Cancelling from the bond screen currently pops (consistent with the seller hold-invoice screen). v1 sends a cancel to release the bond; replicating that release path is a candidate follow-up.
- Re-entering the bond screen from My Trades after popping is not wired yet.
- Maker-side bond on order creation is tracked separately in #191.

## Summary by CodeRabbit

- **New Features**
  - Added an anti-abuse taker-bond “pay invoice” flow with deep-link navigation and a new bond-payment screen (loading/waiting state, invoice QR, copy/share, and payment confirmation).
  - Introduced a dedicated “waiting for bond” trade status and updated primary actions to route into the bond-payment flow.
- **Bug Fixes**
  - Improved bond/invoice progression handling with more resilient polling, one-time navigation as status changes, and updated cancellation/backout behavior.
  - Fixed bond countdown timing to use the bond window while in the bond step, then switch afterward.
- **Documentation**
  - Updated localized UI copy for the bond-payment experience across supported languages (including leave/release dialogs and backout messaging).
- **Tests**
  - Added coverage for the bond-payment screen, countdown, and new bond state transitions.


lib/features/order/screens/pay_bond_invoice_screen.dart
```diff
@@ -0,0 +1,421 @@
...
+            break;
+          default:
+            break;
+        }
+      },
+    );
+
+    return tradeAsync.when(
+      loading: () => Scaffold(
+        appBar: AppBar(title: Text(l10n.payBondInvoiceTitle)),
```
grunch, line 327: **Blocking:** Route every back exit through the protocol-aware cancellation policy. This `AppBar` (and the equivalent AppBars in the error, invoice-waiting, NWC, and manual branches) keeps its automatic back action, and the screen has no `PopScope`. AppBar back, Android system back, and gestures therefore pop without calling `_confirmAndCancel()`, even though the code explicitly notes that a bare pop strands the order at `WaitingTakerBond`. Use one shared scaffold/back handler so all rendered branches either confirm and send `cancelOrder`, or deliberately block popping once payment has been launched. Add a widget test covering both AppBar and system back while the bond is unpaid.

lib/features/order/providers/trade_state_provider.dart
```diff
@@ -10,43 +10,55 @@ import 'package:mostro/src/rust/api/types.dart';
...
     StreamProvider.family.autoDispose<BigInt?, String>((ref, orderId) async* {
   while (true) {
-    final info = await orders_api.getOrder(orderId: orderId);
-    final sats = info?.amountSats;
-    yield sats;
-    if (sats != null) return; // done — no need to keep polling
+    try {
+      final trades = await orders_api.listTrades();
+      final trade = trades.where((t) => t.order.id == orderId).firstOrNull;
+      final sats = trade?.order.amountSats ??
```
grunch: **Blocking:** Do not fall back to the public order amount when a persisted trade row exists. In the post-bond buyer flow, that row can exist while its authoritative per-role `amountSats` is still null; `??` then emits the coarse Kind 38383 amount and line 27 permanently stops polling. Branch on `trade != null`: yield its nullable amount and continue polling, and call `getOrder` only when no trade row exists. Please add the missing regression test for a present trade with null `amountSats` plus a non-null public amount.

grunch (changes requested):
Requesting changes on `cdd737f2129792e5d48626b08ced0db52d18db75`. Three correctness blockers remain:

1. The unresolved IndexedDB finding is valid: the new post-bond `AddInvoice` and pre-commit cancel gates depend on `get_trade_by_order_id`, but the WASM backend always returns `None`. The paid-bond flow therefore cannot advance on web.
2. The bond screen's implicit back paths bypass the only protocol-aware cancellation path and can strand an unpaid take at `WaitingTakerBond` until timeout (inline).
3. `tradeAmountProvider` still falls back to the coarse public order amount when an authoritative persisted trade exists with a temporarily-null per-role amount, so it can stop polling with the wrong invoice amount (inline).

The exact-head Rust tests pass (113 passed, 6 ignored), and all GitHub checks are green, but those checks do not exercise the affected Flutter navigation/provider paths or the WASM persistence transition.

rust/src/api/orders.rs
```diff
@@ -1303,10 +1333,36 @@ pub async fn cancel_order(order_id: String) -> Result<()> {
...
+    // a cooperative cancel of a committed trade drops it. The two need different
+    // optimistic book handling (see the fn doc).
+    let is_precommit = match crate::db::app_db::db() {
+        Some(db) => matches!(
+            db.get_trade_by_order_id(&order_id)
+                .await
+                .ok()
+                .flatten()
+                .map(|t| t.order.status),
+            Some(OrderStatus::WaitingTakerBond) | Some(OrderStatus::Pending)
```
grunch: **Blocking:** Model cancellation by role/ownership, not status alone, and carry that classification through the daemon confirmation. This treats every persisted `Pending` trade as a taker pre-commit back-out, but maker-created orders are also persisted as `Pending`; their optimistic cancellation therefore leaves the listing locally available. Conversely, the real daemon taker-back-out path emits `Action::Canceled` and then resets/republishes the order as `Pending`, while this client's shared `Action::Canceled` arm at lines 1812-1831 unconditionally removes the order and persists `Canceled`. That undoes this branch and hides an order that remains available (especially when concurrent taker bonds remain). Add end-to-end regression cases for both maker `Pending` cancellation and taker `WaitingTakerBond`/`Pending` back-out through the actual dispatcher.

rust/src/api/orders.rs
```diff
@@ -2268,19 +2427,40 @@ async fn subscribe_single_order(order_id: &str) {
                                 order.status
                             );
                             last_activity = crate::rt::time::Instant::now();
-                            // Sync trade status in DB so My Trades reflects it.
+                            // Guard the trade row against coarse-bucket
+                            // downgrades (see `public_status_supersedes`); the
+                            // public order-book cache below is left as-is.
```
grunch: **Blocking:** Preserve the fine-grained state in the order-book cache too. The new guard correctly skips the DB downgrade, but line 2468 still upserts the coarse public event and replaces `WaitingTakerBond`/`WaitingBuyerInvoice`/`WaitingPayment` with `Pending`/`InProgress`. The global ingestion path has the same unconditional upsert at line 2868. This can make a locally taken order reappear as available and makes `getOrder`/order-book streams contradict the authoritative trade. Apply the same merge policy before both upserts, and test the real `OrderBook` path; the current regression helper mirrors only the DB decision and stays green if either production upsert regresses.

lib/features/order/screens/pay_bond_invoice_screen.dart
```diff
@@ -0,0 +1,421 @@
...
+
+                        // Pay with external Lightning wallet (lightning: URI)
+                        SizedBox(
+                          width: double.infinity,
+                          child: FilledButton.icon(
+                            onPressed: () async {
+                              final uri = Uri.parse('lightning:$invoice');
+                              bool launched = false;
+                              try {
+                                launched = await launchUrl(
```
grunch, line 458: **Blocking:** Define and enforce a non-destructive state while the payment outcome is unknown. After launching an external wallet, copying/sharing the invoice, or while NWC payment is in flight, the parent screen still exposes `_cancelButton()` until a success callback/status poll arrives. A user can therefore send `cancelOrder` while the bond HTLC is succeeding or already paid. Enter an explicit payment-pending/ambiguous state at initiation (not only after NWC success), disable destructive cancellation until the daemon status resolves, and provide a safe recovery/refresh path if the wallet launch does not result in payment. Cover the launch → return → delayed status transition race.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -375,6 +395,12 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
               : l10n.tradeTimerWaitingPaymentLabelSeller,
           l10n.tradeTimerWaitingInvoiceConsequence,
         ),
+      // Unpaid bond lapses → same "returns to the book" consequence as a
+      // waiting-invoice timeout.
+      TradeStatus.waitingBond => (
```
grunch, line 426: **Blocking:** This label presents `_remaining` as the bond deadline, but `_loadExpiresAt()` derives it from the public order's `expiresAt` (or resets it to a fresh 15 minutes when the order is absent). The Rust trade stores the actual bond lifecycle deadline separately in `TradeInfo.timeoutAt`; for a normal order these values can differ by many hours. Re-entering the trade can therefore show a dangerously false payment window. Source the countdown from the persisted trade deadline for `waitingBond`, and test re-entry with listing expiry != bond timeout plus an absent public order.

grunch (changes requested):
Supplemental exact-head review after completing the full parallel audit. These are additional blockers discovered after the first review was submitted; they are consolidated here to keep the remediation checklist stable:

1. The cancellation state machine conflates maker cancellation with taker bond back-out, and the later `Action::Canceled` handling reverses the intended taker book restoration (inline).
2. The coarse-status guard protects only the persisted trade; both public ingestion paths can still overwrite the in-memory order book with `Pending`/`InProgress` (inline).
3. Payment launch and destructive cancellation can race while the Lightning outcome is unknown (inline).
4. The newly presented bond countdown uses the public listing expiration rather than `TradeInfo.timeoutAt` (inline).

Please also replace the mirrored bond-flow test with coverage that exercises the real dispatcher/order-book path. As a non-blocking documentation follow-up, `specs/004-mostro-p2p-client/contracts/orders.md` still describes bonds as unsupported and `cancel_order` as maker-only.

Catrya (changes requested):
When an order is in the `waiting-buyer-invoice` or `waiting-payment` status and the taker doesn't respond, the order returns to the `Pending` status. However, the maker's app doesn't update to `Pending`, it stays on the previous status. This doesn't happen on `main`.

**Steps to reproduce:**

1. User A creates a sell order.
2. User B takes the order and pays the bond.
3. Mostro asks User B to provide a Lightning invoice.
4. User A sees the order in the `waiting-buyer-invoice` status.
5. User B does **not** provide the invoice and lets the timeout expire.
6. The order returns to the `Pending` status, but User A continues to see `waiting-buyer-invoice` instead of `Pending`.

Catrya (commented):
It’s weird because sometimes it works fine and other times it doesn't. In these cases, all the orders should appear as "pending" since the timeout is 3 minutes and the new "pending" message has already been sent, but not all of them are updating.
Image

Issue #223, opened 2026-07-24
Backup verification: re-show words after two failed attempts + Account "Secret Words" section


grunch:
Not part of #142 — independent, unblocked UX work on the mnemonic backup flow. Split out of the review discussion on PR #204 so it does not sit behind the restore contract work.

### Scope

**1. Failed verification flow**

When the user fails the seed-words verification challenge **twice**, re-show the words instead of letting them keep guessing. Failing repeatedly almost always means the words were not written down correctly — sending the user back to the word list is the recovery, not another attempt.

**2. Account: "Secret Words" section**

The Account screen needs a dedicated "Secret Words" section so the user can review / re-verify their backup after the initial onboarding, rather than the words being reachable only during first-run.

### Notes

- Any state written by this flow should line up with #141 (`backup_confirmed` moving from SharedPreferences to the identity table in Rust) — coordinate so the two do not diverge.
- All strings via `lib/l10n/app_{en,es,fr,de,it}.arb`; no user-facing prose in Rust.
- Showing the seed words again is a sensitive surface — reuse whatever guard the existing backup screen uses (see also #115, optional biometric auth for Account / Backup screens).

### Acceptance criteria

- [ ] Two failed verification attempts return the user to the words, with a clear explanation
- [ ] Account has a "Secret Words" section that reaches the backup/verify flow
- [ ] Backup-confirmed state stays consistent with whatever #141 lands on
- [ ] Widget tests cover the fail-twice path

codaMW:
Both scope items look complete on current main:

1. Fail-twice re-show merged via #224. Verified: the ritual sends the user back to the word list on the second wrong pick (backupRitualSecondFailureMessage), and verification restarts cleanly. Covered by backup_ritual_screen_test.dart (16 tests passing, including "second wrong pick returns to step 1" and "after restarting, one wrong pick is tolerated again").

2. Account "Secret Words" section present in account_screen.dart: a dedicated card with reveal/mask, the backup-confirm checkbox -> backupCompletedProvider, the _BackedUpBadge once complete, and an entry to the backup flow via showBackupTriggerSheet. Strings are localized across the .arb files.

One coordination note: backup state currently lives in SharedPreferences (kBackupCompletedKey / backupCompletedProvider), not yet the Rust identity table that migration is #141, still open. Flagging so the two stay aligned when #141 lands.

Unless there's a specific gap you still want addressed here, this looks ready to close.

grunch:
[...]

PR #224, opened 2026-07-24
fix(#223): backup verify, redirect to 12 words after second wrong pick

codaMW:
Refs #223 backup-verification UX (fail-twice -> re-show words).

## Problem
On the seed-confirmation step, a wrong pick only showed "incorrect" and let the
user keep tapping the same slot. Across retries they could eliminate the decoys
and land on the correct word by process of elimination  "completing" backup
without knowing the words.

## Fix
A per-word wrong-pick counter (`_failCount`), reset to 0 on each correct pick.
A **second wrong pick on the same word** sends the user back to the 12-words
screen with a localized message and restarts verification. The counter is reset
in both `_backToWords()` (the owner) and `_startVerification()` (belt-and-braces).

Practical effect: a user may miss once on each of the three challenge words and
still finish the round  strictly more lenient than "second wrong pick anywhere",
but elimination is still blocked (at most 2 of 4 options per slot).

## Tests
`test/features/account/backup_ritual_screen_test.dart` (4 cases):
- first wrong pick -> stays on step 2 with feedback
- second wrong pick on the same word -> back to step 1 with the SnackBar
- after restart, one wrong pick tolerated (proves the reset)
- all correct -> backup confirms

A `@visibleForTesting debugWords` seam lets the flow run without the Rust bridge;
it's enforced (a production call site fails `flutter analyze`).

## Note on the second part of #223
#223 also covers the Account "Secret Words" section. The current screen already
shows "To restore your account", Show/Hide, the confirm checkbox, and a
"✓ Backed up" badge after backup (banner disappears). I'll split that into its
own issue and update the #223 checklist using `Refs` here so #223 stays open.

### DIFF
--- a/lib/features/account/screens/backup_ritual_screen.dart
+++ b/lib/features/account/screens/backup_ritual_screen.dart
@@ -172,10 +172,9 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
         // up, then restart verification from scratch (#204 review).
         _backToWords();
         ScaffoldMessenger.of(context).showSnackBar(
-          const SnackBar(
+          SnackBar(
             content: Text(
-              'That was incorrect again. Please review and back up your 12 '
-              'words, then verify from the start.',
+              AppLocalizations.of(context).backupRitualSecondFailureMessage,
             ),
           ),
         );
--- a/lib/features/account/screens/backup_ritual_screen.dart
+++ b/lib/features/account/screens/backup_ritual_screen.dart
@@ -59,6 +59,11 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
   /// Option the user last tapped incorrectly (cleared on next tap).
   String? _wrongPick;
 
+  /// Wrong-pick counter for the current verification round. On the 2nd wrong
+  /// pick the user is sent back to the words screen and verification restarts,
+  /// so the confirmation can't be brute-forced by elimination (#204 review).
+  int _failCount = 0;
+
   bool _confirming = false;
 
   @override
@@ -119,6 +124,7 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
       _filled = [null, null, null];
       _activeSlot = 0;
       _wrongPick = null;
+      _failCount = 0;
       _options = _buildOptions(_challenge[0]);
       _step = 1;
     });
@@ -159,7 +165,23 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
         }
       });
     } else {
-      setState(() => _wrongPick = word);
+      _failCount++;
+      if (_failCount >= 2) {
+        // Second wrong pick this round: don't let the user keep guessing by
+        // elimination. Send them back to the 12 words to actually back them
+        // up, then restart verification from scratch (#204 review).
+        _backToWords();
+        ScaffoldMessenger.of(context).showSnackBar(
+          const SnackBar(
+            content: Text(
+              'That was incorrect again. Please review and back up your 12 '
+              'words, then verify from the start.',
+            ),
+          ),
+        );
+      } else {
+        setState(() => _wrongPick = word);
+      }
     }
   }
 
--- a/lib/features/account/screens/backup_ritual_screen.dart
+++ b/lib/features/account/screens/backup_ritual_screen.dart
@@ -31,7 +31,12 @@ const _fallbackDecoys = [
 /// The words live only in this screen's state and are discarded when the
 /// screen is left, restoring the masked behavior of the Account screen.
 class BackupRitualScreen extends ConsumerStatefulWidget {
-  const BackupRitualScreen({super.key});
+  const BackupRitualScreen({super.key, @visibleForTesting this.debugWords});
+
+  /// Test-only word source. When non-null, [_loadWords] uses these instead of
+  /// calling the Rust bridge, so widget tests can drive verification with a
+  /// known mnemonic. Never set in production.
+  final List<String>? debugWords;
 
   @override
   ConsumerState<BackupRitualScreen> createState() => _BackupRitualScreenState();
@@ -59,10 +64,25 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
   /// Option the user last tapped incorrectly (cleared on next tap).
   String? _wrongPick;
 
-  /// Wrong-pick counter for the current verification round. On the 2nd wrong
-  /// pick the user is sent back to the words screen and verification restarts,
-  /// so the confirmation can't be brute-forced by elimination (#204 review).
+  /// Wrong-pick counter for the word currently being verified. Reset to 0 on
+  /// each correct pick, so it counts misses on the word in front of the user.
+  /// A second wrong pick on the same word sends them back to the 12 words and
+  /// restarts verification, so a single slot can't be solved by tapping every
+  /// option in turn (#223).
   int _failCount = 0;
+
+  /// Test-only: the correct word for the slot currently being verified, or null
+  /// if verification isn't active. Lets widget tests tap the right/wrong option
+  /// deterministically despite the randomised challenge. Never used in prod UI.
+  @visibleForTesting
+  String? get debugCorrectWordForActiveSlot {
+    if (_words == null ||
+        _challenge.isEmpty ||
+        _activeSlot >= _challenge.length) {
+      return null;
+    }
+    return _words![_challenge[_activeSlot]];
+  }
 
   bool _confirming = false;
 
@@ -81,7 +101,8 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
 
   Future<void> _loadWords() async {
     try {
-      final words = await IdentityService.getMnemonicWords();
+      final words =
+          widget.debugWords ?? await IdentityService.getMnemonicWords();
       if (!mounted) return;
       if (words.isEmpty) {
         ScaffoldMessenger.of(context).showSnackBar(
@@ -155,6 +176,7 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
       setState(() {
         _filled[_activeSlot] = word;
         _wrongPick = null;
+        _failCount = 0; // correct pick — reset the per-word miss budget
         final next = _filled.indexWhere((w) => w == null);
         if (next != -1) {
           _activeSlot = next;
@@ -165,11 +187,11 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
         }
       });
     } else {
-      _failCount++;
+      setState(() => _failCount++);
       if (_failCount >= 2) {
-        // Second wrong pick this round: don't let the user keep guessing by
-        // elimination. Send them back to the 12 words to actually back them
-        // up, then restart verification from scratch (#204 review).
+        // Second wrong pick on this word: don't let the user grind through the
+        // options by elimination. Send them back to the 12 words to review,
+        // then restart verification from scratch (#223).
         _backToWords();
         ScaffoldMessenger.of(context).showSnackBar(
           SnackBar(
@@ -218,6 +240,7 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
       _activeSlot = 0;
       _options = const [];
       _wrongPick = null;
+      _failCount = 0; // round is over — the next one starts clean
     });
   }
 
--- a/test/features/account/backup_ritual_screen_test.dart
+++ b/test/features/account/backup_ritual_screen_test.dart
@@ -46,7 +46,9 @@ Finder _aWrongOption(WidgetTester tester) {
   final correct = _correctWord(tester);
   for (final w in _words) {
     if (w == correct) continue;
-    final f = find.text(w);
+    // Scope to the option buttons (InkWell), so a filled answer chip with the
+    // same text can't be matched instead once slots start filling.
+    final f = find.widgetWithText(InkWell, w);
     if (f.evaluate().isNotEmpty) return f.first;
   }
   fail('no wrong option visible');
@@ -106,7 +108,6 @@ void main() {
     await tester.pump();
     expect(find.text(l10n.backupRitualStep1Title), findsOneWidget);
 
-    // Clear the SnackBar so it can't intercept the restart tap, then restart.
     // The failure SnackBar sits at the bottom over the verify button; clear it
     // so it can't intercept the restart tap.
     ScaffoldMessenger.of(tester.element(find.byType(BackupRitualScreen)))
@@ -130,7 +131,7 @@ void main() {
     // Answer each of the three challenge slots with its correct word.
     for (var i = 0; i < 3; i++) {
       final correct = _correctWord(tester);
-      await tester.tap(find.text(correct).first);
+      await tester.tap(find.widgetWithText(InkWell, correct).first);
       await tester.pumpAndSettle();
     }
 
--- a/test/features/account/backup_ritual_screen_test.dart
+++ b/test/features/account/backup_ritual_screen_test.dart
@@ -1,0 +1,143 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_localizations/flutter_localizations.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:shared_preferences/shared_preferences.dart';
+import 'package:mostro/features/account/screens/backup_ritual_screen.dart';
+import 'package:mostro/l10n/app_localizations.dart';
+
+const _words = <String>[
+  'abandon', 'ability', 'able', 'about', 'above', 'absent',
+  'absorb', 'abstract', 'absurd', 'abuse', 'access', 'accident',
+];
+
+Future<void> _pumpRitual(WidgetTester tester) async {
+  tester.view.physicalSize = const Size(1200, 2400);
+  tester.view.devicePixelRatio = 1.0;
+  addTearDown(tester.view.resetPhysicalSize);
+  addTearDown(tester.view.resetDevicePixelRatio);
+
+  await tester.pumpWidget(
+    const ProviderScope(
+      child: MaterialApp(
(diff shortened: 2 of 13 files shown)


codaMW:
Localized via new ARB key backupRitualSecondFailureMessage, count-neutral (dropped "12",  correct for 24-word mnemonics now), with de/es/fr/it translations. Verified live on Linux desktop screenshot shows the new "back up your secret words" message on the redirect. Native-speaker review welcome on the four translations.

Screenshot 2026-07-24 090125
Screenshot 2026-07-24 090139
Screenshot 2026-07-24 090328
Screenshot 2026-07-24 153758
Screenshot 2026-07-24 153815

lib/features/account/screens/backup_ritual_screen.dart
```diff
@@ -59,6 +59,11 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
   /// Option the user last tapped incorrectly (cleared on next tap).
   String? _wrongPick;
 
+  /// Wrong-pick counter for the current verification round. On the 2nd wrong
+  /// pick the user is sent back to the words screen and verification restarts,
+  /// so the confirmation can't be brute-forced by elimination (#204 review).
+  int _failCount = 0;
```
grunch: ⚠️ **Stale cross-reference.**

`#204` is a closed PR for a different feature (restore sessions from mnemonic). The traceable reference for this change is #223. Same applies to the inline comment on lines 170-172.

```suggestion
  /// Wrong-pick counter for the current verification round. On the 2nd wrong
  /// pick the user is sent back to the words screen and verification restarts,
  /// so a single slot can't be solved by tapping every option in turn (#223).
  int _failCount = 0;
```

lib/features/account/screens/backup_ritual_screen.dart
```diff
@@ -159,7 +165,22 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
         }
       });
     } else {
-      setState(() => _wrongPick = word);
+      _failCount++;
+      if (_failCount >= 2) {
```
grunch, line 191: ❓ **Confirm the attempt budget: one miss per round, not per word.**

`_failCount` accumulates across all three slots, so this sequence bounces the user:

1. slot 1 → wrong (`_failCount = 1`), then correct;
2. slot 2 → correct;
3. slot 3 → one wrong pick (`_failCount = 2`) → back to step 1.

A single mistap anywhere in a three-word challenge costs the whole round. That may well be the intent — it is what makes the elimination grind impossible — but the PR title and #223 both read as "two wrong picks", which a user would naturally take as two chances at the word currently in front of them.

Worth an explicit decision, and worth recording in the field doc either way: `/// Wrong-pick counter for the current verification round` is the only place these semantics live.

lib/features/account/screens/backup_ritual_screen.dart
```diff
@@ -159,7 +165,22 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
         }
       });
     } else {
-      setState(() => _wrongPick = word);
+      _failCount++;
```
grunch: 🧹 **State mutated outside `setState`.**

`_failCount` is the only state field in this class mutated outside a `setState` call. Harmless today because nothing in `build()` reads it, but it makes the field an outlier in an otherwise consistent class, and it becomes a stale-UI bug the moment anything renders "1 attempt left" or similar.

```suggestion
      setState(() => _failCount++);
```

lib/features/account/screens/backup_ritual_screen.dart
```diff
@@ -159,7 +165,22 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
...
+      if (_failCount >= 2) {
+        // Second wrong pick this round: don't let the user keep guessing by
+        // elimination. Send them back to the 12 words to actually back them
+        // up, then restart verification from scratch (#204 review).
+        _backToWords();
```
grunch, line 195: 🛠️ **`_backToWords()` doesn't reset `_failCount` — latent trap.**

`_backToWords()` is the "abandon this verification round" helper: it clears `_challenge`, `_filled`, `_activeSlot`, `_options` and `_wrongPick`. After this PR it resets everything about the round *except* the new counter.

It is safe today — I verified that `_step = 1` is assigned only inside `_startVerification()`, which does reset the counter — but the invariant is now split across two methods with nothing enforcing it. Any future path back into step 2 that skips `_startVerification()` (resume, deep link, a "retry verification" shortcut) would start at `_failCount == 2` and bounce the user on their first miss, permanently.

Cheapest fix is to let the helper own the reset:

```dart
void _backToWords() {
  setState(() {
    _step = 0;
    _challenge = const [];
    _filled = [null, null, null];
    _activeSlot = 0;
    _options = const [];
    _wrongPick = null;
    _failCount = 0; // round is over — the next one starts clean
  });
}
```

Keeping the reset in `_startVerification()` too is fine as belt-and-braces, but one of the two should be the clear owner.

**Secondary point on the same lines:** `_backToWords()` clears `_wrongPick`, so the pick that *triggered* the redirect never gets the red highlight — the user is yanked off the screen with no inline indication of what they got wrong. Consider flagging the pick briefly before transitioning, or giving step 1 a persistent explanation instead of relying on a SnackBar that disappears after a few seconds.

grunch (changes requested):

## Walkthrough

Adds a per-round wrong-pick counter to the backup ritual's verification step. The first wrong pick keeps the existing inline "try again" feedback; the second sends the user back to step 1 (the 12 words) through the existing `_backToWords()` helper and surfaces a localized SnackBar. `_startVerification()` resets the counter so each round starts fresh. One new l10n key across all five locales, plus regenerated `AppLocalizations`.

| File | Change |
|---|---|
| `lib/features/account/screens/backup_ritual_screen.dart` | `_failCount` field, second-miss redirect in `_onOptionTap`, reset in `_startVerification` |
| `lib/l10n/app_{en,es,fr,de,it}.arb` | `backupRitualSecondFailureMessage` (+ `@description` in the `en` template) |
| `lib/l10n/app_localizations*.dart` | regenerated |

## Verification performed on the PR head (`f55d31b`)

| Check | Result |
|---|---|
| `flutter gen-l10n` | ✅ no diff — the committed generated files match the ARBs |
| `flutter analyze lib/features/account/screens/backup_ritual_screen.dart` | ✅ No issues found |
| l10n coverage | ✅ all 5 locales updated; `@description` correctly only in the `en` template |
| Terminology | ✅ "secret words" / "palabras secretas" matches `secretWordsTitle`, `tapToViewAndSaveSecretWords`, `generateNewUserDialogContent` |
| Control flow | ✅ `_step = 1` is assigned **only** in `_startVerification()`, so `_failCount` is always reset before a round starts — the missing reset in `_backToWords()` is a latent trap, not a live bug |

---

## Blocking

### 1. No test covers the new behaviour — severity: **high**

`test/features/account/` contains only `backup_reminder_provider_test.dart`; there is no widget test for `BackupRitualScreen`, and none is added here. The repo requires tests for new functionality with an 80% floor, and #223's acceptance criteria lists "Widget tests cover the fail-twice path" explicitly.

This change is exactly the kind that regresses silently: a future refactor of `_onOptionTap` that drops the counter restores unlimited guessing with **no visible symptom**, and "tested live on Linux desktop" will not catch it.

Minimum coverage requested:

- first wrong pick → stays on step 2, `wrongPickMessage` visible;
- second wrong pick → back on step 1 (words grid rendered) with the SnackBar shown;
- after restarting verification, one wrong pick is tolerated again (proves the reset);
- happy path still confirms the backup.

### 2. `Closes #223` closes a half-implemented issue — severity: **high**

#223 has two parts: (a) fail-twice → re-show words, and (b) the Account "Secret Words" section. This PR implements (a) and asks about (b) in the description — so merging as-is auto-closes the issue with (b) unresolved and no longer tracked anywhere.

Please use `Refs #223` here and let the issue close when (b) is settled, or split (b) into its own issue and update the checklist on #223.

---

## Non-blocking

| Severity | Finding | Where |
|---|---|---|
| Medium | `_backToWords()` — the round-reset helper — does not reset `_failCount` | inline |
| Medium | Code comments cite "#204 review"; #204 is a closed PR about restore-from-mnemonic | inline |
| Low | `_failCount++` is mutated outside `setState` | inline |
| Low | The failing pick gets no inline feedback before the redirect; the transient SnackBar is the only explanation | inline |
| Low | One miss per *round* (across all 3 slots), not per word — confirm this is the intended budget | inline |
| Low | Accessibility: the redirect is announced only through a SnackBar; a persistent banner on step 1 would survive the transition and be readable by assistive tech | — |

---

## Note (not a change request)

Step 2 already has a "Show words again" button, so a determined user can re-read the words at any time. That is fine — the goal here is to *force* the review, not to make passing without it impossible. Worth keeping in mind when wording the code comment, which currently claims the confirmation "can't be brute-forced by elimination"; what the change actually prevents is grinding a single slot through all four options.

Nice, focused change otherwise — the fix is in the right place and reuses the existing helper rather than inventing a second path back to the words.

codaMW:
Thanks for the thorough review all addressed in the latest commit:

Blocking:

Tests added test/features/account/backup_ritual_screen_test.dart with all four cases you listed: first wrong pick stays on step 2 with wrongPickMessage; second wrong pick returns to step 1 with the SnackBar; after restart one wrong pick is tolerated (proves the reset); happy path confirms the backup. All 4 pass. Needed a small @visibleForTesting seam (debugWords + debugCorrectWordForActiveSlot) since IdentityService.getMnemonicWords() is a static Rust call the seam lets the flow run without the bridge.
Closes -> Refs  updated the description so #223 stays open for part (b).

Non-blocking:

_backToWords() now owns the counter reset.
_failCount++ wrapped in setState.
Stale #204 refs -> #223. Attempt budget: switched to per-word _failCount resets on each correct pick, so it's "two chances at the word in front of you", matching the reading you flagged. Documented in the field comment.
Reworded the comment: you're right, it prevents grinding a single slot through all options, not brute-force in general.

For part (b) (Account "Secret Words"), I'll open a separate issue and update the #223 checklist, unless you'd rather keep it here.

test/features/account/backup_ritual_screen_test.dart
```diff
@@ -0,0 +1,143 @@
...
+}
+
+String _correctWord(WidgetTester tester) {
+  final dynamic state = tester.state(find.byType(BackupRitualScreen));
+  final String? w = state.debugCorrectWordForActiveSlot as String?;
+  expect(w, isNotNull, reason: 'verification should be active');
+  return w!;
+}
+
+Finder _aWrongOption(WidgetTester tester) {
```
grunch, line 45: 🧹 **Unscoped finder — latent, not live.**

`_aWrongOption` searches `find.text(w)` across the whole tree and returns `.first`. It is correct today for two reasons that are incidental rather than guaranteed:

- it is only ever called before a slot has been filled, and
- the fixture `_words` happens to be 12 distinct words.

Once a slot is filled, its chip renders **above** the options grid, so `.first` would resolve to the filled chip instead of an option button. `tester.tap` on it silently does nothing — no exception — and the test would then assert against stale state and fail with a confusing message rather than pointing at the finder.

I ran the file 12 times to check the randomised challenge doesn't already trip this; it doesn't. Still worth scoping the finder to the option buttons (e.g. `find.widgetWithText(InkWell, w)` or a `find.descendant` under the options grid). Line 133's `find.text(correct).first` in the happy-path test has the same shape and is the one most likely to bite, since it runs *after* slots start filling.

test/features/account/backup_ritual_screen_test.dart
```diff
@@ -0,0 +1,143 @@
...
+
+    await tester.tap(_aWrongOption(tester));
+    await tester.pumpAndSettle();
+    await tester.tap(_aWrongOption(tester));
+    await tester.pump();
+    expect(find.text(l10n.backupRitualStep1Title), findsOneWidget);
+
+    // Clear the SnackBar so it can't intercept the restart tap, then restart.
+    // The failure SnackBar sits at the bottom over the verify button; clear it
+    // so it can't intercept the restart tap.
```
grunch: 🧹 **Duplicated comment.**

These three lines say the same thing twice — looks like an editing leftover.

```suggestion
    // The failure SnackBar sits at the bottom over the verify button; clear it
    // so it can't intercept the restart tap.
```

lib/features/account/screens/backup_ritual_screen.dart
```diff
@@ -59,6 +64,26 @@ class _BackupRitualScreenState extends ConsumerState<BackupRitualScreen> {
...
+  /// each correct pick, so it counts misses on the word in front of the user.
+  /// A second wrong pick on the same word sends them back to the 12 words and
+  /// restarts verification, so a single slot can't be solved by tapping every
+  /// option in turn (#223).
+  int _failCount = 0;
+
+  /// Test-only: the correct word for the slot currently being verified, or null
+  /// if verification isn't active. Lets widget tests tap the right/wrong option
+  /// deterministically despite the randomised challenge. Never used in prod UI.
+  @visibleForTesting
```
grunch, line 77: 🧹 **`@visibleForTesting` here is a no-op.**

`debugCorrectWordForActiveSlot` is declared on `_BackupRitualScreenState`, which is private to this library — nothing outside the file can name it, annotated or not. (Contrast with `debugWords` on line 34, where the annotation *is* enforced: I verified a production call site produces `invalid_use_of_visible_for_testing_member` and fails `flutter analyze`.)

Harmless, but it implies a guarantee the annotation isn't providing here. Worth dropping, or keeping purely as documentation with the doc comment carrying the intent — which it already does.

Related, no change needed: because the state class is private, the test reaches this through `final dynamic state = tester.state(...)`, so the `as String?` cast gives no compile-time safety. Renaming the getter breaks the test at runtime rather than at analysis time. That's an acceptable trade for a private State class, just don't count on the analyzer to catch a rename here.

grunch (commented):
**Re-review of `1cd1cb2` — actionable comments posted: 3**

Everything from the previous round is addressed on the code side. One blocker remains, and it's in the PR description rather than the code.

## Previous round

| Finding | Status |
|---|---|
| No test covers the new behaviour | ✅ 4 widget tests added, covering exactly the requested cases |
| `_backToWords()` doesn't reset `_failCount` | ✅ reset moved into the helper |
| `_failCount++` outside `setState` | ✅ wrapped |
| Comments cite "#204 review" | ✅ now #223 |
| Attempt budget ambiguous | ✅ resolved by switching to **per-word** (reset on each correct pick) and documenting it on the field |
| `Closes #223` closes a half-implemented issue | ❌ **still open** |

Nice call on the per-word semantics — it's the more defensible reading and the field doc now records it.

## Verification performed on `1cd1cb2`

| Check | Result |
|---|---|
| `flutter test test/features/account/backup_ritual_screen_test.dart` | ✅ 4/4 pass |
| Flake check (challenge slots and decoys are randomised, unseeded) | ✅ **12 consecutive runs, 0 failures** |
| `flutter analyze` on both changed files | ✅ No issues found |
| Is the `debugWords` seam actually enforced? | ✅ **verified** — see below |

On the last one: I expected `@visibleForTesting` on a constructor *parameter* (rather than on the field) to be decorative, so I checked it rather than assert it. Adding a production call site under `lib/` that passes `debugWords:` produces

```
warning • The member 'debugWords' can only be used within
'package:mostro/features/account/screens/backup_ritual_screen.dart' or a test
• invalid_use_of_visible_for_testing_member
```

and `flutter analyze` exits 1, which CI runs (`.github/workflows/ci.yml:130`). So the seam is genuinely closed against production use — no action needed. Worth stating explicitly, since a test hook that injects seed words into the screen whose job is protecting them is the kind of thing that deserves proof rather than a reviewer's assumption.

## Blocking

### `Closes #223` + the description no longer matches the code — severity: **high**

Two problems in the PR body, one of them carried over:

1. **`Closes #223`** still closes an issue whose second half (the Account "Secret Words" section) is unresolved — you raise that very question further down in the same description. Please use `Refs #223`.

2. **The description now describes code that no longer exists.** It still says:
   - "Added a `_failCount` field (**per verification round**)" — it is per *word* since `1cd1cb2`;
   - "`_startVerification` resets `_failCount` to 0, so each new round starts fresh" — the reset that matters now is the one on each correct pick, plus the one in `_backToWords()`;
   - it quotes the message as "…back up your **12** words…" — that hardcoded string was removed in `f55d31b`.

   Worth noting the practical consequence, since it's easy to miss: per-word budget means a user can now miss once on *each* of the three words and still finish the round, where before a single miss anywhere ended it. That's the right trade-off (elimination is still blocked — 2 of 4 options max per slot), but it is strictly more lenient than "second wrong pick in a verification round", which is what the description and title still promise.

Fix the body and this is good to merge; I'll flip to approve. My earlier `CHANGES_REQUESTED` stands only on this item.

## Non-blocking

| Severity | Finding | Where |
|---|---|---|
| Low | `_aWrongOption` / `find.text(...).first` are unscoped — latent, not live | inline |
| Low | `@visibleForTesting` on a member of a private class is a no-op | inline |
| Low | Duplicated comment in the third test | inline |
| Low | No test drives `_backToWords()` through the "Show words again" button or the AppBar back arrow — the redirect path is covered, the manual path isn't, and the counter reset now lives in that helper | — |
| Low | Still open from the last round: the pick that triggers the redirect never gets the red highlight, since `_backToWords()` clears `_wrongPick`. The user is moved off the screen with only a transient SnackBar explaining why | — |

grunch (approved):
Approving — the blocker is resolved.

`Refs #223` is in, and the description now matches the code: per-word counter, `_backToWords()` named as the reset owner, and the leniency trade-off spelled out. Part (b) of #223 stays tracked instead of being auto-closed.

Recap of what I verified on `1cd1cb2`: 4/4 widget tests pass, 12 consecutive runs with no flake against the randomised challenge, `flutter analyze` clean on both files, and the `debugWords` seam confirmed closed against production use.

One factual nit for whoever reads this later, no change needed: the description says the seam "lets the flow run without the Rust bridge". It avoids *calling* `IdentityService.getMnemonicWords()`, but the test still needs generated `lib/src/rust/` to compile — I hit `Error when reading 'lib/src/rust/api/identity.dart'` on a clean checkout until I generated it.

The three low-severity inline comments (unscoped finder, no-op `@visibleForTesting` on the private state member, duplicated comment) are yours to take or leave — none of them block. Good iteration on this one.

Issue #219, opened 2026-07-24
Restore: durable background restore (restore_id, persisted pending request, late replies applied)


grunch:
Sub-issue of #142. **Depends on:** #215.

### Problem

The handshake currently waits on a fixed 10-second in-process timeout. The daemon's restore window is on the order of an hour, so a valid reply routinely arrives after the waiter is gone — and is then **discarded**. Logging the late reply only makes the loss visible; it does not fix it.

Restore is not a foreground RPC. It is a **durable background operation**.

### Scope

- Give each restore an explicit `restore_id` / `request_id`, echoed by progress, final and error messages.
- **Persist the pending restore locally**, so it survives navigating away from the screen and a full app restart.
- The dispatcher must apply a valid final `RestoreData` **even when no UI waiter is attached**.
- Drop the fixed short timeout in favour of the durable model; keep a bounded retry/abort path so a restore cannot hang forever with no way out.
- Late, duplicate or out-of-order replies must be handled safely (this pairs with the idempotency requirement in #218).

### Acceptance criteria

- [ ] A reply arriving well after the request (> 10s, and after an app restart) is still applied
- [ ] Navigating away from the restore screen does not discard the result
- [ ] A restore in flight is visible in local state after a cold start
- [ ] A duplicate final reply for the same `restore_id` is a no-op
- [ ] The user can cancel/abandon a restore explicitly, and the abandoned `restore_id` is then ignored

codaMW:
Picking this up. Two scoping questions before I build, since #219 sits between a few moving parts:

1. The #218 boundary. #219's ACs say late replies are "applied" and a duplicate final reply is "a no-op" but applying/reconstruction is #218's scope (validate -> upsert -> persist mappings -> resync -> refresh subs -> is_mine), and #218 depends on #216/#217. My read is that #219 owns the durable transport: give each restore a restore_id, persist the pending restore so it survives navigation + restart, capture a valid RestoreData even with no UI waiter, drop the 10s timeout for a bounded durable model, and dedup/cancel by restore_id then hand the captured snapshot off to #218 to actually apply. Is that the intended split, or do you want an interim apply in #219?
[...]

Issue #217, opened 2026-07-24
Restore: resync trade_key_index to the max recovered index (prevents trade-key reuse)


grunch:
Sub-issue of #142. **Depends on:** #215.

### Problem

After a restore, the local `trade_key_index` counter is still at its post-install value (0/1) while the recovered trades already occupy higher indexes. The next order the user creates therefore **reuses a trade key that is already bound to a recovered trade** — a real correctness bug, not just cosmetics.

### Scope

When a valid `RestoreData` is processed, raise `trade_key_index` to the **maximum recovered index across both orders and disputes**, never lower it. This is a monotonic bump: a restore must never rewind the counter (e.g. a stale or partial reply must not un-advance a counter that is already ahead).

This can land with the payload shape available today (`trade_index` is already present) — it does **not** wait for the snapshot contract in #216.

### Acceptance criteria

- [ ] `trade_key_index` is raised to `max(recovered order indexes, recovered dispute indexes)` on a valid restore
- [ ] The counter is never lowered, including on a restore that returns fewer/older trades
- [ ] Regression test: after restoring a trade at index > 1, the next `derive_trade_key()` returns a **fresh** index, not a recovered one
- [ ] Applying the same `RestoreData` twice leaves the counter unchanged (idempotent)

codaMW:
Note for when this is implemented: #215's restore_session() handshake consumes one trade_key_index via derive_trade_key() (the daemon registers every index it sees, so it's persisted even on timeout). The resync here must therefore be max(recovered, current) never a plain assignment  or it would rewind past an index the handshake already consumed

Issue #215, opened 2026-07-24
Restore: RestoreSession handshake (send, correlate reply, subscribe)


grunch:
Sub-issue of #142.

Send a `RestoreSession` request to the active Mostro daemon and correlate the `RestoreData` reply. **Handshake only** — no state reconstruction, no persistence side effects.

### Context

PR #204 proved this path works end to end against a regtest daemon, but it mixed the handshake with partial reconstruction (trade-index mappings, retroactive `is_mine`). That PR was closed in favour of an issue-per-unit split; this issue re-lands only the handshake, as a clean reviewable unit. The closed PR is a useful implementation reference.

### Scope

- `mostro::actions::restore_session()` — build `Message::new_restore(None)`. The payload MUST be `None`; the daemon rejects anything else in `MessageKind::verify`. The Seal is signed by the **identity** key (the daemon looks up trades by master key), the rumor by a **fresh trade** key (the daemon replies to `event.sender`).
- `PendingRequestKind::Restore` + `take_matching_restore()` in `rust/src/api/orders.rs`. Correlation is **by trade pubkey, not `request_id`** — `RestoreSession` carries no `request_id`, so the usual nonce gate does not apply.
- `Action::RestoreSession` arm in `dispatch_mostro_message()` routing `Payload::RestoreData` to the waiting caller.
- Subscribe gift-wraps on the fresh trade key **and** register the pending record **before** publishing, so the reply cannot race the map.
- Privacy-mode guard: recovery is only possible in Reputation mode — Full-Privacy trades are anonymous by design and cannot be replayed by the daemon. Return a stable marker (e.g. `PrivacyModeRecoveryUnavailable`) for Dart to localize; do not return English prose from Rust.

### Out of scope (separate sub-issues)

- Persisting `order_id -> trade_index` mappings
- Retroactive `is_mine` on cached orders
- `trade_key_index` resync
- Durable/background behaviour, progress, and UI wiring

### Acceptance criteria

- [ ] Handshake round-trips against a regtest daemon, covered by an `#[ignore]`d E2E test with the run command documented in the test module
- [ ] This path performs **no** DB writes, **no** key-counter changes and **no** `is_mine` mutation
- [ ] `restore_session()` stays internal (`#[frb(ignore)]`) until UI wiring lands
- [ ] Privacy-mode identities are rejected before any network traffic is emitted

codaMW:
[...]

Issue #220, opened 2026-07-24
Restore UI: entry points (Walkthrough + Account) and truthful phase progress overlay


grunch:
Sub-issue of #142. **Depends on:** #218, #219.

There is no `lib/features/restore/` in v2. This is the UI wiring for the recovery flow, once the reconstruction and durability work is in place.

### Entry points (v1 parity)

- **Walkthrough** — "I already have a seed"
- **Account** — "Restore from mnemonic"

### Progress model

Until the daemon exposes real server-side progress, the UI shows **indeterminate progress plus truthful local phases** — never a synthetic percentage:

`requesting` -> `response received` -> `rebuilding` -> `resubscribing` -> `complete`

Restore can take a while on a long history, so the overlay is visible by design (v1 has `restore_overlay` / `restore_progress_notifier` / `restore_progress_state` as reference — do not copy v1 blindly, the Rust core provides much of this differently).

If real server-side progress is wanted later, that needs an explicit progress message / chunk contract in `mostro-core` + daemon — out of scope here.

### Also in scope

- Surface the Full-Privacy rejection as a localized message (Rust returns a stable marker such as `PrivacyModeRecoveryUnavailable`; Dart maps it via `AppLocalizations`).
- All new strings added to `lib/l10n/app_{en,es,fr,de,it}.arb`.
- Reflect a restore that is still running after an app restart (see #219) rather than showing a fresh empty screen.

### Acceptance criteria

- [ ] Both entry points reach the restore screen
- [ ] Phases shown match the actual local state, with no fabricated percentage
- [ ] Error and privacy-mode cases show localized, actionable messages
- [ ] No user-facing prose hardcoded in Rust
- [ ] Widget tests cover the phase transitions and the error paths

PR #225, opened 2026-07-24
feat(#215): RestoreSession handshake, send, correlate reply, subscribe

codaMW:
Refs #215 RestoreSession handshake (send, correlate reply, subscribe).

Re-lands the handshake from the closed #204, scoped to the handshake only no
reconstruction side effects (those are #217/#218/#219).

## What it does
Sends a `RestoreSession` request to the active daemon and correlates the
`RestoreData` reply. The Seal is signed by the identity key (daemon looks up
trades by master key); the rumor is signed by a fresh trade key (daemon replies
to `event.sender`). Correlation is by trade pubkey, not `request_id`
`RestoreSession` carries no `request_id`, so `take_matching_restore()` skips the
nonce gate.

## Scope
- `mostro::actions::restore_session()` `Message::new_restore(None)`
- `PendingRequestKind::Restore` + `take_matching_restore()`
- `Action::RestoreSession` arm in `dispatch_mostro_message()`
- subscribe on the fresh trade key + register the pending record before publishing
- privacy-mode guard returning `PrivacyModeRecoveryUnavailable` before any
  network traffic
- `restore_session()` stays `#[frb(ignore)]` (internal until UI wiring)

## Out of scope (separate sub-issues)
- `order_id -> trade_index` persistence
- retroactive `is_mine` on cached orders
- `trade_key_index` resync (#217)
- durable/background behaviour, progress, UI (#219)

## Note on state (corrects an inaccurate claim in an earlier revision)
`restore_session()` consumes one trade-key index via `derive_trade_key()`, which
increments and persists `trade_key_index`. This is inherent to the derivation
the daemon registers every index it sees, so the counter must survive even
timed-out operations. Restore is therefore **not** zero-side-effect: it burns one
index per attempt. #217's resync must be `max(recovered, current)`, never a plain
assignment, or it would rewind past an index consumed here (noted on #217). No
`is_mine` mutation and no reconstruction, as before.

Separately: `import_from_mnemonic` now sources the real `privacy_mode` (was
hardcoded `false`, which meant the recovery guard could never fire). This affects
the `recover: false` path Dart uses today an intentional fix; happy to split it
out if preferred.

## Tests
CI-runnable unit tests (no daemon needed):
- `restore_session_payload_none_action_restore_rumor_by_trade_key` asserts the
  payload is `None` (daemon rejects anything else in `MessageKind::verify`), the
  action is `RestoreSession`, the rumor is authored by the trade key
  (`event.sender`, the daemon's reply target), and the Seal is signed by the
  identity key (how the daemon locates the user's trades) both signing roles.
- `take_matching_restore_returns_restore_and_ignores_others` returns and removes
  a pending RESTORE record, ignores non-RESTORE kinds (pins the nonce-gate
  asymmetry).

`cargo test --lib` -> 113 passed, 0 failed (2 E2E tests still `#[ignore]`d, run
command documented in the test module). `cargo clippy` clean. Build clean, CI green.

### DIFF
--- a/rust/src/api/identity.rs
+++ b/rust/src/api/identity.rs
@@ -164,8 +164,23 @@ pub async fn load_identity_from_mnemonic(
 /// When `recover = true`, the daemon recovery flow is triggered (Phase 7).
 /// Currently this validates and loads the mnemonic; recovery contacts are
 /// initiated separately via the daemon API.
-pub async fn import_from_mnemonic(words: Vec<String>, _recover: bool) -> Result<IdentityInfo> {
-    load_identity_from_mnemonic(words, 0, false, None).await
+pub async fn import_from_mnemonic(words: Vec<String>, recover: bool) -> Result<IdentityInfo> {
+    // Source the authoritative privacy mode BEFORE loading, so the imported
+    // identity reflects the current session's mode and the recovery guard is
+    // enforced consistently (a hardcoded `false` here left info.privacy_mode
+    // always false, so the guard below could never fire — while
+    // restore_session() reads the real flag from reputation::get_privacy_mode()).
+    let privacy_mode = crate::api::reputation::get_privacy_mode();
+    let info = load_identity_from_mnemonic(words, 0, privacy_mode, None).await?;
+    if recover {
+        // Recovery is only possible in Reputation mode; Full-Privacy trades
+        // are anonymous by design and cannot be replayed by the daemon.
+        if privacy_mode {
+            bail!("PrivacyModeRecoveryUnavailable");
+        }
+        crate::api::orders::restore_session().await?;
+    }
+    Ok(info)
 }
 
 /// Import identity from an nsec (bech32-encoded Nostr secret key).
--- a/rust/src/api/identity.rs
+++ b/rust/src/api/identity.rs
@@ -165,19 +165,18 @@ pub async fn load_identity_from_mnemonic(
 /// Currently this validates and loads the mnemonic; recovery contacts are
 /// initiated separately via the daemon API.
 pub async fn import_from_mnemonic(words: Vec<String>, recover: bool) -> Result<IdentityInfo> {
-    // Source the authoritative privacy mode BEFORE loading, so the imported
-    // identity reflects the current session's mode and the recovery guard is
-    // enforced consistently (a hardcoded `false` here left info.privacy_mode
-    // always false, so the guard below could never fire — while
-    // restore_session() reads the real flag from reputation::get_privacy_mode()).
+    // Source the authoritative privacy mode up front. Recovery is only possible
+    // in Reputation mode; Full-Privacy trades are anonymous by design and can't
+    // be replayed by the daemon.
     let privacy_mode = crate::api::reputation::get_privacy_mode();
+    // Reject privacy-mode recovery BEFORE loading — otherwise the identity is
+    // already swapped when we bail, leaving the user in a mutated state, and it
+    // violates the "reject before any network traffic" contract for restore.
+    if recover && privacy_mode {
+        bail!("PrivacyModeRecoveryUnavailable");
+    }
     let info = load_identity_from_mnemonic(words, 0, privacy_mode, None).await?;
     if recover {
-        // Recovery is only possible in Reputation mode; Full-Privacy trades
-        // are anonymous by design and cannot be replayed by the daemon.
-        if privacy_mode {
-            bail!("PrivacyModeRecoveryUnavailable");
-        }
         crate::api::orders::restore_session().await?;
     }
     Ok(info)
--- a/rust/src/api/identity.rs
+++ b/rust/src/api/identity.rs
@@ -177,6 +177,12 @@ pub async fn import_from_mnemonic(words: Vec<String>, recover: bool) -> Result<I
     }
     let info = load_identity_from_mnemonic(words, 0, privacy_mode, None).await?;
     if recover {
+        // NOTE: recovery is best-effort relative to the import, but this `?`
+        // propagates a restore failure AFTER the identity has already been
+        // swapped — so a slow/unreachable daemon makes the caller see "import
+        // failed" when the import itself succeeded and only recovery didn't.
+        // Not reachable today (identity_service.dart passes recover: false).
+        // #219 restructures the waiting; revisit this propagation when it lands.
         crate::api::orders::restore_session().await?;
     }
     Ok(info)
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -53,6 +53,10 @@ enum DaemonReply {
     Acknowledged,
     /// Daemon rejected the request with a CantDo reason.
     Rejected { reason: String, message: String },
+    /// Daemon replied to a RestoreSession with the user's active trades and
+    /// disputes. Correlated by identity pubkey (RestoreSession carries no
+    /// request_id) — see take_matching_restore.
+    Restored(mostro_core::message::RestoreSessionInfo),
 }
 
 /// What kind of outgoing request a pending record tracks.
@@ -72,6 +76,10 @@ enum PendingRequestKind {
     Take,
     /// A buyer's add-invoice awaiting the daemon's acknowledgement.
     AddInvoice,
+    /// A session-restore awaiting the daemon's RestoreData reply. Correlated
+    /// by identity pubkey, not request_id (the RestoreSession message carries
+    /// no request_id — see mostro-core Message::new_restore).
+    Restore,
 }
 
 /// Everything one outgoing daemon request needs tracked until its reply is
@@ -124,6 +132,18 @@ fn request_id_matches(expected: u64, got: Option<u64>) -> bool {
 /// in place: relays can replay historical events, and a stale reply must not
 /// confirm, reject, or reconcile a live request — the genuine reply (carrying
 /// the nonce) arrives later and finds the record.
+/// Remove and return the pending RESTORE request for `pubkey_hex`. Unlike
+/// `take_matching_request`, there is no request_id gate: the RestoreSession
+/// message carries no request_id, so the daemon's RestoreData reply is
+/// correlated purely by the identity pubkey it is addressed to.
+fn take_matching_restore(pubkey_hex: &str) -> Option<PendingRequest> {
+    let mut map = pending_requests().lock().ok()?;
+    match map.get(pubkey_hex) {
+        Some(p) if matches!(p.kind, PendingRequestKind::Restore) => map.remove(pubkey_hex),
+        _ => None,
+    }
+}
+
 fn take_matching_request(trade_pubkey_hex: &str, got: Option<u64>) -> Option<PendingRequest> {
     let mut map = pending_requests().lock().ok()?;
     match map.get(trade_pubkey_hex) {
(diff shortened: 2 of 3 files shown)


codaMW:
Both addressed:

Moved the privacy-mode guard before load_identity_from_mnemonic good catch. The identity was being swapped before the bail, leaving a Full-Privacy user in a mutated state, and it violated #215's "reject before any network traffic" criterion.
Restore dispatch now handles DaemonReply::Rejected by logging reason message and returning the message, matching the existing create_order/take_order rejection arms instead of dropping the daemon's rejection detail.

rust/src/api/orders.rs
```diff
@@ -124,6 +132,18 @@ fn request_id_matches(expected: u64, got: Option<u64>) -> bool {
 /// in place: relays can replay historical events, and a stale reply must not
 /// confirm, reject, or reconcile a live request — the genuine reply (carrying
 /// the nonce) arrives later and finds the record.
+/// Remove and return the pending RESTORE request for `pubkey_hex`. Unlike
+/// `take_matching_request`, there is no request_id gate: the RestoreSession
+/// message carries no request_id, so the daemon's RestoreData reply is
+/// correlated purely by the identity pubkey it is addressed to.
+fn take_matching_restore(pubkey_hex: &str) -> Option<PendingRequest> {
```
grunch, line 134: ⚠️ **This doc block now documents the wrong function.**

`take_matching_restore` was inserted directly beneath the doc comment that belongs to `take_matching_request`, with no blank line between them. That block ends:

> …relays can replay historical events, and a stale reply must not confirm, reject, or reconcile a live request — the genuine reply (carrying the nonce) arrives later and finds the record.

It is now attached to `take_matching_restore` — the one function in this file that deliberately has **no** nonce gate — while `take_matching_request`, whose rationale it actually describes, is left with no documentation at all. Precisely inverted, and the two functions sit adjacent so the next reader has every reason to trust it.

Add a blank line after the original block and keep the new doc with the new function.

rust/src/api/orders.rs
```diff
@@ -53,6 +53,10 @@ enum DaemonReply {
     Acknowledged,
     /// Daemon rejected the request with a CantDo reason.
     Rejected { reason: String, message: String },
+    /// Daemon replied to a RestoreSession with the user's active trades and
+    /// disputes. Correlated by identity pubkey (RestoreSession carries no
```
grunch: ⚠️ **"identity pubkey" is wrong here — correlation is by TRADE pubkey.**

Three new comments claim identity-pubkey correlation:

- this one, on `DaemonReply::Restored`;
- `PendingRequestKind::Restore` — "Correlated by identity pubkey, not request_id";
- `take_matching_restore`'s own doc — "correlated purely by the identity pubkey it is addressed to".

The code does the opposite, and `restore_session()`'s doc gets it right:

> the request is sent from a fresh TRADE key (event.sender) … the daemon looks up trades by identity/master key and replies to the trade key … so we subscribe on the trade key and correlate the reply by that pubkey

The map is keyed by `trade_pk_hex`, and the dispatcher looks up `take_matching_restore(trade_pubkey_hex)`. The identity key's role is the *seal* — it is what the daemon uses to find the trades, not what the reply is addressed to. Conflating the two is the kind of thing that sends the next person debugging correlation down the wrong path for an afternoon.

rust/src/mostro/actions.rs
```diff
@@ -459,3 +459,22 @@ mod tests {
...
     }
 }
+
+/// Build a `RestoreSession` request (mostro-core `Message::new_restore`).
+///
+/// Payload MUST be `None` — the daemon rejects any other payload
+/// (`MessageKind::verify`). The Seal is signed by the identity key (used by
+/// the daemon to look up the user's trades); the rumor is signed by a fresh
+/// trade key, and the daemon's restore reply is addressed to that trade key.
+/// Returns the NIP-59 Gift Wrap event JSON.
```
grunch: ⚠️ **This is not NIP-59 Gift Wrap — it is transport v2.**

`wrap_message` delegates to `gift_wrap::wrap_mostro_message`, documented at `rust/src/nostr/gift_wrap.rs:37` as:

> Wrap a `Message` destined for `receiver` (typically the Mostro node) as a **protocol-v2 NIP-44 direct event (signed Kind 14, authored by the trade key)**.

Restore is a daemon message, so v2 is the correct transport and the code is right — only the comment is wrong. NIP-59 / Kind 1059 is peer and dispute chat.

The pre-existing doc on `wrap_message` (line 341) carries the same error, and line 31 repeats it. Since this PR quotes the claim into a new function, worth correcting all three rather than propagating it further.

```suggestion
/// Returns the transport-v2 (NIP-44, signed Kind 14) event JSON.
```

rust/src/api/identity.rs
```diff
@@ -164,8 +164,22 @@ pub async fn load_identity_from_mnemonic(
...
+    let privacy_mode = crate::api::reputation::get_privacy_mode();
+    // Reject privacy-mode recovery BEFORE loading — otherwise the identity is
+    // already swapped when we bail, leaving the user in a mutated state, and it
+    // violates the "reject before any network traffic" contract for restore.
+    if recover && privacy_mode {
+        bail!("PrivacyModeRecoveryUnavailable");
+    }
+    let info = load_identity_from_mnemonic(words, 0, privacy_mode, None).await?;
+    if recover {
+        crate::api::orders::restore_session().await?;
```
grunch, line 186: ⚠️ **Restore failure propagates after the identity has already been swapped.**

You moved the privacy-mode guard ahead of `load_identity_from_mnemonic` for exactly the right reason, stated in the comment two lines up:

> otherwise the identity is already swapped when we bail, leaving the user in a mutated state

This line has the same problem and did not get the same treatment. By the time it runs, `load_identity_from_mnemonic` has succeeded and the identity is loaded. A daemon that is slow, unreachable, or simply busy makes `restore_session()` return `Err("NoDaemonResponse")` after the 10s timeout, `?` propagates it, and `import_from_mnemonic` reports failure to the caller — while having fully swapped the identity as a side effect. The caller has no way to tell "import failed" from "import succeeded, recovery didn't".

Not reachable today (`identity_service.dart:71` passes `recover: false`), which is exactly what makes it a trap: it will surface when #220 wires the UI and starts passing `true`, far from this line.

Recovery is best-effort relative to the import — the import genuinely did succeed. Suggest logging the failure and returning `Ok(info)`, with recovery state reported separately, or returning a result that distinguishes the two outcomes. This overlaps #219, which will restructure the waiting anyway; a comment pointing there would do for now if you would rather not decide it here.

grunch (changes requested):

Scope discipline is good — this really is the handshake and nothing else, and moving the privacy-mode guard ahead of `load_identity_from_mnemonic` is the right fix from the #204 round. Three things block, and the first is the one worth a design call before merge.

## Verification performed on `b982d89`

| Check | Result |
|---|---|
| `cargo test --lib` | ✅ 110 passed, 0 failed, **8 ignored** — matches the description |
| `cargo clippy --all-targets` | ❌ **new warning introduced** — see blocker 3 |
| Is the flow user-reachable? | ✅ **no** — `identity_service.dart:71` is the only Dart caller and passes `recover: false` |
| Is the transport correct? | ✅ yes — `wrap_message` → `gift_wrap::wrap_mostro_message`, i.e. protocol-v2 NIP-44 signed Kind 14, which is right for a daemon message. The doc comment says otherwise — see inline |
| Does `derive_trade_key()` touch persistent state? | ❌ **yes, it writes to the DB** — see blocker 1 |

## Blocking

### 1. `restore_session()` consumes and persists a trade-key index — acceptance criterion 2 is checked but false — severity: **high**

The description states, ticked:

> - [x] No DB writes, no key-counter changes, no `is_mine` mutation

`restore_session()` opens with `derive_trade_key()`, and that function (`rust/src/api/identity.rs:245-271`) increments `state.identity_info.trade_key_index` **and immediately persists it**:

```rust
state.identity_info.trade_key_index = candidate_index;
// Persist immediately: an index is consumed the moment it is derived.
db.save_identity(&state.identity_info).await...
```

So every restore attempt writes the identity row and burns an index — including attempts that time out. That rationale is sound for order flows, where the daemon registers every index it sees. Restore is different: it is a read-only query, and nothing binds the consumed index to a trade.

The part that needs a decision before this merges: **#217 exists to raise `trade_key_index` to the maximum recovered index**, and #218 builds reconstruction on top of that counter. Having the restore handshake itself advance the counter by one, for reasons unrelated to any trade, is state that both of those issues will have to reason about. Either:

- accept it and say so explicitly in #217 (the resync must be `max(recovered, current)`, never a plain assignment — worth stating since a naive implementation would rewind), or
- draw the restore key from outside the trade-index sequence.

Either way the checkbox and the "no key-counter changes" claim need to come out of the description. Flagging as blocking because two downstream issues inherit the answer, not because the code is wrong on its own.

### 2. No CI-runnable test covers any of the new logic — severity: **high**

Both new tests are `#[ignore]`d behind a live regtest stack, so CI exercises none of this — `cargo test --lib` reports 110 passed with 8 ignored, exactly as before the PR in terms of what actually runs.

Two pieces are trivially testable without a daemon, and the file already shows how:

- **`actions::restore_session()`** — the test right above it, `new_order_carries_request_id_and_trade_index`, builds keys with `Keys::generate()`, wraps, and asserts on the resulting `Message`. The same shape asserts the contract this PR documents as load-bearing: payload is `None` (the daemon rejects anything else in `MessageKind::verify`), seal signed by identity, rumor by the trade key.
- **`take_matching_restore()`** — a pure operation over the pending map: returns the record for a `Restore` entry, returns `None` for a non-`Restore` kind on the same key, and removes on take.

The nonce-gate asymmetry is the single subtlest thing in this PR, and right now nothing pins it.

### 3. `cargo clippy` no longer passes cleanly; "Build clean" is inaccurate — severity: **medium**

```
warning: items after test module
   --> src/mostro/actions.rs:470
   = note: `#[warn(clippy::items_after_test_module)]` on by default
   = help: move the items to before the test module was defined
```

Both new `restore_session` functions are appended *after* `#[cfg(test)] mod tests`, in `actions.rs` and in `orders.rs`. I checked the rest of the output so this isn't a false alarm: the other warnings are all `await_holding_lock` in `settings.rs` and `reputation.rs`, pre-existing and unrelated. This one is new, and `cargo clippy` is a mandated gate in this repo.

## Non-blocking

| Severity | Finding | Where |
|---|---|---|
| Medium | An existing doc block now documents the wrong function | inline |
| Medium | Three new comments say "identity pubkey"; correlation is by **trade** pubkey | inline |
| Medium | `restore_session()` doc claims NIP-59 Gift Wrap; it is transport v2 Kind 14 | inline |
| Medium | Restore failure propagates *after* the identity was already loaded | inline |
| Medium | Poisoned-mutex path registers nothing, publishes anyway, then waits the full 10s | `orders.rs:3407` |
| Low | A timed-out restore leaks a pending-map entry permanently | `orders.rs:193-201` |
| Low | `&trade_pubkey_hex[..8]` byte-slices a `&str` — panics under 8 chars, in the branch reached by *unexpected* input | `orders.rs`, the "no waiting caller" arm |
| Low | Both E2E tests hardcode the regtest pubkey and mutate global state (`initialize`, `set_active_mostro_pubkey`); un-ignoring both would race in one process | `restore_e2e_tests` |
| Low | `orders.rs` is now ~3530 lines; #120 tracks the split | — |

On the poisoned-mutex one: `if let Ok(mut map) = pending_requests().lock() { map.insert(...) }` silently skips registration on `Err`, then publishes anyway. The caller then waits the full 10 seconds for a reply that can never be correlated and reports `NoDaemonResponse` — a lock bug wearing a network bug's clothing. Bail instead.

Yes to making the regtest pubkey env-configurable, as you offered, and a shared setup helper would remove the duplicated 8-line preamble.

## Also worth a look

`import_from_mnemonic` now passes `crate::api::reputation::get_privacy_mode()` into `load_identity_from_mnemonic` where it previously passed a hardcoded `false`. That changes how an imported identity is loaded **even when `recover == false`** — which is the only path Dart uses today (`identity_service.dart:71`). It looks like a genuine fix rather than an accident, but it is behaviour change outside the stated scope of a handshake PR, and no test pins it. Worth either calling it out in the description or splitting it.

codaMW:
Good catch added assert_eq!(unwrapped.identity, identity_keys.public_key()) so the test now covers both signing roles: rumor by trade key (daemon's reply target) and Seal by identity key (how the daemon locates the trades).

codaMW:
@grunch addressed the full review:

**Blockers**
1. **Trade-key index** you're right, restore consumes an index via
   `derive_trade_key()`. Removed the false "no key-counter changes" checkbox,
   documented the behaviour in the description, and left a note on #217 that the
   resync must be `max(recovered, current)`. Kept the handshake as-is drawing
   keys outside the sequence is counter-design that belongs in #217/#218 rather
   than this handshake PR.
2. **Tests** added two CI-runnable unit tests: the `restore_session` builder
   (payload `None`, action `RestoreSession`, rumor by the trade key, and per
   CodeRabbit Seal by the identity key, covering both signing roles) and
   `take_matching_restore` (returns/removes RESTORE, ignores other kinds pins
   the nonce-gate asymmetry). 112 pass, CI green.
3. **clippy** moved both `restore_session` fns before their test modules;
   `items_after_test_module` is gone.

**Non-blocking**
- Poisoned pending-requests lock now bails before publishing.
- Guarded both `[..8]` pubkey slices in the RestoreData dispatch arms.
- Fixed the identity -> trade pubkey comments (3), the NIP-59 -> transport-v2 wording
  (all refs in actions.rs), and restored the `take_matching_request` doc block to
  its own function.
- Added a #219 note on the identity-already-swapped propagation on the recover path.

On `import_from_mnemonic` sourcing the real `privacy_mode` intentional (the
hardcoded `false` meant the recovery guard could never fire). Flagged in the
description; happy to split it if you'd prefer.

codaMW:
Also addressed a restore-rejection correlation bug CodeRabbit surfaced on the stacked #217: the Action::CantDo handler was nonce-gated, so a rejected RestoreSession (which is nonce-less) waited the full 10s and returned NoDaemonResponse instead of the reason. Now tries take_matching_restore first  restore rejections surface their reason, order requests keep their nonce gate and replay protection. Regression test added (take_matching_restore_matches_restore_records_only).

Issue #233, opened 2026-07-25
Web: IndexedDB storage backend is a stub — nothing persists across a reload


grunch:
## Summary

`rust/src/db/indexeddb.rs` is a compile-only stub. Every method either returns
`Err("IndexedDB not yet implemented")` or silently succeeds with a no-op, so on
the **web build nothing persists**: orders, trades, chat messages, relays, the
identity, the outbox queue, trade-key indices and the active Mostro node are all
lost on reload. The file says as much in its header ("Full implementation is
deferred to Phase 3").

This is not a regression — it has been true since the WASM target was added —
but it is worth tracking on its own, because new storage keys keep landing on
top of it and each one silently becomes native-only.

## Current behaviour, method by method

| Group | Web behaviour today |
|---|---|
| `save/get/delete/list_order`, `save/get/list_trade`, `save/list_message`, `mark_messages_read`, `save/delete/list_relay`, `save/get/delete_identity`, queued messages | return `Err("IndexedDB not yet implemented")` |
| `save/get/delete/clear_trade_key`, `get_order_id_by_trade_index`, `get_trade_by_order_id`, `update_trade_order_id`, `update_trade_fields` | no-op `Ok(())` / `Ok(None)` |
| `save/get_active_mostro_pubkey` | no-op + `log::warn!` — node selection does not survive a reload |
| `get/set/delete_setting` (added by the Cashu C1b PR) | no-op + `log::warn!` — the escrow-mode / mint-URL overrides do not survive a reload |

## Features deferred on top of this stub

Each of these works on native and is deliberately kept on a Dart-side store on
web until this issue lands:

- `backup_confirmed` (identity blob, #266) — the Dart notifier keeps
  SharedPreferences authoritative on web, because `save_identity` errors out
  here and `initDb` is `!kIsWeb`-guarded in `app_bootstrap.dart`. Tracked in #403.

The no-op variants are the dangerous ones: callers cannot distinguish "nothing
stored" from "storage does not work", so failures surface far from the cause
(e.g. a trade key that cannot be re-derived after a reload mid-protocol).

## Why it was left this way in C1b

The Cashu C1b PR adds a generic key/value accessor to the `Storage` trait
(`get_setting` / `set_setting` / `delete_setting`, backed by the existing
`settings` table on SQLite). Implementing it for real on IndexedDB while orders,
trades and messages still error out would leave the web build in an inconsistent
state — some data persisted, most not — so it follows the existing stub pattern
[...]

PR #234, opened 2026-07-25
feat(cashu): C1b — persist escrow-mode overrides and surface the backend in About

grunch:
Closes the **C1** phase of `docs/cashu/README.md`. C1a (#230) added the tri-state
detection and the `is_cashu_mode()` gate; this half adds persistence, the bridge
surface, and the About screen.

**Zero behaviour change on Lightning.** With no `escrow_mode` tag — every daemon
in the wild today — the resolution stays `Unknown`, every Cashu path stays shut,
and About renders exactly what it rendered before this PR.

## Rust

**Settings k/v store.** `Storage` gains `get_setting` / `set_setting` /
`delete_setting` over the `settings` table that already existed in the schema.
SQLite implements it for real, and the two named active-node accessors become
thin wrappers over it rather than a second copy of the SQL. IndexedDB follows
its existing stub contract (warn + no-op), so on web the overrides apply for the
session only — the whole IndexedDB backend is a stub, now tracked in #233.

**Resolution on read.** `mostro/escrow_mode.rs` previously cached the *resolved*
value. It now keeps what the node advertised and the developer overrides in
separate globals and runs `resolve()` on read. Two consequences worth the
change:

- flipping an override takes effect immediately instead of waiting for the next
  relay fetch, and there is no second copy of the answer that can go stale;
- a node switch clears the tags but keeps the overrides. Forcing Cashu is a
  statement about this build, not about a particular node.

Every mutator emits on a broadcast channel, so "changed" and "notified" cannot
drift apart.

**`api/escrow.rs`** (new): `get_escrow_mode`, `set_escrow_mode_override`,
`set_cashu_mint_url_override`, `rehydrate_escrow_overrides`, and an
`on_escrow_mode_changed` stream. Mint URLs must be `http(s)` with a host —
validated on write and *again* on rehydrate, so a stored value that no longer
passes tightened rules is dropped rather than silently used.
`EscrowModeInfo.mode` is a stable marker (`unknown` / `lightning` / `cashu`);
Rust does not translate.

The gate stays `is_cashu_available`, not `mode == "cashu"`: a node can advertise
Cashu and publish no usable mint, and that combination must not open anything.

## Dart

- `MostroInstance.fromTags` parses `escrow_mode` and the three `cashu_*` tags,
  tri-state and parameter-gated exactly like `BondPolicy` — a Lightning node
  carrying a stale `cashu_mint_url` never surfaces it as live.
- **About reports what the node advertises**, so the backends are mutually
  exclusive on screen: a Cashu node gets a "Cashu escrow" section (mint,
  locktime, settlement margin) *instead of* the Lightning Network section. A
  Cashu node that published no mint says "Not advertised" rather than rendering
  an empty section. The mint renders through the plain info row, not the
  copyable one, whose 20-character abbreviation would hide the host — the one
  part of a mint URL that matters.
- The client-side override never changes what About says. It is surfaced only in
  a `kDebugMode`-only settings card, which shows the effective resolution next
  to the toggle, so a tester is never fooled about which is which.
- `escrowModeProvider` / `isCashuAvailableProvider` expose the resolution that
  gates behaviour, for the phases stacked on top of this one.
- Strings translated in all five locales.

## Tests

**Rust** — k/v round-trip and key independence; override re-resolves without a
fetch; a node switch keeps the override; every mutator notifies subscribers; a
rejected mint URL leaves the previous value in place; forcing Cashu without a
mint does not open the gate.

**Dart** — tag parsing including gating, case/whitespace, unrecognised backends
and malformed day counts; widget tests asserting **no trace of Cashu** on a
Lightning or silent node (including one carrying stale cashu tags), and no
Lightning section on a Cashu one.

## Verification

```
cargo test                                  145 passed, 6 ignored
flutter test                                155 passed
```

`./scripts/frb-generate.sh` was run after the `rust/src/api/` change.

## Test plan

- [ ] Against any current daemon: About is identical to `main`, no Cashu strings
      anywhere, trades behave as before.
- [ ] Debug build → Settings → "Escrow backend (developer)": toggling "Force
      Cashu escrow" flips the effective mode instantly, with no relay round trip.
- [ ] Entering a garbage mint URL is rejected with a localized message and the
      previous value survives.
- [ ] Restart the app: the override and mint URL are still in force (native).
- [ ] Switch Mostro nodes: the override survives, the node's advertised tags do
      not.
- [ ] Release build: the developer card is absent.

Follow-ups: #233 (IndexedDB persistence on web). Next in the plan is the `cdk`
spike, before C2.

## Summary by CodeRabbit

- **New Features**
  - Updated the About screen to show the escrow backend as **Lightning** or **Cashu** (including mint URL, locktime, and settlement margin) and hide the non-matching section.
  - Added a **debug-only** settings card to preview the effective escrow backend, force **Cashu**, override the Cashu mint URL, and display availability/warnings (with “not advertised” messaging when applicable).
- **Documentation**
  - Updated Cashu escrow docs to clarify escrow-mode reporting and override persistence (persisted on desktop databases; session-only on web).
- **Localization / Tests**
  - Added Cashu escrow translations across supported languages and expanded UI/widget coverage to match the new rendering.

---

# Manual verification

Two things a reviewer needs to be sure of: **nothing changed for a Lightning
user**, and **the override does what it claims without leaking into About**.
No public daemon runs Cashu yet, so the Cashu-side rendering is covered by the
widget tests rather than by hand.

## Already run on this branch

```
cargo test                                  149 passed, 6 ignored
flutter test                                155 passed
```

## A · Lightning regression — the part that must not move

Run against the default node (`flutter run -d linux`, or your usual device).

1. **About screen** — open it and compare against `main`:
   - the **Lightning Network** section is present with the `lnd_*` values;
   - there is **no** "Cashu escrow" section and **no** row containing the word
     Cashu, a mint URL, a locktime or a settlement margin;
   - every other section (General info, Anti-abuse bond, Technical details) is
     byte-identical to `main`.
2. **Take and complete a trade** as you normally would. Nothing in this PR
   touches the order or trade flow; if anything behaves differently, that is a
   bug in this PR.
3. **Logs** (Settings → Log report) should show
   `[escrow-mode] active node advertises unknown` **at most once** per node,
   not once per reconnect.

## B · The developer card

The card is `kDebugMode`-only, so it only exists in a debug build.

4. **Debug build**: Settings → scroll to the bottom → **"Escrow backend
   (developer)"** is there, below "Mostro node".
5. **Release build**: `flutter build linux --release` (or your platform) and
   open Settings → the card is **absent**. This is the check that matters most
   — it is the one thing `flutter test` cannot verify, because `kDebugMode` is
   true under the test harness.

## C · The override behaves

In the debug build, on the developer card:

6. Toggle **Force Cashu escrow** on. "Effective backend" flips to **Cashu**
   *immediately* — no relay round trip, no app restart. This is the whole point
   of resolving on read.
7. With the toggle on and no mint set, the card shows the red
   "Cashu cannot run without a mint" line. That is the gate refusing to open,
   which is correct.
8. **Open About again.** It must still say Lightning and show the Lightning
   Network section. About reports what the *node* advertised; the override is a
   client-side statement and must never change it. If About flips, that is the
   regression this contract exists to prevent.
9. Type `not a url` in **Mint URL override** and apply → a localized error, and
   the previous value is still in the field.
10. Type `http://localhost:3338` and apply → accepted, and "Effective mint"
    updates.
11. **Kill the app and relaunch** → both the toggle and the mint URL are still
    set. (Native only; on web they last the session — #233.)
12. Switch to a different Mostro node in Settings → the override **survives**
    (it is a statement about this build, not about a node), while the node's own
    advertised mode is re-fetched.
13. Turn the toggle off → "Effective backend" returns to what the node says.

## What is not covered by hand

The Cashu-mode About rendering — the "Cashu escrow" section replacing the
Lightning one — has no daemon to produce it yet. It is covered by
`test/features/about/screens/about_screen_test.dart`, which asserts both
directions: no Cashu string on a Lightning or silent node, and no Lightning
section on a Cashu one.


grunch:
## Review — C1b (strict pass)

> **Note:** GitHub does not allow *Request changes* on your own PR, so this is posted as a comment. Treat it as **🔴 Request changes** — one finding is a real Flutter anti-pattern in a screen that writes persisted state, and one is a concurrency bug that will bite the moment two overrides are written close together.

**Scope reviewed:** `rust/src/db/{mod,sqlite,indexeddb}.rs`, `rust/src/mostro/escrow_mode.rs`, `rust/src/api/escrow.rs`, `rust/src/api/types.rs`, `lib/features/about/**`, `lib/features/settings/**`, l10n ×5.

---

### 🔴 Major

**1. Read-modify-write race on the overrides** — `rust/src/api/escrow.rs`

Both setters do the same thing:

```rust
escrow_mode::set_overrides(EscrowOverrides {
    mode,
    ..escrow_mode::get_overrides()   // ← read
});                                   // ← write
```

`get_overrides()` releases its lock before `set_overrides()` takes the write lock. Two concurrent calls — the dev card can produce exactly this by toggling the switch while the mint-URL field is submitted — interleave as read/read/write/write and the second write silently discards the first. The persisted value and the in-memory value then disagree until the next rehydrate.

Fix: give `escrow_mode` a `update_overrides(impl FnOnce(&mut EscrowOverrides))` that mutates under a single write lock, and have both setters go through it. The persist call should stay outside the lock.

**2. `TextEditingController` mutated during `build()`** — `lib/features/settings/widgets/escrow_mode_dev_card.dart`

```dart
final info = ref.watch(escrowModeProvider).valueOrNull;
if (info != null) {
  _syncMintField(info.mintUrlOverride);   // assigns _mintController.text
}
```

Assigning `.text` notifies listeners, which marks the `TextField` dirty *during* the build that is currently laying it out. It happens to work today, but it is the standard cause of "setState() or markNeedsBuild() called during build" once anything else in the subtree listens to the controller — and it will also stomp on the user's in-progress typing whenever an unrelated escrow event fires (a node switch, a capability re-fetch), because the guard only compares against the last *stored* value.

Fix: move it to a listener —

```dart
ref.listen(escrowModeProvider, (_, next) {
  final stored = next.valueOrNull?.mintUrlOverride;
  if (stored != _syncedMintOverride) { ... }
});
```

---

### 🟡 Minor

**3. `is_overridden` conflates "forced" with "genuinely Cashu"** — `escrow_mode.rs`, `resolve()`

`is_overridden` is `true` whenever the override is `ForceCashu`, even when the node independently advertises Cashu. The dev card then labels a perfectly genuine Cashu node as forced, which is precisely the confusion the flag exists to prevent. Consider `overridden && from_tags != EscrowMode::Cashu`.

**4. The broadcast payload is computed and thrown away** — `escrow_mode.rs` / `api/escrow.rs`

`notify()` builds a full `ResolvedEscrowMode` on every event, and `EscrowModeStream::next()` discards it (`Ok(_) => Ok(snapshot())`). The rebuild is deliberate and well-argued in the comment, but then the channel should be `broadcast::Sender<()>` — as written, every notification does the resolution work twice.

**5. `clear()` now emits on every failed capability fetch** — `api/nostr.rs`

`clear()` notifies, and it is called on both the `Ok(None)` and `Err` arms of `fetch_and_set_node_capabilities`. On a flaky relay this produces a stream of identical "still unknown" events. Harmless today (the UI is idempotent), but worth suppressing when the state did not actually change.

---

### 🔵 Nit

**6. Inconsistent stubs** — `rust/src/db/indexeddb.rs`: `set_setting` warns, `delete_setting` returns `Ok(())` silently. Both are no-ops; either both should warn or the comment should say why removal is different.

---

### ✅ What is right

- The move from caching the *resolved* value to resolving on read is the correct call, and the reasoning in the doc comment (no second copy that can go stale) is the right one to record.
- Keeping the override across a node switch while clearing the tags is subtle and correctly justified.
- The About decision — report what the node advertises, never what the override forces — is the honest one, and the widget tests actually assert the absence of every Cashu string rather than just the presence of the right ones.
- `EscrowModeOverride::from_stored` failing closed on garbage is the right default for a flag that unlocks a money path.

### Test gaps

- Nothing covers the persistence path end to end (`rehydrate_escrow_overrides` after a real `set_*`), only the pieces.
- No test asserts the dev card is absent from a release build. `kDebugMode` is `true` under `flutter test`, so this is untestable as written — worth a comment saying so, rather than leaving the reader to assume it is covered.

codaMW (dismissed):
Reviewed C1b independently, full pass over `escrow_mode.rs`, `api/escrow.rs`,
`db/*`, the About rendering, and the dev card, on top of your strict round and
the CodeRabbit rounds. Approving.

**Both Majors verified fixed:**
- Override race, both setters now go through `update_overrides(impl FnOnce)`
  under a single write lock; no read-then-write window.
- Controller-during-build  `_syncMintField` moved out of `build()` into
  `ref.listen`.

**Money-path safety (the part that must not move), fail-safe by construction:**
- No `escrow_mode` tag -> `Unknown` -> About keeps the Lightning section it always
  showed.
- The gate is `is_cashu_usable` (mode Cashu **and** a non-empty mint), not
  `mode == cashu` a Cashu node with no usable mint opens nothing.
- About branches on `node.escrowMode` (node-advertised), never the client
  override; the widget test asserts the *absence* of every Cashu string on a
  legacy/silent node, not just the presence of the right ones.
- `validate_mint_url` requires http(s)+host on write and drops non-conforming
  stored values on rehydrate fails closed.
- The `Err`/`Ok(None)` fetch arms both `clear()` to fail closed on an unreachable
  node, with PoW deliberately left alone the reasoning in those comments is
  the right call.

**Your earlier minors all confirmed addressed:** #3 (is_overridden no longer
conflates forced with genuinely-Cashu), #4 (broadcast is now `Sender<()>`), #5
(`clear()` only notifies when the cached state actually changed), #6 (both
IndexedDB stubs warn).

**One heads-up (non-blocking):** the 3 `escrow_mode_dev_card_test` cases fail on
my local Flutter 3.44 with a `ListTile background color or ink splashes may be
invisible` assertion the `ListTile` is inside a `DecoratedBox` with a bg color.
CI passes them on the pinned 3.38.2, so it's a version-strictness gap today, but
that pattern will start failing once `FLUTTER_VERSION` is bumped. Wrapping that
`ListTile` in its own `Material` (or dropping the intermediate `DecoratedBox`
background) now would future-proof it. Flagging so it isn't a surprise later.

`cargo test` 149 passed / clippy `-D warnings` clean / `cargo check
--target wasm32-unknown-unknown` clean / `flutter analyze` clean. The
resolve-on-read design and the honest About contract (report what the node
advertises, never what the override forces) are the right calls.

AndreaDiazCorreia (dismissed):
tACK

BraCR10 (dismissed):
tACK

PR #236, opened 2026-07-25
feat(cashu): C4 — escrow primitives (2-of-3 NUT-11 lock, sign, redeem, reclaim)

grunch:
Phase **C4** of `docs/cashu/README.md` — the cryptographic heart of Cashu-mode
trading. No UI, no bridge surface, so a review can be about correctness alone.

> **Stacked on #235 (C2).** Base is `feat/cashu-c2-wallet-core`; GitHub will
> retarget to `main` when C2 merges. The escrow operations hang off the wallet
> that phase introduced.

## What it does

The escrow is one NUT-11 P2PK secret per proof, locked 2-of-3 (§2 of the doc):

```
data          = P_S                (seller)
pubkeys       = [P_B, P_M]         (buyer, Mostro)
n_sigs        = 2                  (any two of the three)
sigflag       = SIG_INPUTS
locktime      = now + escrow_locktime_days
refund        = [P_S]              (seller alone, after locktime)
n_sigs_refund = 1
```

- **`xonly_to_cashu_pubkey`** — the `02`-prefix mapping the daemon applies in
  `cashu_pubkey_from_xonly_hex`. Anything that is not exactly 64 hex characters
  is rejected rather than padded or truncated: a silently reinterpreted key
  locks funds to nobody, and nothing downstream would notice.
- **`escrow_conditions` / `fee_conditions`** — the 2-of-3 above, and a plain
  1-of-1 to Mostro for the fee. The fee is a payment, not an escrow; giving it
  the escrow's conditions would make it unspendable for the node.
- **`verify_escrow_token`** — the client-side mirror of the daemon's composite
  check, run **per proof**. A token whose first proof is correct and whose
  second is locked to the builder alone would pass a spot check and walk away
  with the difference. Also checks mint and total amount.
- **`sign_proofs` / `combine_and_redeem` / `reclaim_after_locktime`** — one
  signature per proof, keyed by that proof's secret, matching the wire form C0
  pinned. A missing peer signature fails before the mint is contacted, so a
  half-signed spend is never attempted.

The spike (`docs/cashu/cdk-spike.md`, in #235) established that cdk exposes
NUT-11 with custom tags in full — so none of this hand-builds secrets or witness
encodings, which is what kept the phase to this size.

## Tests

**Unit** (run in CI): the `02` mapping against a fresh key; five malformed-key
shapes rejected, including a 33-byte compressed key passed where x-only was
expected; the built condition asserted field by field against §2, so a wrong
default cannot hide behind a passing round trip; a past locktime refused at
construction; and verification refused for every way an escrow can be wrong —
wrong seller key, `n_sigs=1`, an extra refund key, a bare P2PK, a missing
counterparty, a locktime shorter than required.

**Integration** (`#[ignore]`, needs `MOSTRO_TEST_MINT_URL`) — the "done when" of
the phase:

- full lock → verify → sign → combine → redeem against a real mint;
- one signature alone cannot move an escrow (the seller's premature reclaim is
  refused by the mint);
- an impostor's signature does not settle one;
- a short-changed escrow fails verification before anything is submitted.

```
docker run -p 3338:3338 cashubtc/nutshell:latest poetry run mint
MOSTRO_TEST_MINT_URL=http://localhost:3338 cargo test -- --ignored
```

## Verification

```
cargo test                                  149 passed, 14 ignored
```

No `frb-generate` needed — this phase adds nothing under `rust/src/api/`.

## Test plan

- [ ] Run the ignored suite against a funded nutshell wallet; all four pass.
- [ ] Confirm the compressed keys this produces match the daemon's for the same
      x-only input (compare against `cashu_pubkey_from_xonly_hex` on the Track A
      branch).
- [ ] Nothing in a Lightning-mode build reaches this module — it has no callers
      until C5.

Next: C5 wires this into the take flow.

---

# Manual verification

This PR is pure cryptography with no UI, so the verification that matters is
the integration suite — and it now genuinely runs (it could not before; see the
last commit). The security properties are the point: **verify those four tests
fail if you break the code**, not just that they pass.

## Already run on this branch

```
cargo test                                  152 passed, 14 ignored
```

Plus the integration suite against a real mint — **8/8, twice in a row**
(nutshell 0.20.3).

## A · Integration suite against a live mint

```bash
docker run -d --name nutshell -p 3338:3338 \
  -e MINT_LISTEN_HOST=0.0.0.0 -e MINT_LISTEN_PORT=3338 \
  -e MINT_BACKEND_BOLT11_SAT=FakeWallet \
  -e MINT_PRIVATE_KEY=TEST_PRIVATE_KEY \
  -e MINT_RATE_LIMIT=FALSE \
  cashubtc/nutshell:latest poetry run mint

cd rust
MOSTRO_TEST_MINT_URL=http://localhost:3338 \
  cargo test --lib -- --ignored --test-threads=1 cashu::
```

Expect **8 passed**, and 8 again on a second run. `MINT_RATE_LIMIT=FALSE` is
required for the second run (nutshell rate limits by default).

Four of those eight are the security properties, against a real mint with real
blind signatures:

| Test | Proves |
|---|---|
| `an_escrow_locks_and_settles_with_two_of_three_signatures` | the happy path works end to end: lock → verify → seller signs → buyer combines and redeems |
| `one_signature_is_not_enough_to_move_an_escrow` | the seller cannot take the escrow back alone before the locktime |
| `a_signature_from_the_wrong_key_does_not_settle_an_escrow` | an impostor's signature plus a real one does not satisfy the 2-of-3 |
| `an_escrow_for_the_wrong_amount_fails_verification` | a short-changed escrow is caught before anything is submitted |

## B · Confirm the fix that this PR's review found

The verifier used to accept **extra** pubkeys, which let a seller add a key they
also control and spend the escrow alone. To convince yourself it is closed:

```bash
cd rust
cargo test --lib verification_rejects_every_way_an_escrow_can_be_wrong
```

Then break it deliberately — in `rust/src/cashu/escrow.rs`, delete the
`pubkeys.len() != 2` check — and re-run. The test must **fail** on the
"extra pubkey" row. Put it back.

Do the same with `three_parties_must_be_three_different_keys` by removing the
`ensure_distinct()` call in `escrow_conditions`.

## C · Key encoding against the daemon

The `02`-prefix mapping must match the daemon's `cashu_pubkey_from_xonly_hex`
exactly, or the escrow locks to a key nobody holds. `cargo test --lib
an_xonly_key_maps_to_the_daemons_compressed_form` covers the shape; if you have
the daemon's Track A branch checked out, feed the same x-only hex to both and
compare the output byte for byte.

## D · Nothing else moved

This PR adds a module and two crate-internal accessors on the wallet. It has
**no callers** until C5, so a Lightning build cannot reach any of it. The
regression check is simply that `cargo test` and the app behave as on the base
branch.

## Summary by CodeRabbit

- **New Features**
  - Added Cashu-based escrow support for trades using 2-of-3 spending conditions involving the buyer, seller, and Mostro.
  - Added fee token creation with authorization restricted to Mostro.
  - Added escrow token verification, multi-party signing, redemption, and seller refunds after the configured locktime.
  - Added validation for participant keys, amounts, signatures, locktimes, and token conditions.
- **Compatibility**
  - Cashu escrow functionality is available in native builds; WebAssembly builds remain unchanged.


grunch:
## Review — C4 (strict pass)

> **Note:** GitHub does not allow *Request changes* on your own PR, so this is posted as a comment. Treat it as **🔴 Request changes — blocking.** Finding 1 is a spendability bug in the verifier that the buyer's safety depends on, and the test suite in this PR asserts around it without catching it.

**Scope reviewed:** `rust/src/cashu/escrow.rs`, `rust/src/cashu/wallet.rs` (accessors), `rust/src/cashu/mod.rs`.

---

### 🛑 Critical

**1. `verify_conditions` accepts *extra* pubkeys — the 2-of-3 is not verified to be a 2-of-3** — `rust/src/cashu/escrow.rs:146-152`

```rust
if !pubkeys.contains(&parties.buyer) || !pubkeys.contains(&parties.mostro) {
    bail!("InvalidEscrowToken: buyer or Mostro key missing");
}
```

This checks the two keys are *present*. It never checks nothing else is.

A seller can build an escrow with `pubkeys = [P_B, P_M, P_attacker]`, where `P_attacker` is a key they also control. `data = P_S`, `n_sigs = 2`, so the condition is satisfied by **any two** of `{P_S, P_B, P_M, P_attacker}` — and the seller holds two of them. They can drain the escrow unilaterally the moment it is funded, and this verifier says the token is good.

Every other field is exact-matched, including `refund_keys` (`keys == &vec![parties.seller]`), which makes the omission look deliberate on a read. It should be the same shape:

```rust
if pubkeys.len() != 2
    || !pubkeys.contains(&parties.buyer)
    || !pubkeys.contains(&parties.mostro)
{
    bail!("InvalidEscrowToken: pubkeys must be exactly [buyer, mostro]");
}
```

Note `verification_rejects_every_way_an_escrow_can_be_wrong` covers the *missing*-key case and the extra-**refund**-key case — the extra **pubkey** case is the one hole in an otherwise thorough table, which is what let it through. Add it to that table.

---

### 🔴 Major

**2. No currency-unit check** — `verify_escrow_token()`

`token.value()` sums proof amounts irrespective of unit. A token denominated in something other than `sat` with the right numeric total passes `expected_amount`. `cdk`'s `receive` enforces the unit later, so this is not exploitable end to end today — but this function is documented as the pre-submit mirror of the daemon's check, and the daemon does bind the unit. Add `token.unit()` against `CurrencyUnit::Sat` next to the mint check.

**3. No distinctness check on the three parties** — `escrow_conditions()` / `verify_conditions()`

Nothing rejects `buyer == seller`, `mostro == seller`, or `buyer == mostro`. A degenerate set collapses the 2-of-3 into something weaker, and whether one signature can satisfy `n_sigs = 2` under a duplicated key is mint-implementation-defined — which is exactly the kind of thing not to leave to the mint. Reject at construction and in verification.

**4. Reserved-proof leak on a failed `confirm`** — `build_locked_token()`

Same defect as `create_token` in #235, and it matters more here: the amount is the whole escrow, not a user-chosen send. A `confirm` failure leaves the escrow amount reserved and invisible, and C5 then reports the wallet as short. Revoke on failure.

---

### 🟡 Minor

**5. `combine_and_redeem` matches peer signatures by first hit** — if a token ever carried two proofs with the same secret, the second would get the first's signature and the swap would fail at the mint with an opaque error. Secrets are unique in practice; a `HashMap<&str, &ProofSignature>` built once would be both clearer and O(n) instead of O(n²).

**6. `reclaim_after_locktime` cannot detect "too early" locally** — it signs and submits, and relies on the mint to refuse. The comment says so, which is honest, but the token's own `locktime` is right there and comparing it to `now` would turn a confusing mint error into a precise one.

---

### ✅ What is right

- **Per-proof verification** rather than per-token, with the reasoning stated (a token whose first proof is correct and whose second is not). That is the non-obvious half of this file and it is right.
- Rejecting a key that is not exactly 64 hex chars instead of padding or truncating — including the 33-byte compressed key passed where x-only was expected, which is the mistake a caller will actually make.
- `escrow_conditions` asserted field by field in the tests rather than through a round trip, so a wrong default cannot hide.
- Refusing a past locktime at construction: an already-refundable escrow is worthless to the buyer and should never exist.
- The integration tests test the *security* properties (one signature is not enough; an impostor's signature does not settle), not just the happy path.

### Test gaps

- The extra-pubkey case (finding 1) — the single most important missing row in `verification_rejects_every_way_an_escrow_can_be_wrong`.
- No unit-mismatch case.
- No degenerate-parties case.
- `sign_proofs` has no unit test at all; its output shape (one entry per proof, keyed by that proof's secret) is a wire contract with the daemon and is currently only exercised inside the ignored integration tests.

grunch:
@coderabbitai review

grunch:
@coderabbitai Review

21Mill (commented):
Reviewed in depth, with the integration suite actually running: nutshell 0.20.3 in Docker, tests built against cdk 0.17.3 as pinned in `Cargo.lock`. Your baseline reproduces — **8 passed**, including the four security properties. I read your strict pass first; all six of its findings are addressed in the tree.

The shape is right and the round-1 fix is real: `pubkeys.len() != 2` closes the extra-key hole exactly as described, `ensure_distinct()` on both the way in and the way out is the correct place for it, per-proof verification is the right granularity, and checking the unit before the amount is a genuinely non-obvious ordering that pays for itself. I also chased the duplicate-proof short-change variant — a token repeating one proof to inflate `value()` — and it is already closed upstream: `TokenV4::value()` returns `Error::DuplicateProofs` when secrets collide, and `verify_escrow_token` calls `value()` before it iterates. Worth knowing so it doesn't get re-raised.

One finding below is exploitable against a real mint, and I have a PoC. Two are minor.

## 🛑 `sign_proofs` is a signing oracle — demonstrated, 15 sat taken

`Proof::sign_p2pk` signs `self.secret.to_bytes()` and nothing else. Under `SIG_INPUTS` the signature commits to the secret string alone — not the amount, the keyset, the `C` value, or the token it arrived in. So a BIP-340 signature over secret `S` by key `K` is valid for **any** proof carrying secret `S`, wherever it appears. `sign_proofs` signs every proof in whatever token it is handed, and in the release and cooperative-cancel flows the counterparty is exactly who supplies that token.

Against a live mint:

1. Seller builds a genuine 16 sat escrow; buyer verifies it with `verify_escrow_token` — passes, as it should.
2. Buyer fabricates a decoy: the escrow's proofs with the **same secrets**, same keyset, amounts rewritten to 1 sat, witnesses stripped, re-encoded with a `"trivial refund, please sign"` memo. A different token by any comparison.
3. Seller calls `sign_proofs` on the decoy. They never touch the real escrow in this step.
4. Buyer replays those signatures against the real escrow with their own key: **two of three**.

```
[PoC] ORACLE OPEN — redeemed 15 sat with a harvested signature
```

15 sat of the 16 sat escrow (the missing sat is the mint's swap fee), no fiat paid. Reproduced twice.

### The fix, verified

I applied the change and re-ran everything:

```rust
pub async fn sign_proofs(
    &self,
    encoded: &str,
    key: SecretKey,
    parties: &EscrowParties,
    expected_amount: u64,
    min_locktime: u64,
) -> Result<Vec<ProofSignature>> {
    self.verify_escrow_token(encoded, parties, expected_amount, min_locktime)
        .await?;
    // … unchanged
```

The PoC then dies before any signature exists — `[PoC] signing refused: InvalidEscrowToken: expected 16 sat, got 1` — and your original 8 integration tests still pass. Cost: three call sites in tests, all of which already have `parties` and `locktime` in scope. There is no legitimate non-escrow use of this function (the fee token is 1-of-1 to Mostro and is redeemed by the node with its own key, not signed through here), so requiring the escrow's parameters is the honest signature rather than an added restriction. And it is free today precisely because the module has no callers until C5.

I'd rather see this than a documented precondition. A primitive whose misuse costs the escrow should not have an unsafe form for C6/C7 to forget about — C5 only locks, so `sign_proofs` gets its first real callers when release and cooperative cancel land.

## ⚠️ A residue the fix does not remove — and I don't think it can, here

I tested the fix against a smarter decoy: identical amounts and conditions, differing only in the memo. It passes `verify_escrow_token`, so the seller signs, and the signatures redeem the real escrow.

That is **not** an escalation of the bug above — the decoy is the same escrow with the same parties, amount and locktime, and the caller supplied those parameters themselves, so signing it is consenting to release that escrow. The fix holds.

But it exposes something worth a decision at the spec level, not in this PR. Under `SIG_INPUTS` the signed message is the input secrets only; under `SIG_ALL` it is `secret_0 || C_0 || … || amount_0 || B_0 || …`, which commits to the **outputs**. So with `SIG_INPUTS`, a signature authorises *spending the proof*, never *where the value goes*. A seller's release signature and a seller's cooperative-cancel signature over the same escrow are byte-identical in what they authorise: whoever holds one plus a second key can direct the funds either way.

§2 of `docs/cashu/README.md` pins `SIG_INPUTS`, so this is upstream of the PR. But C6 and C7 are where it becomes reachable — release and cooperative cancel are the two flows that produce signatures over the same escrow for opposite destinations, and they will need to guarantee that no party ever signs for both. Worth an issue against the spec doc so the constraint is written down before the flow that has to honour it gets built.

## Two smaller things

**`normalize_token` is bypassed.** `wallet.rs` strips `cashu:` / `cashu://` prefixes before parsing, with a comment explaining that QR payloads routinely carry them. All four escrow entry points use bare `encoded.trim().parse()` instead. A token the wallet's own receive path accepts will fail escrow verification with "unparseable", which is the wrong explanation. One line each.

**The finding-5 fix addressed the complexity but not the diagnosability.** The `HashMap` is the right structure and the O(n) half of the comment is correct, but it also says the index "makes a duplicate secret in the peer's list a visible collision rather than a silent first-wins" — `collect()` into a `HashMap` does neither, it silently keeps the **last** value for a repeated key. Your original concern stands where it was: a duplicate secret still produces no error here and still surfaces as an opaque mint failure later. Only the complexity changed. Either reject duplicates explicitly when building the index, or drop the clause — a comment asserting a safety property that isn't there is worse than none, because the next reader builds on it.

**`verify_escrow_token` is purely structural, and the doc comment doesn't say so.** It proves the token is *shaped* like the right escrow; it does not prove the proofs are unspent (no NUT-07 check) or mint-issued (cdk verifies DLEQ on receive — after the fiat leg). A seller can hand over a structurally perfect token whose proofs were spent an hour ago. If that check belongs to the daemon or to C5, fine, but the boundary should be stated on the function, because "returned Ok" reads as a much stronger claim than it is.

## Reproducing the PoC

Against the mint your own instructions describe:

```
docker run -d --name nutshell -p 3338:3338 \
  -e MINT_LISTEN_HOST=0.0.0.0 -e MINT_LISTEN_PORT=3338 \
  -e MINT_BACKEND_BOLT11_SAT=FakeWallet -e MINT_PRIVATE_KEY=TEST_PRIVATE_KEY \
  -e MINT_RATE_LIMIT=FALSE cashubtc/nutshell:latest poetry run mint
```

The test builds a genuine escrow, rewrites its proofs' amounts while keeping their secrets, re-encodes with `Token::new`, has the seller sign that, and feeds the resulting signatures to `combine_and_redeem` on the original token. I'm happy to open it as a PR against your branch if you'd like it as a regression test — it belongs next to `a_signature_from_the_wrong_key_does_not_settle_an_escrow`.

## Housekeeping

GitHub reports this as `CONFLICTING` against `main` — expected while stacked on #235, but it needs a rebase before it can land.

grunch:
@coderabbitai review

grunch:
## Review — C4, round 2 (strict pass, 2026-09-10)

> Same caveat as round 1: GitHub does not allow *Request changes* on your own PR. Treat this as **🔴 Request changes** as of the head this review was written against (`60453c2`), and as **✅ resolved** by the commits pushed with it — every finding below is fixed in the tree, and the integration suite was re-run against nutshell 0.20.3 to prove it.

**Scope:** `rust/src/cashu/escrow.rs`, `rust/src/cashu/wallet.rs`, `rust/src/cashu/mod.rs`, plus the drift between this branch and `main` (435 commits behind; C2 merged with a reworked wallet).

### Status of the earlier comments

| # | Finding (round 1, 2026-07-25) | Status at `60453c2` |
|---|---|---|
| 1 | extra pubkeys accepted | ✅ fixed (`pubkeys.len() != 2`), row added to the table |
| 2 | no unit check | ✅ fixed, and ordered before the amount |
| 3 | no distinctness check | ✅ fixed on construction and on verification |
| 4 | reserved-proof leak on failed `confirm` | ❌ **regressed by the rebase** — see finding B |
| 5 | first-hit signature match | ⚠️ complexity fixed, comment wrong — see finding D |
| 6 | reclaim cannot detect "too early" | ✅ fixed |

| 21Mill's review (2026-09-07) | Status at `60453c2` |
|---|---|
| 🛑 `sign_proofs` is a signing oracle (PoC, 15 sat taken) | ❌ **not addressed** — finding A |
| ⚠️ `SIG_INPUTS` residue (release vs. coop-cancel sign the same thing) | spec-level; noted in the doc, see "Not fixed here" |
| `normalize_token` bypassed | ❌ not addressed — finding C |
| HashMap comment claims a property it doesn't have | ❌ not addressed — finding D |
| `verify_escrow_token` is purely structural, doc doesn't say so | ❌ not addressed — finding E |

All three of 21Mill's smaller points and the critical one were still open. They are all closed now.

---

### 🛑 Critical

**A. `sign_proofs` signs whatever token it is handed** — `escrow.rs:325-345` at `60453c2`

Confirmed against cdk 0.17.3: `Proof::sign_p2pk` signs `self.secret.to_bytes()` and nothing else. Under `SIG_INPUTS` the signature commits to the secret string alone — not the amount, the keyset, `C`, or the token. A signature harvested over a decoy carrying the escrow's secrets with the amounts rewritten is therefore valid for the real escrow, and in release / coop-cancel the counterparty is exactly who supplies the token.

**Fix:** `sign_proofs` now takes `parties`, `expected_amount`, `min_locktime` and runs `verify_escrow_token` before producing a signature. There is no unchecked form. The three test call sites already had those values in scope. Regression test: `signing_refuses_a_decoy_that_reuses_the_escrows_secrets` (integration) builds the decoy the way 21Mill's PoC did — same secrets and keyset, every amount 1 sat, witnesses stripped, friendly memo — and asserts the refusal is on the amount, before any signature exists.

**B. The rebase silently breaks the build, and the fix it removes was the wrong fix** — `escrow.rs:build_locked_token`

The branch calls `self.check_proofs_state()` on a failed `confirm`. That method no longer exists on `main`: the C2 review round (`0794879`) replaced it with `sweep_spent_proofs` and established that cdk's state check **skips proofs an operation reserved** — so even at the old head, the "reclaimed N sat" this reported was never a reclaim of the escrow amount. Round-1 finding 4 was marked fixed, but the fix could not have worked.

`main`'s `create_token` now does the right thing: keep `operation_id` before `confirm` consumes the handle, call `revoke_send` on failure, and refuse to claim "reclaimed 0" when the revoke itself fails. That logic is subtle and must not be duplicated, so it moved into a shared `CashuWallet::send_with_conditions(amount, Option<SpendingConditions>)`; `create_token` passes `None`, the escrow passes the 2-of-3 or the fee lock. Markers are the same set for both (`CashuSendFailed` / `CashuSendUnresolved`) because it is the same send.

### 🔴 Major

**C. `normalize_token` is bypassed** — all four escrow entry points used bare `encoded.trim().parse()`. A `cashu:` URI the wallet's own receive path accepts failed here as "unparseable". Now one `parse_token` helper, using the wallet's `normalize_token`; unit test covers the three prefix shapes.

**D. The `HashMap` comment asserts a safety property the code does not have** — `collect()` into a map keeps the *last* entry for a repeated key, silently; the comment said "a visible collision". Now `index_signatures` inserts one by one and returns `DuplicatePeerSignature` on a repeat, **before** the mint is contacted. Two unit tests (index shape, duplicate rejected).

**E. `verify_escrow_token` proved shape only** — a structurally perfect token whose proofs were spent an hour ago passed, and a proof carrying no DLEQ passed (cdk skips those on receive; C2's review closed the same hole on `receive_token`). The buyer is the party who would find out last, after the fiat leg. Now: shape → `verify_token_dleq` (NUT-12) → `check_proofs_spent` (NUT-07), in that order so the cheapest and most diagnostic failure fires first. Markers: `CashuTokenUnverified`, `CashuEscrowSpent`, `CashuEscrowPending`, `CashuMintUnreachable`. Regression test: `a_spent_escrow_fails_verification` locks, redeems, presents the same token again. The doc comment now states exactly what `Ok` means and that it is a point-in-time statement.

### 🟡 Minor

- `CurrencyUnit` was referenced by full path inside the function; imported.
- `docs/cashu/README.md` §C4 updated to the new `sign_proofs` contract and the widened `verify_escrow_token` scope (spec is a living artifact).

### Not fixed here — needs a decision at the spec level

21Mill's `SIG_INPUTS` observation stands and is upstream of this PR: a release signature and a coop-cancel signature over the same escrow are byte-identical in what they authorise (spend the proofs), so whoever holds one plus a second key can direct the funds either way. The `sign_proofs` doc now says so, and §C4 of the doc carries the corollary for C6/C7 (no party ever signs for both). This deserves an issue against `docs/cashu/README.md` §2 before C6/C7 are built; not opened from this review.

### ✅ What is right (unchanged)

Per-proof verification, exact-match on every field of the condition, rejection of a 33-byte key where x-only was expected, a past locktime refused at construction, and integration tests that assert the security properties rather than only the happy path.

### Verification

Run on `c661410` (this branch, merged with `main` at `13e8a42`), rustc 1.98.0, cdk 0.17.3, nutshell 0.20.3 in Docker:

```
cargo test --locked                           439 passed, 24 ignored
cargo clippy --locked -- -D warnings          clean
cargo check --locked --target wasm32-unknown-unknown   clean
MOSTRO_TEST_MINT_URL=http://localhost:3338 cargo test --lib -- --ignored --test-threads=1 cashu::
                                              12 passed (6 escrow, 6 wallet), twice in a row
```

The two new integration tests are the ones to break deliberately: remove the `verify_escrow_token` call at the top of `sign_proofs` and `signing_refuses_a_decoy_that_reuses_the_escrows_secrets` must fail; remove the `check_proofs_spent` block and `a_spent_escrow_fails_verification` must fail.

Not a CI gate, but for the record: `cargo fmt --check` reports diffs across most of the crate on `main` already, so this branch does not try to be fmt-clean either.

**Commits:** `1cd8b0d` (merge + finding B, so the merge itself compiles) and `c661410` (findings A, C, D, E, docs).

PR #238, opened 2026-07-25
feat(cashu): C5 — seller escrow lock flow (Track A)

grunch:
Phase **C5** of `docs/cashu/README.md` — the Cashu replacement for "the seller
pays the hold invoice", and the first end-to-end flow.

> **Stacked.** Base `feat/cashu-c3-wallet-ui` (#237) → #236 → #235, with #234
> merged in. Merge order: #234, #235, #236, #237, then this.

## Read the daemon, didn't guess it

The plan flagged the escrow request's exact shape as unconfirmed. It is now
confirmed against the daemon's own branches (`feat/cashu-ta2-take-flow`,
`feat/cashu-ta1f-fee-token`):

- the request is `Action::WaitingSellerToPay` + `Payload::Order`, status
  `WaitingPayment`, both trade pubkeys, **no buyer invoice** — so it is
  classified by payload shape, not by a new action;
- the escrow token must be worth **exactly** `order.amount`;
- the fee token is `2 * order.fee`, where `order.fee = round(fee * amount / 2)`.
  The daemon rounds the *half*, so doubling the rounded half is the only
  expression that agrees with it — `round(fee * amount)` differs by a satoshi on
  half of all amounts, and every one of those would be rejected;
- the proof's stated pubkeys must equal the order's trade keys, which the daemon
  re-derives rather than trusting.

## Rust

- **`mostro/node_fee.rs`** — the node's advertised fee, read from the same 38385
  fetch that already yields PoW and the escrow mode, cleared on node switch. A
  malformed fee is discarded rather than stored: a wrong fee surfaces much later
  as an unexplained lock failure.
- **`mostro/actions.rs`** — `add_cashu_escrow`, beside `add_invoice`.
- **`api/cashu.rs`** — `cashu_escrow_quote` (amount, fee, total, balance, mint,
  locktime, all shown before the seller commits) and `lock_escrow`:
  1. refuse below `amount + fee`;
  2. build the escrow token, then **verify it with the same check the daemon
     runs** — catching a mistake here means the seller learns before publishing,
     not after;
  3. build the fee token when the node charges one;
  4. publish, then persist.

  Persistence deliberately follows the mint swap and happens **even if the
  publish fails**: the ecash is already committed by then, and a token we did
  not record is money we cannot find again. The daemon's handler is idempotent
  on re-submission, which is what makes that safe.
- **`TradeInfo`** gains `cashu_mint_url` / `cashu_escrow_token` /
  `cashu_locked_at`, `#[serde(default)]` so rows written by older builds load.
- A node switch now also drops the fee and disconnects the wallet, which was
  bound to the previous node's mint.

## Dart

`lock_escrow_screen.dart` — the Cashu sibling of the pay-invoice screen: escrow,
fee and total against the balance, the mint, and when the seller can reclaim
unilaterally if Mostro vanishes. A short balance offers "fund your wallet"
rather than a failure. The take flow branches on `isCashuAvailable`: the buyer
goes straight to the trade (there is no invoice step in Cashu mode), the seller
to the lock screen. Strings in all five locales; unknown markers fall back.

## A real bug this phase surfaced

`api::escrow` and `api::cashu` serialized the *same* globals with two different
mutexes. That only fails under parallel execution and reads as flakiness. Fixed
properly: one global, one lock (`escrow_mode::test_lock`).

## Verification

```
cargo test                                  166 passed, 14 ignored
flutter test                                161 passed
```

`./scripts/frb-generate.sh` run after the `rust/src/api/` changes. The wasm stub
grew the escrow surface so the bridge stays one codebase; it still refuses with
`CashuUnsupportedOnWeb`.

## Test plan

- [ ] Lightning node: take flow unchanged, no lock screen, `lock_escrow`
      returns `CashuNotEnabled`.
- [ ] Cashu node + nutshell, funded wallet: take a buy order as seller → lock
      screen shows amount/fee/total; confirm; daemon accepts and the buyer is
      told to send fiat.
- [ ] **Verify the fee against the daemon's stored `order.fee`** on an amount
      where the halves round differently (e.g. 500 sat at 0.6%) — this is the
      single most likely thing to be wrong.
- [ ] Empty wallet: "fund your wallet" leads to the C3 wallet screen.
- [ ] Kill the app between the mint swap and the publish: the token is still
      recorded against the trade, and re-submitting is accepted (the daemon's
      replay path).

Remaining: C6 (release), C7 (cooperative cancel), C8 (disputes), C9 (web
storage), C10 (resilience/backup).

---

# Manual verification

**Be honest about the ceiling here:** the escrow lock cannot be exercised
end to end without a `mostrod` running Cashu mode, and no public node does. So
this PR splits into *what can be proven today* (the crypto, the arithmetic, the
gate, the Lightning regression) and *what must wait for the daemon* (an accepted
`AddCashuEscrow`). Both are listed, and the second is the reviewer's real risk.

## Already run on this branch

```
cargo test                                  175 passed, 14 ignored
flutter test                                173 passed
```

Plus the integration suite against a real mint — **8/8, twice in a row**
(nutshell 0.20.3), covering the escrow primitives this flow calls into.

## A · The fee arithmetic — check this first

This is the single most likely thing to be wrong in production, because a
one-satoshi disagreement with the daemon is a rejection with no useful message.

```bash
cd rust
cargo test --lib node_fee
```

The rule is `2 * round(fee * amount / 2)` — the daemon rounds the *half*, so
doubling the rounded half is the only expression that agrees with it.
`the_total_fee_is_twice_the_rounded_half` pins 500 sat at 0.6%, where the naive
`round(fee * amount)` gives 3 and the correct answer is 4.

**Cross-check against a daemon if you can**: create an order for 500 sat on a
node charging 0.6%, read `order.fee` from its database, and confirm the app's
quote shows exactly `2 × order.fee`.

## B · The critical fix from review — party derivation

The first version took the buyer key from `counterparty_pubkey`, which is empty
for a maker seller and the *order-book* key for a taker. Both produce an escrow
the buyer cannot spend, and the taker case swaps the funds before the daemon
rejects it.

```bash
cargo test --lib the_buyer_key_comes_from_the_escrow_request_not_the_order_book
cargo test --lib the_locktime_clears_the_daemons_floor
```

To convince yourself the plumbing is real rather than just the test: in
`rust/src/api/cashu.rs`, temporarily change `trade.buyer_trade_pubkey` back to
`trade.counterparty_pubkey` and run the lock flow (section D) — it must fail
with `CashuEscrowRequestMissing` on a maker, rather than building a token.

## C · Lightning regression — the part that must not move

Run against the default (Lightning) node, with the developer override **off**.

1. Take a **buy** order as the seller → you land on the **hold invoice** screen,
   exactly as on `main`. Not the escrow screen.
2. Take a **sell** order as the buyer → the add-invoice / LN-address flow is
   unchanged.
3. Complete a trade end to end. Nothing in this PR is reachable.
4. **Take an order within a second or two of app launch.** This is the specific
   regression the review found: the routing gate used to be read before it had
   resolved, which could send a seller to the wrong screen. It must land on the
   hold invoice every time.

## D · The escrow screen (override + local mint)

Setup is the same as #237 (nutshell + the developer override + mint URL), and
you will need a funded app wallet — fund it through the C3 wallet screen with a
token from `docker exec nutshell poetry run cashu --host http://localhost:3338 send 64`.

5. Take a **buy** order as the seller with the override on → you land on
   **Lock the escrow** instead of the hold invoice screen.
6. The screen states **escrow**, **Mostro fee**, **total**, **your balance**,
   the **mint** and **when you can reclaim it**. Check the total is
   `amount + fee` and that the fee matches section A.
7. **With an empty wallet**: the button reads **Fund your wallet** and takes you
   to the C3 wallet screen — not a failure message.
8. **With funds**: tapping **Lock escrow** swaps at the mint. Against a
   Lightning-only daemon the submission will be rejected — that is expected,
   and the interesting part is what the screen does: it must show a **Retry
   sending** button and the *"locked but the node has not confirmed"* note,
   because the funds have already moved and the daemon's handler is idempotent.
9. Confirm the funds really moved: the wallet balance drops by roughly
   `amount + fee` (slightly more — the mint charges a swap fee, see below).
10. **Failures before the swap must not offer a retry.** Turn the override off
    and open the escrow screen for a trade: you get a localized message and
    **no** retry button, because nothing moved.

## E · Still needs a Cashu daemon

Everything below is unverifiable today and is where the remaining risk sits:

- an `AddCashuEscrow` the daemon **accepts**;
- the buyer receiving `cashu-escrow-locked` and being told to send fiat;
- the locktime clearing the daemon's floor in practice (the margin is reasoned
  and unit-tested, not observed);
- the fee token matching the daemon's stored `order.fee` on a live order.

When testing against the daemon's Track A branch, watch for `InvalidCashuToken`
and `InvalidMintUrl` — those are the two rejections this flow can still earn.

## Known limitation

A mint that charges swap fees makes locking cost slightly more than
`amount + mostro_fee`, and `cashu_escrow_quote` does not account for that. A
seller holding *exactly* the total can fail at the mint. Flagged rather than
fixed here; it wants a decision about whether the quote should pad or the UI
should warn.


grunch:
## Review — C5 (strict pass)

> **Note:** GitHub does not allow *Request changes* on your own PR, so this is posted as a comment. Treat it as **🔴 Request changes — blocking, do not merge.** Finding 1 means the flow this PR exists to deliver cannot succeed against a real daemon on either path, and the tests do not catch it because none of them reaches the party-derivation code.

**Scope reviewed:** `rust/src/api/cashu.rs`, `rust/src/mostro/{node_fee,actions}.rs`, `rust/src/api/{orders,nostr,types}.rs`, `lib/features/cashu/screens/lock_escrow_screen.dart`, `lib/features/order/screens/take_order_screen.dart`, l10n ×5.

---

### 🛑 Critical

**1. The buyer's key is taken from the wrong field — the lock is rejected on both paths** — `rust/src/api/cashu.rs:315`

```rust
let buyer_hex = trade.counterparty_pubkey.clone();
```

`counterparty_pubkey` is written in exactly two places, both at construction, and never updated afterwards (`rust/src/api/orders.rs:844` and `:1068`):

- **Maker seller** (line 844) — `String::new()`. `EscrowParties::from_xonly_hex("")` fails with `InvalidTradeKey`. A user who *created* a sell order can never fund the escrow. This is the most common seller in the product.
- **Taker seller** (line 1068) — `order.creator_pubkey`, i.e. the pubkey on the Kind 38383 order-book event. The daemon validates `proof.buyer_pubkey` against `order.get_buyer_pubkey()`, which is the buyer's **per-order trade key**. These are different keys, so the daemon returns `InvalidCashuToken` — after the seller's ecash has already been swapped into a token locked to a key the buyer does not hold.

The second case is worse than a failed lock: the funds are committed at the mint before the rejection arrives, and the escrow is unspendable by the intended buyer.

The right source is the escrow request itself. The daemon's `show_cashu_escrow_request` sets both:

```rust
new_order.buyer_trade_pubkey  = Some(buyer_pubkey.to_string());
new_order.seller_trade_pubkey = Some(seller_pubkey.to_string());
```

`classify_take_reply` receives that `Payload::Order(SmallOrder)` and drops both fields. C5 must capture them — persisted on `TradeInfo` alongside `cashu_mint_url` — and `lock_escrow` must read the buyer key from there, never from `counterparty_pubkey`. It is also worth asserting our own trade key equals the daemon's `seller_trade_pubkey` before locking; a mismatch means the wrong trade key was derived and the escrow would be unspendable by us.

---

### 🔴 Major

**2. The locktime is computed at exactly the daemon's floor, and evaluated earlier than the daemon evaluates it** — `lock_escrow()`

```rust
let locktime = now_secs() + locktime_days * SECONDS_PER_DAY;
```

The daemon's check is `locktime >= now + escrow_locktime_days`, evaluated when the submission is *validated*. Our `now` is strictly earlier than the daemon's `now` by the publish latency plus relay propagation. The comment argues that matching the floor is the shortest lock the daemon accepts — but matching it exactly means every lock is a race against network delay, and the losing side is a rejection with the funds already swapped.

Add a margin (an hour is invisible to the user and larger than any plausible propagation delay), or derive the locktime from a timestamp the daemon supplied.

**3. The fee is computed from the node's *current* advertised rate, not the order's stored fee**

`node_fee` is refreshed from the Kind 38385 `fee` tag; the daemon compares against `order.fee`, which was computed and stored **when the order was created**. An operator who changes the fee between order creation and the lock invalidates every in-flight order's fee token, and the client has no way to know.

The arithmetic itself is right and the rounding note is the good part of this PR — but the input is the wrong one. If the order does not carry its fee to the client today, that is worth raising upstream; until then the failure mode deserves an explicit error rather than a generic `InvalidCashuToken`.

**4. The escrow is committed before the fee token is built** — `lock_escrow()`

`build_escrow_token` confirms — funds swapped, irreversible — and only then is `build_fee_token` attempted. A failure there (insufficient balance for the fee after mint fees, mint hiccup) leaves the escrow locked, unsubmitted, and reclaimable only after the locktime. Prepare both sends, then confirm both; or on fee failure, immediately re-swap the escrow proofs back to the wallet and report it.

**5. `lock_escrow` reads the balance without ensuring the wallet is connected**

`cashu_escrow_quote` returns `balance_sats: 0` when no wallet is bound, so a call that does not go through the screen (a retry, a future notification action) reports `CashuInsufficientFunds` on a fully funded wallet. Either connect inside `lock_escrow` or return `CashuNotConnected`.

**6. `take_order_screen` reads the gate before it has resolved** — `lib/features/order/screens/take_order_screen.dart`

`ref.read(isCashuAvailableProvider)` is `false` while `escrowModeProvider` is still `AsyncLoading`. Take an order quickly after launch on a Cashu node and the seller is routed to the Lightning pay-invoice screen, which will sit waiting for a hold invoice that never arrives. Await the provider (or block the take button until it resolves) rather than reading it optimistically.

---

### 🟡 Minor

**7. No re-submit path after a failed publish.** The code correctly persists the token even when the publish fails, and correctly notes the daemon's handler is idempotent — but nothing in the UI ever calls `lock_escrow` again, so the user is left with locked funds and a trade that looks stuck. The retry the design assumes does not exist yet.

**8. `node_fee::set_fee` accepts 0.0 ≤ f < ∞ with no upper bound.** A malformed tag of `2.0` (200%) is stored and produces a fee token larger than the escrow. Cap it at something sane (1.0) and warn.

**9. `now_secs()` returns 0 on a clock before the epoch**, which then produces a locktime in 1970 and a confusing `InvalidEscrowConditions`. Prefer surfacing the clock error.

---

### ✅ What is right

- **Reading the daemon's branches instead of inferring the contract.** The fee-rounding discovery (`2 * round(fee*amount/2)`, not `round(fee*amount)`) is exactly the class of bug that would have shipped and been near-impossible to diagnose from the client side, and `node_fee`'s tests pin it with the 500-sat case where the two expressions diverge.
- Verifying our own escrow token with the same check the daemon runs, before publishing.
- Persisting the token even when the publish fails, with the reasoning stated: the ecash is already committed, and an unrecorded token is money we cannot find again.
- Fixing the two-mutexes-one-global test isolation bug properly rather than papering over the flake.
- The screen shows the total against the balance and offers "fund your wallet" instead of failing — the right shape for the most common seller error.

### Test gaps

The Rust tests cover the gate (`CashuNotEnabled` on every entry point) and the fee arithmetic. Nothing covers:

- party derivation — which is finding 1, and a single test constructing a maker `TradeInfo` and calling `lock_escrow` would have caught it;
- the locktime actually landing above the daemon's floor;
- the quote's insufficient-funds branch;
- any Dart coverage of `lock_escrow_screen` (no widget test), including the "fund your wallet" branch that is the phase's stated acceptance criterion.

grunch:
@coderabbitai Review

Forte11Cuba (changes requested):
### Blocking

**1. The phase's message lifecycle is one incomplete row: nothing that comes back is handled.**

Four symptoms, one work item:

- `Action::CashuEscrowLocked` has no arm in `dispatch_mostro_message` — it falls into `action => log::debug!("unhandled")` (`rust/src/api/orders.rs:2156`) and `status_for_action` doesn't map it. The daemon sends it to **both** parties after a successful lock (`add_cashu_escrow.rs`, step 10). So on the happy path: the buyer is never told to send fiat, and neither trade ever reaches `Active` — the fiat-sent button (gated on `(active, true)` in `trade_detail_screen.dart`) never appears. The spec's buyer side ("status → Active, notify 'escrow locked — send fiat now'") is explicit C5 scope.
- There is no `PendingRequestKind::CashuLock`. `lock_escrow` generates a `request_id` and publishes fire-and-forget (`rust/src/api/cashu.rs:446-465`). The correlation hook exists on the wire: the daemon's ack to the seller is `CashuEscrowLocked` carrying the submission's `request_id`.
- A rejection is **silently dropped**. With no pending record, `CantDo(InvalidCashuToken/InvalidMintUrl/CashuMintUnavailable)` dies in the `CantDo` arm's `"no matching pending request, ignoring event"` (`orders.rs:2098`). The seller saw the "submitted" snackbar; the trade hangs in `waitingPayment` forever with the funds already swapped.
- The Track A `CantDoReason`s are not mapped to l10n (spec requires it); `cashu_error_messages.dart` only covers client-local markers.

One wire nuance to encode while fixing this (verified against the daemon handler and `mostro-core`'s `check_status`, `order.rs:362-369`, identical across 0.14.1–0.14.6): a re-submission after a lock that actually succeeded fails the step-3 `check_status(WaitingPayment)` and comes back as **`CantDo(InvalidOrderStatus)`** — the retry path must **not** read that as failure. It means the order moved past `WaitingPayment`: usually to `Active` (the lock landed), but possibly to canceled/dispute — so reconcile against the order's current state rather than mapping it blindly to success; the authoritative success signal is the `CashuEscrowLocked` the daemon already sent. (Careful when implementing: the client's existing `CantDo` string mapper handles `NotAllowedByStatus`, which is a *different* variant than this path emits.) And a replay that loses the CAS race returns `Ok` with **no notification at all**, so the retry can't await an ack on that path.

**2. A maker seller cannot lock an escrow, period.**

`lockEscrowPath` is reachable from exactly one place: `take_order_screen.dart:149` — the taker path. The maker seller's escrow request arrives via status-sync (round 1 correctly captures their trade pubkeys there), but `my_order_screen.dart:161-163` routes `waitingPayment && isSelling` to `PayLightningInvoiceScreen`, which waits for a hold invoice that is `None`. Re-entry is equally broken for takers: `trade_detail_screen.dart:847-855` and `trades_list_item.dart:82-83` send a Cashu seller in `waitingPayment` to the Lightning screen. The spec lists all five routing sites to gate; the PR gates one. A `grep payInvoicePath` against that list finds these mechanically. The most common seller in the product (the maker) is the one with no path.

**3. Retry re-swaps instead of re-publishing — each tap costs the seller another `amount + fee`.**

`lock_escrow` never checks `trade.cashu_escrow_token`: it is written at `cashu.rs:472` and **read nowhere in the codebase**. The screen's Retry calls `lock_escrow` again, which (a) swaps another `amount + fee` out of the wallet, (b) **overwrites** the previous token record via `save_trade` — destroying the only reference to escrow #1 (recoverable only via the 15-day refund path, and only if cdk's internal store retains it), and (c) orphans the previous fee token (see finding 4 — it was never persisted). The stated rationale ("resubmission is safe, the daemon is idempotent") is true for re-sending the *same* token — I verified the daemon's validate-then-CAS discipline — but that's not what the code does. Realistic scenario: relays down, three Retry taps → 3×(amount+fee) gone from the wallet.

Also concurrency: `lifecycle_lock()` guards only connect/disconnect (`cashu.rs:141,267`); `lock_escrow` takes no lock. The only double-entry protection is the Dart screen's `_locking` flag. The fix belongs in Rust, where the money moves: one per-operation mutex, plus "if a token is persisted for this order, rebuild the message from it and only republish."

### Major

**4. Persistence happens after the publish, and the fee token is never persisted at all.**

The spec says *"persist the built tokens **before** publishing"* (plural, deliberately). The code does swap → publish → persist (`cashu.rs:465-477`), so a process death during the publish loses the record of an already-irreversible token — and the manual-verification claim "kill the app between the mint swap and the publish: the token is still recorded" is not true with this ordering. Worse: the **fee token** only ever exists inside the outgoing message. Its conditions are P2PK 1-of-1 to `P_M` with **no locktime and no refund** (`rust/src/cashu/escrow.rs:131-137`), so on any path where the message never arrives, those sats are unrecoverable by anyone who knows about them. Persist both tokens before publishing. (Separately worth raising upstream: a locktime+refund on the fee token would not stop the node redeeming it immediately, but would stop burning sats on failed delivery.)

**5. "On restart, an unacknowledged lock is retriable" — but nothing retries.**

No startup scan resubmits a persisted-but-unacknowledged token, and with finding 2 no UI route leads back to the lock screen after the app restarts. The persisted token is write-only state.

### Minor

**6. The quote's fallbacks contradict the PR's own principle.** `CashuNodeFeeUnknown` correctly fails instead of guessing zero — but `locktime_days` does `unwrap_or(15)` and `mint_url` does `unwrap_or_default()` (`cashu.rs:324-325`). The mint case is only reachable in a node-switch race (the gate requires a usable mint), but the locktime guess bites any node configured above 15 days — and since the §4.1 tags don't exist upstream yet, the client currently *always* guesses. A wrong guess is a post-swap rejection, the exact class of failure the fee principle exists to avoid.

**7. Doc upkeep (living-artifact rule).** The README header still says "C0 merged; C1a in review". And two wire facts verified in this review deserve pinning (the `cashu_wire.rs` pattern), so they stop being assumptions: the `fee` tag and its unit, and the idempotency semantics (below).

### Verified against the daemon — in the PR's favor

- **The `fee` tag and its unit are confirmed.** `nip33.rs:531` publishes the `fee` tag with the raw `mostro_settings.fee` and `util.rs:72-77` (`get_fee`) multiplies that value directly by the amount and rounds the half — the client's `split_fee_sats`/`total_fee_sats` match it expression-for-expression, including the 500-sat divergence case the tests pin. No unit bug.
- **Idempotency is real.** The handler validates fully before mutating, then commits via compare-and-set `WaitingPayment → Active`; replays are safe no-ops. Resubmitting the *same* token is exactly as safe as the PR claims (see finding 1 for the `InvalidOrderStatus` handling the retry must add).
- Daemon authentication is solid: `dispatch_mostro_message` verifies `sender == active mostro` before anything — including the newly persisted trade pubkeys — and the 38385 fetch filters by author.
- The round-1 fix cluster is genuinely correct: buyer key from the escrow request on both maker and taker paths, seller-key cross-check, locktime margin, fee-token-first ordering, verify-before-publish.
- The `CashuNotEnabled` gate covers every entry point, with the test to prove it.

### Test gaps (seam-level, per the lessons from #343)

The current tests are piece-level (arithmetic, gates) and could not have caught findings 1–3. Wanted: a dispatch test feeding a complete `CashuEscrowLocked` rumor and asserting the trade transition (both roles); a `lock_escrow` test with a token already persisted asserting republish-without-swap; a routing test for the maker seller in Cashu mode; and each "needs a Track A daemon" item in the test plan turned into an `#[ignore]`d test (the `MOSTRO_TEST_MINT_URL` pattern already exists for the mint).

PR #247, opened 2026-07-29
feat(chat): migrate P2P chat to the gift-wrap-free envelope

### LINKED ISSUE #246: Migrate peer-to-peer chat to the gift-wrap-free envelope (protects against the gift wrap apocalypse attack)
## Goal

Migrate the peer-to-peer chat to the new envelope specified in **[Peer-to-peer Chat](https://mostro.network/protocol/chat.html)** (approved and merged in MostroP2P/protocol#52), replacing the simplified NIP-59 gift wrap. A working reference implementation, with tests, lives in **[MostroP2P/mostro-chat#1](https://github.com/MostroP2P/mostro-chat/pull/1)**.

## Why this matters: the gift wrap apocalypse attack

The current chat wraps every message in a kind 1059 event signed by a **random ephemeral key**, addressed to the ECDH shared pubkey via a `p` tag. That shared pubkey travels **in clear text** on every event, so anyone scraping relays can harvest the address of every active conversation. And because each event carries a fresh ephemeral author, an attacker's events are **indistinguishable from genuine ones** until the recipient has already downloaded and attempted to decrypt them — one ECDH per event. There is nothing cheap to filter on and, by design, no author to rate-limit.

That turns into an attack that costs the attacker nothing and can cost the victim their money:

1. Alice sells, Bob takes the order, Alice pays the hold invoice, the trade goes active.
2. Bob sends the fiat and marks `fiat-sent`.
3. Alice starts flooding gift wraps at the conversation's shared pubkey.
4. Bob's client is buried in events it must decrypt before it can tell they are junk. Worse, the subscription has no `since`/`limit`, so the flood is **re-downloaded on every app start** — reinstalling does not help. To Bob it looks like the app is broken.
5. Bob never gets his sats and cannot reach the UI to **open a dispute**. Mostro sees `fiat-sent` followed by silence.
6. The hold invoice expires, the sats return to Alice, and she keeps the fiat too.

The attacker does not even need to be the counterparty — the conversation address is public — so this can be run against many trades at once, anonymously.

## Why the fix costs no privacy

The ephemeral key is protecting nothing here. In standard NIP-59 it matters because the `p` tag points at the recipient's **real identity key**. In our protocol the `p` tag already points at a shared key that is anonymous and unique per order — it is nobody's identity. An observer saw `ephemeral → shared`; under the new scheme it sees `shared → shared`. **Neither form ever exposes a trade key, and both group events identically.**

So the ephemeral key's only real effect was making participants indistinguishable from attackers. Signing the outer event with a key derived from the shared secret removes the attack at its root: the address stays observable, but **producing a valid event requires the shared private scalar**, which only the two parties can compute. Third parties become cryptographically unable to publish into a conversation, and clients drop everything else **at the relay** with an `authors` filter — at zero cost.

The only remaining flooder is the counterparty, who is now a single, stable, attributable author that a token bucket can cut off while the trade stays fully operational.

## What changes here

Envelope (see the spec for the normative details):

- Outer event becomes **kind 14 signed with `K_sign`**, `p`-tagged to `pub(K_conv)`, with the real timestamp (no NIP-59 tweaking — it breaks `since`-based sync).
- Inner event keeps its shape (kind 1 signed by the sender's trade key) but becomes the **only** authentication of who wrote a message, since both parties hold `K_sign`.
- The ECDH secret is split with HKDF-SHA256 into `K_conv` (encryption, and the `p` tag) and `K_sign` (outer signature).

Concrete work in this repo:

- [ ] `rust/src/nostr/gift_wrap.rs`: add the new wrap/unwrap for peer chat. The transport-v2 functions (`wrap_mostro_message` / `unwrap_mostro_message`) are unaffected.
- [ ] Key derivation: HKDF split of the existing ECDH secret (`on_peer_pubkey_received` in `rust/src/api/orders.rs` already derives it).
- [ ] `rust/src/api/messages.rs:673`: the subscription is `Filter::new().kind(1059).pubkey(shared_pubkey)` — **no author pinning and no `since`/`limit`**. It must become `authors = [pub(K_sign)]` bounded by a persisted cursor. Filtering by `p` tag alone reintroduces the vulnerability even after the envelope changes.
- [ ] `rust/src/api/messages.rs:758`: the inner sender is read from `inner.get("pubkey")` **without verifying the inner signature**. That is already a spec deviation; under the new scheme it is the only sender authentication, so it must verify and check the signer is a party to the order.
- [ ] Implement the client security requirements: cheapest-check-first validation order, absolute clock bound, durable inner-id dedup, rate limiting, and the isolation invariant — **chat must never be able to block the order state machine or a dispute**.
- [ ] Persist chat history, which the `since` cursor depends on (known gap).
- [ ] Disambiguate kind 14 by author: node pubkey → daemon reply, `pub(K_sign)` → peer chat.
- [ ] Dual-read window accepting both envelopes during migration, plus a deprecation date.

Dependency worth flagging: **`mostro-core` 0.14.1 still ships the superseded envelope** (`wrap_chat_message` / `unwrap_chat_message`, `SharedKey`, `chat_filter`, all kind 1059). That is the canonical home for this code, so it needs the same change; until it does, any client using it stays on the old scheme.

## Acceptance criteria

- Two clients exchange messages over the new envelope, and no published event contains either party's pubkey in any field.
- An event authored by anyone other than `pub(K_sign)` is rejected without decryption, and the relay filter never delivers it.
- A flood aimed at the conversation address does not degrade the order state machine, and a dispute can still be opened throughout.
- Restarting the app does not re-download an unbounded backlog.
- Disclosing `K_conv` to a solver grants read access without the ability to write into the conversation.

## References

- Specification: <https://mostro.network/protocol/chat.html> (MostroP2P/protocol#52)
- Reference implementation with tests: <https://github.com/MostroP2P/mostro-chat/pull/1>

grunch:
Closes #246.

## What

Migrates the peer-to-peer chat to the envelope specified in [Peer-to-peer Chat](https://mostro.network/protocol/chat.html) (MostroP2P/protocol#52), replacing the simplified NIP-59 gift wrap. The outer event is now a **kind 14 signed with `K_sign`** and `p`-tagged to `pub(K_conv)` — both HKDF-SHA256 derivations of the trade-key ECDH secret — carrying a NIP-44 encrypted kind 1 inner event signed by the sender's trade key. Third parties become cryptographically unable to publish into a conversation, and everything else is dropped **at the relay** by the `authors` filter.

## Changes

- **`rust/src/crypto/chat_keys.rs`** (new): `derive_chat_keys` — ECDH (`generate_shared_key`, raw x-coordinate) + HKDF-SHA256 split with domain separation (`mostro:chat:conv:v1` / `mostro:chat:sign:v1`) and the retry-byte fallback. Verified against the spec's **test vector**.
- **`rust/src/nostr/gift_wrap.rs`**: `mostro_wrap` / `mostro_unwrap` implementing the envelope and the crypto-side validation steps (author, exactly-one `p` tag, tag-count bound before signature work, absolute future-timestamp bound, size, outer signature, NIP-44 self-decryption, **inner signature verification**, allowed-signer check, inner kind, relative timestamp bound). Send side rejects with a stable `MessageTooLarge` anything a receiver would discard, and each inner event carries a signed uniqueness nonce so identical same-second sends keep distinct ids. `wrap`/`unwrap` (NIP-59) remain for dispute admin chat.
- **`rust/src/api/messages.rs`**:
  - Subscription pinned to `kind 14 + authors=[pub(K_sign)]`, bounded by a **persisted per-order `since` cursor** (advanced only past durably stored messages, clamped to `min(ts, local_now)` against cursor poisoning) plus a `limit`.
  - Cheapest-check-first pipeline: author → outer-id LRU → **token bucket (30 msg/min sustained, burst 60), metering only the live stream (post-EOSE)** so stored catch-up above the burst is never dropped → `mostro_unwrap` → durable inner-id dedup (**fail-closed**: lookup errors drop the event) → per-trade retention quotas (1000 msgs / 5 MiB). Sustained violation trips a flood breaker that halts chat processing while the trade stays fully operational.
  - **Dual-read migration window** (write-new/read-both): outbound is always the new envelope; inbound additionally accepts kind 1059 from pre-migration peers until `LEGACY_CHAT_DEPRECATION_TS` (2026-12-31T00:00:00Z), bounded by the same LRU/budget/size/dedup/quota. Unauthorized senders on the legacy path count toward the flood breaker.
  - **Lifecycle**: one chat task per order (spawn guard), explicit subscription ids unsubscribed on every exit path, no idle timeout, and `resubscribe_active_chats()` rebuilds listeners for persisted active trades whenever the relay pool comes online (startup and reconnect).
  - Kind 14 disambiguation by author: daemon subscriptions pin `author = mostro_pubkey`; chat pins `pub(K_sign)`; anything else is ignored.
- **Persistence** (closes the known chat-history gap): write-through store over the `messages` table, hydrated per trade; `message_exists` on the `Storage` trait is the durable replay dedup. The `messages → trades` FK is dropped (chat keys are per **order id**; a taker's trades row is a fresh UUID) with a crash-safe, re-runnable table-rebuild migration. `mark_messages_read` rewrites the JSON blob so read state survives restarts.
- **Web**: IndexedDB implements chat messages + the settings KV (cursor) with the same semantics as SQLite, fail-closed on lookup errors. The rest of the backend stays under #233.
- Docs: `specs/004` messages contract and `CLAUDE.md` transport section updated.

## Review history

- Round 1 (Codex ×3, CodeRabbit ×2, strict review ×7 + red CI): all fixed in b9f0ff3 — see the inline threads.
- Round 2 (CodeRabbit ×3 + 1 nitpick): fixed in 67f03fe. The IndexedDB `trade_id`-index/batching nitpick is deliberately deferred to #233 (full IndexedDB backend): per-trade quotas cap the store at 1000 messages, so the full scan stays small, and the index belongs with the real schema work.
- Follow-ups flagged (not this PR): live-relay integration harness for idle/restart recovery, browser-level reload-replay test, proposing the inner-event nonce tag upstream in MostroP2P/protocol, upstreaming the envelope to mostro-core (0.14.1 still ships the superseded one).

## Manual testing

Prereqs: a running `mostrod` + relay reachable by both clients (local regtest stack or the test node), and **two devices/profiles** of this app build — call them **Alice** (maker) and **Bob** (taker). Watch logs with `flutter run` consoles; chat lines are tagged `[messages]`.

1. **Happy path over the new envelope**
   1. Alice creates a sell order; Bob takes it; complete the steps until the trade is Active (peer pubkeys exchanged).
   2. Both consoles must log `incoming-chat subscription active order=<id> author=<pub(K_sign)> since=0 legacy=true`.
   3. Alice sends "hola" from the trade chat; Bob must receive it. Bob replies; Alice must receive it.
   4. On the relay (e.g. `nak req -k 14 <relay>`), verify the chat events are **kind 14**, authored by the same `pub(K_sign)` both logs printed, with exactly one `p` tag — and that **no event field contains either trade pubkey**.
2. **Restart / catch-up**
   1. Kill Bob's app. Have Alice send 2–3 messages.
   2. Relaunch Bob: after the pool logs ONLINE, expect `resubscribing chat order=<id>`, then the missed messages appear once, **no duplicates**, and history from before the restart is still there (persistence).
   3. Relaunch Bob again without new messages: the log line shows `since=<recent ts>` (cursor persisted) and no backlog is re-downloaded.
3. **Attachments**: Alice sends an image; Bob sees the attachment message and can download/open it (bytes travel via Blossom, pointer via the chat envelope).
4. **Send-size guard**: paste a >64 KB text into the chat input and send — the send must fail with the MessageTooLarge error, and **no** message may appear as "sent".
5. **Double-send**: send the exact same short text twice within one second — both must appear on the peer's side (uniqueness nonce).
6. **Legacy dual-read (mixed version)**: run Bob on a pre-migration build (v1 app or pre-#247 v2). Bob's messages (kind 1059 gift wrap) must still arrive to Alice. Alice's replies use the new envelope (old Bob won't see them — expected: write-new/read-both).
7. **Flood resistance (optional, test relay only)**: publish junk kind-14 events `p`-tagged to the conversation address from a third key (e.g. with `nak`). The clients must not log rejections (the relay-side `authors` filter drops them) and the trade UI must stay fully responsive; a dispute can still be opened.
8. **Web**: repeat step 1–2 on `flutter run -d chrome` — messages persist across a reload (IndexedDB) and no message is accepted twice after reload.

## Test plan (automated)

- [x] `cargo test` — 193 passed (30+ new: spec test vector, envelope rejection paths incl. junk-tag padding and send-size boundary, nonce uniqueness, token bucket + EOSE gating, LRU, quotas, payload parsing, durable dedup + FK migration + read-state rehydration via SQLite, legacy dual-read mixed-version path)
- [x] `cargo clippy --all-targets` — clean on touched files
- [x] `./scripts/frb-generate.sh --check` — bridge surface unchanged
- [x] `flutter analyze` / `flutter test` — clean / 173 passed
- [x] wasm type-check with the build-web RUSTFLAGS
- [ ] Manual testing above (needs the two-client setup)

### DIFF
--- a/rust/src/api/messages.rs
+++ b/rust/src/api/messages.rs
@@ -1346,23 +1346,33 @@ async fn handle_legacy_chat_event(
             return;
         }
     };
-    state.consecutive_rejected = 0;
 
     // The rumor is unsigned (NIP-59), so the claimed sender is only checked
     // for membership: this is exactly the weakness the new envelope fixes.
+    // Anyone can wrap to the public shared address, so an unauthorized or
+    // malformed sender still counts toward the flood breaker — resetting the
+    // streak before this check would let a flooder keep it at zero forever.
     let sender_hex = rumor.get("pubkey").and_then(|v| v.as_str()).unwrap_or("");
     let Ok(sender) = nostr_sdk::PublicKey::from_hex(sender_hex) else {
+        state.reject(order_id);
         return;
     };
     if !allowed_signers.contains(&sender) {
+        state.reject(order_id);
         return;
     }
+    state.consecutive_rejected = 0;
 
-    let rumor_id = rumor
+    // No id, no dedup identity: a fabricated fallback id would make the same
+    // rumor accepted again on every replay, so it is rejected outright.
+    let Some(rumor_id) = rumor
         .get("id")
         .and_then(|v| v.as_str())
         .map(|s| s.to_string())
-        .unwrap_or_else(|| uuid::Uuid::new_v4().to_string());
+    else {
+        log::debug!("[messages] legacy rumor without id order={order_id} — dropped");
+        return;
+    };
     match message_store().is_known(order_id, &rumor_id).await {
         Err(e) => {
             log::warn!("[messages] {e} — dropping legacy event order={order_id}");
--- a/rust/src/api/messages.rs
+++ b/rust/src/api/messages.rs
@@ -1,9 +1,23 @@
 /// Messages API — encrypted P2P chat during trades.
 ///
-/// P2P chat uses the ECDH-derived shared key (NIP-44 v2).
-/// Admin/dispute chat uses the BIP-32 trade key.
-/// All outbound messages are NIP-59 Gift Wrapped.
-/// Messages persist locally (in-memory until DB layer is wired in Phase 10+).
+/// P2P chat rides the chat envelope of the protocol spec
+/// (<https://mostro.network/protocol/chat.html>, issue #246): a kind 14 outer
+/// event signed with `K_sign` and `p`-tagged to `pub(K_conv)` — both derived
+/// from the trade-key ECDH secret via `crate::crypto::chat_keys` — carrying a
+/// NIP-44 encrypted kind 1 inner event signed by the sender's trade key. The
+/// old NIP-59 gift wrap (kind 1059) is gone from this channel: its random
+/// ephemeral authors made third-party flooding unattributable and unfilterable.
+/// Admin/dispute chat (api/disputes.rs) still uses gift wrap.
+///
+/// Messages persist to the `messages` table (native; web is memory-only until
+/// IndexedDB lands, #233); the in-memory store is a write-through cache. The
+/// stored inner-event ids double as the durable replay dedup the spec
+/// requires, and the per-order `chat_cursor:` setting bounds the subscription
+/// backlog.
+///
+/// **Isolation invariant**: everything here runs on its own spawned task and
+/// bounded channels; a chat failure or flood must never block the order state
+/// machine, the daemon transport, or opening a dispute.
 ///
 /// Streams: `on_new_message(trade_id)`, `on_unread_count_changed()`,
 /// `on_attachment_progress(message_id)`.
@@ -13,6 +27,7 @@ use std::sync::{Arc, OnceLock};
 use tokio::sync::{broadcast, RwLock};
 
 use crate::api::types::{AttachmentInfo, ChatMessage, DownloadStatus, FileType, MessageType};
+use crate::db::Storage;
 use crate::nostr::blossom;
 
 // ── Types ────────────────────────────────────────────────────────────────────
@@ -30,8 +45,13 @@ pub struct FileDownloadResult {
 // ── Message store ─────────────────────────────────────────────────────────────
 
 struct MessageStore {
-    /// Messages keyed by trade_id.
+    /// Messages keyed by trade_id. Write-through cache over the `messages`
+    /// table: adds persist immediately, reads hydrate from the DB once per
+    /// trade. Where no DB backend exists (web, unit tests) it degrades to
+    /// memory-only.
     messages: Arc<RwLock<HashMap<String, Vec<ChatMessage>>>>,
+    /// Trades whose persisted history has been loaded into `messages`.
+    hydrated: Arc<RwLock<std::collections::HashSet<String>>>,
     /// Broadcast channel for new messages (payload = trade_id of new message).
     new_message_tx: broadcast::Sender<ChatMessage>,
     /// Broadcast channel for global unread count changes.
@@ -47,13 +67,46 @@ impl MessageStore {
         let (attachment_tx, _) = broadcast::channel(64);
         Self {
             messages: Arc::new(RwLock::new(HashMap::new())),
+            hydrated: Arc::new(RwLock::new(std::collections::HashSet::new())),
             new_message_tx,
             unread_tx,
             attachment_tx,
         }
     }
 
+    /// Load the persisted history for `trade_id` into memory, once.
+    ///
+    /// Memory wins on id collision: an in-flight message may already sit in
+    /// the cache with fresher state (e.g. attachment download progress).
+    async fn ensure_hydrated(&self, trade_id: &str) {
+        if self.hydrated.read().await.contains(trade_id) {
+            return;
+        }
+        let persisted = match crate::db::app_db::db() {
+            Some(db) => match db.list_messages(trade_id).await {
+                Ok(msgs) => msgs,
+                Err(e) => {
+                    log::warn!("[messages] history load failed trade={trade_id}: {e}");
+                    Vec::new()
+                }
+            },
+            None => Vec::new(),
+        };
+        let mut store = self.messages.write().await;
+        let entry = store.entry(trade_id.to_string()).or_default();
+        for msg in persisted {
+            if !entry.iter().any(|m| m.id == msg.id) {
+                entry.push(msg);
+            }
+        }
+        drop(store);
+        self.hydrated.write().await.insert(trade_id.to_string());
+    }
+
     async fn add_message(&self, msg: ChatMessage) {
+        // Hydrate first so the persisted history is not masked by a fresher
+        // in-memory entry created before the first read.
+        self.ensure_hydrated(&msg.trade_id).await;
         {
             let mut store = self.messages.write().await;
             store
@@ -61,17 +114,48 @@ impl MessageStore {
                 .or_default()
                 .push(msg.clone());
         }
+        // Write-through: chat history and the durable replay dedup both live
+        // in the `messages` table. Failure is logged, never propagated — a
+        // full disk must not take the chat (let alone the trade) down.
+        if let Some(db) = crate::db::app_db::db() {
+            if let Err(e) = db.save_message(&msg).await {
+                log::warn!("[messages] persist failed id={}: {e}", msg.id);
+            }
+        }
         let _ = self.new_message_tx.send(msg.clone());
         let unread = self.unread_count_inner().await;
         let _ = self.unread_tx.send(unread);
     }
 
+    /// `true` if this message id was already accepted, in memory or on disk.
+    ///
+    /// This is the spec's durable inner-id replay dedup: a re-wrapped inner
+    /// event keeps the id it had the first time, so a hit here rejects it.
+    async fn is_known(&self, trade_id: &str, id: &str) -> bool {
+        {
+            let store = self.messages.read().await;
+            if let Some(msgs) = store.get(trade_id) {
+                if msgs.iter().any(|m| m.id == id) {
+                    return true;
+                }
+            }
+        }
+        match crate::db::app_db::db() {
+            Some(db) => db.message_exists(id)
+                .await
+                .unwrap_or(false),
+            None => false,
+        }
+    }
+
     async fn get_messages(&self, trade_id: &str) -> Vec<ChatMessage> {
+        self.ensure_hydrated(trade_id).await;
         let store = self.messages.read().await;
         store.get(trade_id).cloned().unwrap_or_default()
     }
 
     async fn mark_as_read(&self, trade_id: &str) {
+        self.ensure_hydrated(trade_id).await;
         let mut store = self.messages.write().await;
         if let Some(msgs) = store.get_mut(trade_id) {
             for m in msgs.iter_mut() {
@@ -79,6 +163,11 @@ impl MessageStore {
             }
         }
         drop(store);
+        if let Some(db) = crate::db::app_db::db() {
+            if let Err(e) = db.mark_messages_read(trade_id).await {
+                log::warn!("[messages] mark_messages_read failed trade={trade_id}: {e}");
+            }
+        }
         let unread = self.unread_count_inner().await;
         let _ = self.unread_tx.send(unread);
     }
@@ -103,10 +192,56 @@ fn message_store() -> &'static MessageStore {
 
 // ── Public API ────────────────────────────────────────────────────────────────
 
+/// The chat-key material for one conversation, derived from the session.
+struct ChatContext {
+    trade_keys: nostr_sdk::Keys,
+    /// `K_conv` — NIP-44 encryption; `pub(K_conv)` is the `p` tag.
+    conv: nostr_sdk::Keys,
+    /// `K_sign` — outer-event author; what relays and clients filter on.
+    sign: nostr_sdk::Keys,
+}
+
+/// Derive the conversation keys for a session's trade-key index and peer.
+///
+/// Cheap enough to derive on demand (one ECDH + two HKDF expands), which
+/// keeps the secrets out of long-lived session state.
+async fn chat_context(trade_key_index: u32, peer_hex: &str) -> Result<ChatContext> {
+    let trade_keys = crate::api::identity::get_active_trade_keys(trade_key_index)
+        .await
+        .map_err(|e| anyhow!("key retrieval failed: {e}"))?;
+    let peer_pubkey = nostr_sdk::PublicKey::from_hex(peer_hex)
+        .map_err(|e| anyhow!("invalid peer pubkey: {e}"))?;
+    let (conv, sign) = crate::crypto::chat_keys::derive_chat_keys(&trade_keys, &peer_pubkey)?;
+    Ok(ChatContext {
+        trade_keys,
+        conv,
+        sign,
+    })
+}
+
+/// Wrap `payload` in the chat envelope and publish it.
+///
+/// Returns the signed inner event on success — its id and timestamp are the
+/// message's durable identity (shared with the recipient's replay dedup).
+async fn publish_chat_payload(ctx: &ChatContext, payload: &str) -> Result<nostr_sdk::Event> {
+    let (outer, inner) =
+        crate::nostr::gift_wrap::mostro_wrap(&ctx.trade_keys, &ctx.conv, &ctx.sign, payload)
+            .await?;
+    let pool = crate::api::nostr::get_pool().map_err(|_| anyhow!("relay pool not ready"))?;
+    pool.client()
+        .send_event(&outer)
+        .await
+        .map_err(|e| anyhow!("publish failed: {e}"))?;
+    Ok(inner)
+}
+
 /// Send an encrypted text message to the trade counterparty.
 ///
-/// Validates that `content` is non-empty. Encrypts via NIP-59 and publishes
-/// to relays. If offline, the message is queued (queue wired in Phase 10+).
+/// Validates that `content` is non-empty, wraps it in the chat envelope
+/// (kind 14 signed with `K_sign`, inner kind 1 signed with the trade key) and
+/// publishes it. If the session, peer, or relay pool is not available the
+/// message is stored locally with a warning — same graceful degradation as
+/// before, chat never throws for transport reasons.
 ///
 /// Returns the sent `ChatMessage` (with `is_mine: true`).
 pub async fn send_message(trade_id: String, content: String) -> Result<ChatMessage> {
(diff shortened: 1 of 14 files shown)


rust/src/api/messages.rs
```diff
@@ -669,9 +869,19 @@ pub(crate) async fn subscribe_incoming_chat(
...
-        .kind(nostr_sdk::Kind::from(1059u16))
-        .pubkey(shared_pubkey);
+    let sign_pubkey = sign.public_key();
+    let allowed_signers = [my_trade_pubkey, peer_pubkey];
+
+    // `since` from the persisted cursor: everything older is already stored
+    // locally (the cursor only advances on accepted messages).
+    let mut cursor = load_chat_cursor(&order_id).await.unwrap_or(0);
+    let mut filter = nostr_sdk::Filter::new()
+        .kind(nostr_sdk::Kind::PrivateDirectMessage)
```
grunch, line 1075: **Blocking: this is a hard cutover even though #246 requires a dual-read migration window.** The pre-migration client listens for kind 1059 and publishes the old envelope; this head listens only for kind 14 here and publishes only the new envelope. During a staggered rollout, an updated peer and an older peer therefore cannot receive either direction of chat, including chats for already-active trades.

Please use a write-new/read-both transition (or another explicit version-negotiation strategy), retain the legacy receiver until the documented deprecation point, and test both mixed-version directions. Removing compatibility in the same release that introduces the new wire format breaks the linked issue's acceptance criterion.

rust/src/api/messages.rs
```diff
@@ -700,104 +913,113 @@ pub(crate) async fn subscribe_incoming_chat(
 
         match timeout(remaining, rx.recv()).await {
```
grunch: **Blocking: normal inactivity permanently removes the only receiver for this conversation.** `timeout(remaining, rx.recv())` exits the task after 30 minutes without a matching event, but no timer, relay reconnect hook, chat-screen open path, or startup rehydration path starts it again; the only caller is `on_peer_pubkey_received`, which runs when the daemon supplies the peer key. The next peer message after a routine 30-minute quiet period is therefore not processed, and after an app restart active trades are not resubscribed either.

This is separate from the existing subscription-ID/leak finding: even after that leak is fixed, the chat needs a durable owner that recreates listeners for active persisted trades and restarts them after idle/reconnect, or the listener must remain alive for the trade lifetime. Add an integration test that idles past the boundary and another that reconstructs an active conversation after process restart.

rust/src/api/messages.rs
```diff
@@ -645,21 +698,168 @@ fn mime_to_file_type(mime: &str) -> FileType {
...
 
-/// Spawn a background task that listens for NIP-59 gift-wrap (Kind 1059) events
-/// addressed to `shared_pubkey` and delivers decrypted messages into the in-memory
-/// message store, firing `on_new_message` so the Dart UI can react.
+/// Cap on the backlog requested from relays in one subscription.
+const CHAT_BACKLOG_LIMIT: usize = 500;
+
+/// Token bucket sizing per the spec: ~30 messages/minute sustained with a
+/// burst of 60, refused **before** any cryptographic work.
+const RATE_CAPACITY: f64 = 60.0;
```
grunch, line 758: **Blocking: the rate limiter bounds CPU rate but not durable storage growth, so the isolation invariant is still bypassable.** A counterparty sending continuously at or below `RATE_PER_SEC` consumes no rejected-event streak, never trips `FLOOD_TRIP_REJECTIONS`, and every accepted payload is appended to the in-memory vector and SQLite without any per-trade message-count or byte quota. Over a long-running trade this can fill memory/disk; a full database then affects order/dispute persistence, exactly what chat isolation is supposed to prevent. The normative spec explicitly calls for caps on both message count and total bytes per trade.

Enforce bounded retention/quota before accepting or persisting, keep the durable inner-id replay set consistent with whatever retention window the cursor can still fetch, and test a sustained valid-rate stream—not only an over-rate burst.

rust/src/api/messages.rs
```diff
@@ -117,95 +252,40 @@ pub async fn send_message(trade_id: String, content: String) -> Result<ChatMessa
...
+
+    match &session {
+        None => log::warn!("[messages] no session for trade={trade_id} — local-only"),
+        Some(s) => match &s.peer_pubkey {
+            None => log::warn!("[messages] session exists but peer not yet known — local-only"),
+            Some(peer_hex) => match chat_context(s.trade_key_index, peer_hex).await {
+                Err(e) => log::warn!("[messages] send_message trade={trade_id}: {e}"),
+                Ok(ctx) => {
+                    sender_pubkey = ctx.trade_keys.public_key().to_hex();
+                    match publish_chat_payload(&ctx, &content).await {
```
grunch, line 321: **A sender can report success for a message that every receiver running this code must reject.** `send_message` accepts arbitrarily large non-empty content and publishes it here, while `mostro_unwrap` rejects any encrypted outer content larger than `MAX_CONTENT_BYTES` before decrypting. Since the NIP-44 ciphertext also includes the signed inner-event JSON and encoding overhead, a sufficiently long UI message is published and stored locally as sent but is silently discarded by the recipient.

Validate the final serialized/encrypted envelope size before publishing (or enforce a conservative plaintext limit), return a stable `MessageTooLarge` error, and add boundary tests that round-trip the largest accepted payload and reject the next byte.

rust/src/db/indexeddb.rs
```diff
@@ -49,6 +49,12 @@ impl Storage for IndexedDbStorage {
     async fn mark_messages_read(&self, _trade_id: &str) -> Result<()> {
         Err(anyhow!("IndexedDB not yet implemented"))
     }
+    async fn message_exists(&self, _id: &str) -> Result<bool> {
```
grunch: **Blocking: the MUST-level durable inner-ID replay check is not implemented on the web target.** `save_message` always returns an unsupported error and `message_exists` always returns `false` here; cursor reads/writes are also no-ops. After every browser reload, a counterparty can re-wrap an already accepted signed inner event in a fresh valid outer event and it is accepted again as new/unread. The outer LRU only protects the current process and does not address this replay.

This cannot be described as a harmless persistence fallback because durable inner-event identity is part of the security protocol. Implement the message/cursor records in IndexedDB with the same semantics as SQLite, and make dedup lookup/storage errors fail closed rather than converting them to “not seen.” Add a browser reload test that re-wraps the same inner event under a new outer ID.

grunch (commented):
## Strict review — blocking changes required

Reviewed exact head `8adaf6574fed215e452f6402b254ccd82344d743` after reading all existing review threads and the completed automated reviews.

I confirmed the existing blockers and did not duplicate them inline: native message persistence fails on the `messages.trade_id -> trades.id` foreign key; catch-up can be dropped by the live token bucket; read state is not serialized durably; the cursor advances before durable storage; and subscriptions/tasks have no single-owner/unsubscribe lifecycle.

The new-envelope crypto and signature checks look sound against the normative test vector, but the migration and lifecycle are not safe to ship yet. In addition to the inline blockers below, the Rust check is red: CI reproduced the foreign-key failure and `mark_as_read_updates_count` failed (`183 passed, 1 failed`), while the same exact-head suite passed locally (`184 passed`). That inconsistent result also shows the global message-store tests are not isolated.

Please address all unresolved blocking threads and add integration coverage for mixed-version peers, idle/restart recovery, catch-up above the burst size, and persistence using the real `TradeInfo.id` / order-id relationship.

rust/src/nostr/gift_wrap.rs
```diff
@@ -86,10 +92,188 @@ pub async fn unwrap_mostro_message(
...
+    // Absolute bound against our own clock. Without it a counterparty can
+    // date both events far in the future — they agree with each other, so the
+    // relative check below passes — and poison the `since` cursor, silencing
+    // the conversation until that date. The past is unbounded: catching up
+    // after being offline is legitimate.
+    if outer.created_at.as_secs() > now.as_secs().saturating_add(MAX_CLOCK_SKEW_SECS) {
+        return Err(anyhow!("outer event is dated too far in the future"));
+    }
+
+    if outer.content.len() > MAX_CONTENT_BYTES {
```
grunch, line 274: **The 64 KiB check bounds only ciphertext, not the raw event that must be parsed and verified first.** The protocol reference explicitly requires the caller to cap the raw wire event as well. With the current nostr-sdk defaults, a valid kind-14 relay message can be roughly 5 MiB and carry up to 2,000 tags. A peer holding `K_sign` can therefore keep `content` small, add a large set of irrelevant tags, and force JSON parsing, event-ID construction, and signature verification before this check and before the per-chat token bucket sees the event; the initial 60-event burst can make this hundreds of MiB of work.

Configure a raw event-size limit for kind 14 at the relay/client boundary and a tight tag-count limit appropriate to this envelope, then retain this payload limit for decryption. Add a test with small ciphertext plus oversized extra tags/raw JSON.

rust/src/nostr/gift_wrap.rs
```diff
@@ -86,10 +92,188 @@ pub async fn unwrap_mostro_message(
...
+    conv: &Keys,
+    sign: &Keys,
+    message: &str,
+) -> Result<(Event, Event)> {
+    // One timestamp for both events: the real moment the message is sent.
+    // Recipients reject a mismatch, which is what bounds replays. No NIP-59
+    // timestamp tweaking — it would break `since`-based sync.
+    let now = Timestamp::now();
+
+    let inner = EventBuilder::text_note(message)
```
grunch, line 155: **Two intentional identical messages sent in the same second collapse to one durable identity.** The inner event uses second-resolution `created_at`, fixed empty tags, the same sender key/kind, and the text as content. Two rapid sends of the same text therefore produce the same signed inner event ID even though NIP-44 gives the outer envelopes different ciphertext/IDs; the receiver's inner-ID dedup drops the second one as a replay.

Give each inner chat message a signed uniqueness value (for example, a protocol-defined random nonce tag) and cover two identical same-second sends in the round-trip/dedup tests. Otherwise a normal double-send such as “yes” can be silently lost.

rust/src/api/messages.rs
```diff
@@ -47,38 +67,107 @@ impl MessageStore {
...
                 .entry(msg.trade_id.clone())
                 .or_default()
                 .push(msg.clone());
         }
+        // Write-through: chat history and the durable replay dedup both live
+        // in the `messages` table. Failure is logged, never propagated — a
+        // full disk must not take the chat (let alone the trade) down.
+        if let Some(db) = crate::db::app_db::db() {
+            if let Err(e) = db.save_message(&msg).await {
+                log::warn!("[messages] persist failed id={}: {e}", msg.id);
```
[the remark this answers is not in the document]
grunch: Fixed in b9f0ff3. Root cause confirmed: chat keys (and `messages.trade_id`) are per **order id**, while a taker's `trades.id` is a fresh UUID, so every taker `save_message` failed its FK check. The FK is dropped from the schema, and a one-off table-rebuild migration (`SQLITE_DROP_MESSAGES_FK_SQL`, detected via `pragma_foreign_key_list`) converts existing v2 databases preserving rows. Tests: `message_exists_is_durable_replay_dedup` now stores under an order id with no trades row at all, and `v2_messages_fk_is_dropped_and_rows_survive` builds a v2-era DB by hand and asserts the migration keeps the data.

rust/src/api/messages.rs
```diff
@@ -700,104 +913,113 @@ pub(crate) async fn subscribe_incoming_chat(
...
-                        continue;
+                // Step 5 — outer-id LRU: duplicate relay deliveries cost one
+                // hash lookup, nothing more.
+                if !outer_seen.insert(&event.id.to_hex()) {
+                    continue;
+                }
+
+                // Step 6 — rate-limit budget, before any cryptographic work.
+                if !bucket.try_take(crate::rt::time::Instant::now()) {
+                    consecutive_rejected += 1;
```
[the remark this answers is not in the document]
grunch: Fixed in b9f0ff3. The token bucket now meters only the **live** stream: `ChatRxState.live` flips on the relay's `EndOfStoredEvents` for our subscription ids, and `budget_ok()` is a no-op before that — stored catch-up is bounded by the filter `limit(500)` instead, so a backlog above the burst size is fully accepted and the cursor can no longer advance past dropped history. Covered by `bucket_is_bypassed_during_stored_catchup` (5× burst accepted pre-EOSE, normal metering post-EOSE).

rust/src/api/messages.rs
```diff
@@ -47,38 +67,107 @@ impl MessageStore {
...
         let mut store = self.messages.write().await;
         if let Some(msgs) = store.get_mut(trade_id) {
             for m in msgs.iter_mut() {
                 m.is_read = true;
             }
         }
         drop(store);
+        if let Some(db) = crate::db::app_db::db() {
+            if let Err(e) = db.mark_messages_read(trade_id).await {
+                log::warn!("[messages] mark_messages_read failed trade={trade_id}: {e}");
```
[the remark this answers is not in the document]
grunch: Fixed in b9f0ff3. `mark_messages_read` now also rewrites the JSON blob (`json_set(data, '$.is_read', json('true'))` — `json('true')` keeps it a JSON boolean so deserialization doesn't break). `mark_messages_read_survives_rehydration` closes and reopens the DB and asserts the flag survives.

rust/src/api/messages.rs
```diff
@@ -700,104 +913,113 @@ pub(crate) async fn subscribe_incoming_chat(
...
+                    is_read: is_echo,
+                    has_attachment: attachment.is_some(),
+                    attachment,
+                    // Presentation orders by the inner timestamp, which the
+                    // relative bound has already tied to the outer one.
+                    created_at: inner.created_at.as_secs() as i64,
                 };
 
                 log::debug!("[messages] incoming-chat rx order={order_id} id={}", msg.id);
                 message_store().add_message(msg).await;
```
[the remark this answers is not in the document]
grunch: Fixed in b9f0ff3. `add_message` now returns whether the message was durably stored, and `ChatRxState::advance_cursor` runs only after a successful store (duplicates advance too — they are already durable). A dedup lookup error also leaves the cursor put, so the event is re-fetched once storage recovers. The debouncing suggestion is noted as an optional follow-up — at 30 msg/min sustained the write amplification is modest.

rust/src/api/messages.rs
grunch: Fixed in b9f0ff3 — write-new/read-both, as requested. Outbound is always the new envelope; inbound additionally accepts kind 1059 from pre-migration peers until `LEGACY_CHAT_DEPRECATION_TS` (2026-12-31T00:00:00Z), after which the legacy subscription is simply not created. The legacy path decrypts with the shared secret as the recipient key (as v1 wraps it) and is bounded by the same outer-id LRU, live-stream budget, pre-decryption size cap, durable dedup and retention quota. Mixed-version coverage: `legacy_gift_wrap_is_accepted_during_the_window` exercises v1→v2 delivery, durable replay dedup, and rejection of a stranger-authored rumor.

rust/src/api/messages.rs
grunch: Fixed in b9f0ff3. The idle timeout is gone — the listener lives until relay-pool shutdown or a flood trip. Lifecycle now has a single owner: an `ACTIVE_CHATS` spawn guard makes repeated `on_peer_pubkey_received` invocations no-ops, subscriptions use explicit ids (`mostro-chat-<order_id>` / `mostro-chat-legacy-<order_id>`) and are unsubscribed on every exit path, and `resubscribe_active_chats()` rebuilds listeners for persisted active trades every time the pool transitions Online (startup and reconnect; selection logic covered by `chat_still_relevant_selects_only_live_trades`). A full idle-past-boundary integration test needs a live relay harness, which the unit CI doesn't have — flagging that as follow-up work.

rust/src/api/messages.rs
grunch: Fixed in b9f0ff3. Per-trade retention quotas are enforced before accepting/persisting any incoming message on both envelopes: `MAX_STORED_MESSAGES_PER_TRADE = 1000` and `MAX_STORED_BYTES_PER_TRADE = 5 MiB` (`MessageStore::quota_exceeded`). A counterparty writing forever at a legitimate rate now hits the cap and further messages are dropped and logged, leaving order/dispute persistence untouched. The durable inner-id dedup set is the messages table itself, so it stays consistent with what the cursor can still fetch. Covered by `quota_bounds_messages_and_bytes_per_trade` (sustained valid-rate growth up to the byte boundary, not an over-rate burst).

rust/src/api/messages.rs
grunch: Fixed in b9f0ff3. `mostro_wrap` rejects post-encryption anything whose ciphertext exceeds `MAX_CONTENT_BYTES` with a stable `MessageTooLarge:` marker, `send_message` propagates it to the caller instead of storing a phantom 'sent' message, and a cheap plaintext bound short-circuits before any crypto. Boundary tests in `oversized_message_is_refused_at_send_time`: 60 KiB fails with the marker, 30 KiB round-trips. The exact largest-accepted byte is deliberately not pinned — NIP-44 padding plus JSON escaping make it content-dependent — so the guarantee tested is 'never publish what a receiver must reject'.

rust/src/db/indexeddb.rs
grunch: Fixed in b9f0ff3. IndexedDB now implements the message records and the settings KV (chat cursor) with the same semantics as SQLite: `save_message`/`list_messages`/`mark_messages_read`/`message_exists` over a `messages` object store keyed by inner-event id, `get/set/delete_setting` over a `settings` store. Dedup lookups fail **closed**: any storage error is `Err`, and the pipeline drops the event without advancing the cursor. The rest of the backend stays stubbed under #233. A browser-reload replay test isn't automatable from cargo; the web smoke test exercises boot with the new backend, and a browser-level replay test is flagged as follow-up.

rust/src/nostr/gift_wrap.rs
grunch: Fixed in b9f0ff3. `mostro_unwrap` now bounds the raw event before any signature work: `MAX_OUTER_TAGS = 8` (the envelope defines exactly one p tag plus optional NIP-13 nonce) rejects junk-tag padding ahead of hashing/verification, in addition to the existing pre-decryption ciphertext cap. Covered by `junk_tag_padding_is_rejected_before_verification` (small ciphertext + 2000 tags). Relay-side `max_event_size`-style limits remain an operator concern noted in the spec's relay considerations.

rust/src/nostr/gift_wrap.rs
grunch: Fixed in b9f0ff3. Each inner event now carries a signed 8-byte random nonce tag (`[\"u\", <hex>]`), so two identical same-second sends produce distinct inner ids and the receiver's dedup no longer eats the second one — covered by `identical_same_second_sends_keep_distinct_identities`. Agreed this wants a protocol-level definition: the spec's inner event shows empty tags, so I'd propose the nonce tag upstream in MostroP2P/protocol as part of the mostro-core migration.

grunch:
## Review round 1 addressed — b9f0ff3

All 11 inline threads (Codex ×3, CodeRabbit ×1, strict review ×7) are fixed and replied to individually, plus the two non-inline findings:

- **CodeRabbit (outside diff) — subscription leak / duplicate spawns**: chat tasks now have a single owner per order (`ACTIVE_CHATS` guard makes replayed `on_peer_pubkey_received` calls no-ops), subscriptions use explicit per-order ids, and every exit path unsubscribes both the envelope and legacy subscriptions.
- **Red CI**: `mark_as_read_updates_count` asserted on the process-global unread counter and raced with parallel tests (the local-vs-CI inconsistency you flagged) — it now asserts per-trade read state deterministically.

Also in this round, per the strict review: dual-read migration window (write-new/read-both until 2026-12-31T00:00:00Z, mixed-version test), no idle death + startup/reconnect resubscription, per-trade retention quotas (1000 msgs / 5 MiB), `MessageTooLarge` at send time, raw-event tag bound before signature work, signed uniqueness nonce on inner events, cursor advancing only after durable store, fail-closed dedup, and the IndexedDB implementation of messages + settings so the durable replay check holds on web.

**Verification**: 193 Rust tests green (17 new), clippy clean on touched files, wasm check green, `flutter analyze`/`test` green, FRB `--check` unchanged.

**Flagged as follow-ups** (not in this PR): live-relay integration harness for idle/restart recovery, a browser-level reload-replay test, and proposing the inner-event nonce tag upstream in MostroP2P/protocol.

grunch:
@coderabbitai review

rust/src/api/messages.rs
```diff
@@ -645,169 +750,721 @@ fn mime_to_file_type(mime: &str) -> FileType {
...
+    active_chats().lock().await.remove(&order_id);
+    if let Ok(pool) = crate::api::nostr::get_pool() {
+        let client = pool.client();
+        client.unsubscribe(&chat_subscription_id(&order_id)).await;
+        client
+            .unsubscribe(&legacy_chat_subscription_id(&order_id))
+            .await;
+    }
+    log::debug!("[messages] incoming-chat subscription exiting order={order_id}");
+}
```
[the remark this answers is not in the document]
grunch: Verified — no bridge change needed. `subscribe_incoming_chat` is `pub(crate)` and not part of the FRB surface (FRB only exports public items of `crate::api`); `./scripts/frb-generate.sh --check` runs clean on 67f03fe, and the generated files are untouched in this PR.

rust/src/api/messages.rs
```diff
@@ -645,169 +750,721 @@ fn mime_to_file_type(mime: &str) -> FileType {
...
+        .map(|s| s.to_string())
+        .unwrap_or_else(|| uuid::Uuid::new_v4().to_string());
+    match message_store().is_known(order_id, &rumor_id).await {
+        Err(e) => {
+            log::warn!("[messages] {e} — dropping legacy event order={order_id}");
+            return;
+        }
+        Ok(true) => return,
+        Ok(false) => {}
+    }
```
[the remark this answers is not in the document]
grunch: Fixed in 67f03fe. The streak reset now happens only after the sender passes the allowed-signers check; malformed or unauthorized senders go through `state.reject()` — anyone can wrap to the public legacy address, so they must count toward the flood breaker. A rumor without an id is now rejected outright instead of getting a fabricated UUID (which would have made the same rumor accepted again on every replay).

rust/src/db/schema.rs
```diff
@@ -1,6 +1,27 @@
...
+    data            TEXT NOT NULL,
+    is_read         INTEGER NOT NULL DEFAULT 0,
+    created_at      INTEGER NOT NULL
+);
+INSERT INTO messages_v3 SELECT id, trade_id, data, is_read, created_at FROM messages;
+DROP TABLE messages;
+ALTER TABLE messages_v3 RENAME TO messages;
+CREATE INDEX IF NOT EXISTS idx_messages_trade ON messages(trade_id);
+PRAGMA foreign_keys = ON;
+"#;
```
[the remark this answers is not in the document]
grunch: Fixed in 67f03fe. The rebuild now runs inside one transaction (an interruption rolls back to the untouched v2 table), `DROP TABLE IF EXISTS messages_v3` clears a leftover from a previous interrupted attempt, and the `foreign_keys` pragma toggles stay outside the transaction (SQLite silently ignores that pragma inside one). The migration test now seeds a stray `messages_v3` to cover the rerun path.

Issue #197, opened 2026-07-18
Session lifecycle: defer session deletion on cancel to catch a trailing bond-slashed


AndreaDiazCorreia:
**Part of #145.**

## Goal
Ensure a trailing `bond-slashed` notice is never lost. On a timeout slash the daemon sends `canceled` first, then `bond-slashed` ~150 ms later; deleting the session on `canceled` drops the trade key from the subscription filter and the decryption key, so the notice can never be received or decrypted.

## Parity reference (v1)
`MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` sections 7 and 8.

## Tasks
- On a cancel that could precede a slash, defer session deletion for ~60 s instead of deleting immediately (session logic in `rust/src/mostro/session.rs`).
- Cancel the defer (delete immediately) once `bond-slashed` is received and handled.
- Do not defer on a voluntary cancel (that returns the bond).

## Acceptance criteria
- A `canceled` -> `bond-slashed` sequence is received and decrypted end-to-end.
- Defer/reconcile logic is covered by pure tests (mirror v1 `bond_cancel_helpers_test.dart`).

AndreaDiazCorreia:
## The premise here does not hold for v2

This issue is written against v1's session lifecycle, where deleting the session
on `canceled` drops the trade key from the subscription filter and discards its
decryption key — which is what makes the trailing `bond-slashed` unreachable
there. Two things about v2 turn out to be different, both verified in the code:

1. **v2 never deleted sessions at all.** `remove_session` had no production
   caller, and the `Canceled` arm only touched the order book and the trade row.
2. **Reception and decryption do not depend on the session anyway.**
   `subscribe_gift_wraps` captures its `recipient_keys` inside the spawned
   receiver, and `ensure_global_dm_coverage` keeps freshly derived trade keys in
   a process-lifetime map, refreshing the bulk Kind-14 filter. Removing a
   `Session` neither unsubscribes that key nor discards those keys. The
   `BondSlashed` handler does not read the session either — it infers the slash
   cause from the persisted trade's status.

So the acceptance criterion "a `canceled` -> `bond-slashed` sequence is received
and decrypted end-to-end" is not something a session deferral establishes here.

## What #258 does deliver

Refs, not Closes, for that reason. It is the session-cleanup policy this issue
asks for, plus the correctness work it turned out to need:

- The cleanup policy on cancel (immediate / defer 60 s / keep), with pure tests
  — the second acceptance criterion.
[...]

PR #258, opened 2026-07-31
feat(#197): defer session removal so a trailing bond-slashed still decrypts

AndreaDiazCorreia:
Refs #197 — deliberately not auto-closing it, see **Scope** below. Part of the
anti-abuse bond epic #145.

## Problem
Cancel is where a session's life ends, and on a timeout slash the daemon sends
`canceled` first with `bond-slashed` ~150 ms behind it. v2 turned out never to
delete sessions at all — `remove_session` had no production caller — so the
policy had to be defined before #191 and #208 introduce that deletion, and the
leak closed with the deferral already in place.

## Fix
- `SessionManager` tracks deferred removals (defer / resolve / reconcile),
  with sessions and deferrals under one lock.
- `cancel_cleanup` decides from the status held *before* the cancel: `Pending`
  removes at once (the cancel returns the bond), dispute and admin states keep
  the session (the admin chat still needs its keys), anything else — including
  an unknown status — defers 60 s.
- `Action::Canceled` reads that status before syncing `Canceled`;
  `Action::BondSlashed` settles a pending deferral and is a no-op without one.
- Both are gated on the recipient trade key, and deferrals carry the trade key
  index they were armed for, so a delivery from a superseded take cannot claim
  the session of the one that replaced it.
- `install_session`: an accepted take replaces whatever it finds instead of
  discarding a `SessionAlreadyExists` error, which previously left a confirmed
  retake with no session at all.

## Scope
This is session-cleanup policy, not a transport guarantee, and #197's
end-to-end acceptance criterion is **not** proven here — hence `Refs`, not
`Closes`.

v1's rationale for the grace period does not carry over: there, deleting the
session drops the trade key from the subscription filter and discards its
decryption key. In v2 the per-trade receiver captures its own trade keys and
`ensure_global_dm_coverage` retains them for the life of the process, so
reception and decryption never depend on the session. The window is a margin
for handling that needs session state, plus parity with v1's policy.

Exercising `canceled` → `bond-slashed` end to end needs a receiver/dispatcher
regression and the taker bond flow from #208. Serializing the generation gate
with the order-book and DB writes is tracked in #259.

## Divergence from v1
v1 keys off `!userInitiated && hadBond`; this keys off the pre-cancel status,
needing neither an outbound cancel marker nor a bond-policy lookup. It defers in
a few cases v1 would not — the cost is a session held 60 s longer. Sessions are
in-memory only, so v1's restart reconciliation does not apply. Documented in
`specs/004-mostro-p2p-client/contracts/orders.md`.

## Tests
16 session tests plus 6 over the cancel-cleanup and generation helpers.
`cargo test` 256 passing, `cargo clippy` clean, `flutter analyze` clean,
`flutter test` 195 passing, `frb-generate --check` unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved order cancellation and bond-slashing handling to prevent stale events from affecting newer trades.
  * Added safer session cleanup based on order status, including deferred removal during bond grace periods.
  * Preserved sessions correctly when orders are retaken.

* **Documentation**
  * Documented session lifecycle behavior, cancellation handling, retakes, and bond-grace-period rules.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1830,6 +1830,9 @@ async fn dispatch_mostro_message(
             log::info!("[orders] gift-wrap Canceled for trade={trade_pubkey_hex}");
             if let Some(order_id) = &kind.id {
                 let oid = order_id.to_string();
+                if !is_current_generation(&oid, trade_index).await {
+                    return;
+                }
                 order_book().remove_order(&oid).await;
                 // Read before the sync below overwrites it with Canceled.
                 let prev_status = match crate::db::app_db::db() {
@@ -1856,7 +1859,7 @@ async fn dispatch_mostro_message(
                         log::warn!("[orders] failed to sync Canceled status for {oid}: {e}");
                     }
                 }
-                apply_cancel_cleanup(&oid, prev_status.as_ref()).await;
+                apply_cancel_cleanup(&oid, prev_status.as_ref(), trade_index).await;
             }
         }
         // Seller receives BuyerTookOrder → peer is buyer_trade_pubkey.
@@ -2127,6 +2130,9 @@ async fn dispatch_mostro_message(
                     return;
                 }
             };
+            if !is_current_generation(&order_id, trade_index).await {
+                return;
+            }
             let small_order = match &kind.payload {
                 Some(mostro_core::message::Payload::Order(so)) => so,
                 _ => {
@@ -2163,7 +2169,7 @@ async fn dispatch_mostro_message(
                 "[orders] gift-wrap BondSlashed: order={order_id} amount={amount_sats} cause={cause:?}"
             );
             crate::mostro::session::session_manager()
-                .resolve_deferred_removal(&order_id)
+                .resolve_deferred_removal(&order_id, trade_index)
                 .await;
             crate::api::bond::emit_bond_slashed(crate::api::types::BondSlashedEvent {
                 event_id: event_id.to_string(),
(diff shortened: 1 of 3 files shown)


AndreaDiazCorreia:
@codex review

AndreaDiazCorreia:
Thanks — all four findings were the same defect seen at successively narrower
windows, and each round is in the branch:

- **Two-map race** (Codex, CodeRabbit): `sessions` and `deferred_removals` now
  share one `RwLock<SessionState>`, so no operation observes one mid-update
  against the other.
- **Generation binding** (Codex): `Canceled` and `BondSlashed` are gated on the
  recipient trade key before any mutation; deferrals carry the trade key index
  they were armed for, and neither the trailing notice nor the timer removes a
  session of another generation. Writing that exposed a real bug — the first
  version compared the generation against the deferral instead of against the
  session it was about to drop — which the new test caught.
- **TOCTOU after the gate** (ermeme): the severe half is closed. `take_order`
  now uses `install_session`, which replaces whatever it finds and clears any
  pending deferral under the same lock, so an accepted take keeps its session
  whoever wins the race. The `SessionAlreadyExists` error is no longer
  discarded.

What is not in this PR: serializing the gate together with the order-book and DB
writes. That needs a per-order keyed mutex across the dispatcher and test
scaffolding to drive `dispatch_mostro_message`, and the `Canceled` arm has
mutated book and DB without any generation check since well before this branch.
Tracked separately in #259 rather than growing a PR about not losing a
`bond-slashed`.

Verified: `cargo test` 256 passing, `cargo clippy` clean, `flutter analyze`
clean, `flutter test` 195 passing, `frb-generate --check` unchanged.

Catrya (commented):
@AndreaDiazCorreia please fix the conflicts

Catrya:
Closing this. Two things changed under it since it was opened on July 31, and both go to its premise.

**The premise no longer holds.** The description says `remove_session` had no production caller, so the deletion 
Closing this. Two things changed under it since it was opened on July 31, and both go to its premise.

**The premise no longer holds.** The description says `remove_session` had no production caller, so the deletion policy had to be defined before #191 and #208 introduced one. It has two callers on `main` now: the `Action::Canceled` arm when an order is canceled before going active (`orders.rs:1574`, commented *"Mirrors v1, which deletes pending/waiting sessions on cancel"*), and the stale-trade sweep (`orders.rs:2525`). That policy landed by another route and matches the `Immediate` half proposed here, so a rebase would mean reconciling with an existing policy rather than reapplying this design — on a branch that is 129 commits behind and conflicts in `orders.rs`, `session.rs` and `contracts/orders.md`.

**The deferral has no consumer.** The description already notes that reception and decryption don't depend on the session. Neither does the handler for the trailing message: `Action::BondSlashed` on `main` infers the cause from the trade row in the database (`db.get_trade_by_order_id` → `infer_slash_cause`), not from session state. So the grace window is, in the PR's own words, *"a margin for handling that needs session state"* — handling that doesn't exist yet.

For reference, v1 gates the same deferral on `!userInitiated && hadBond` (`bond_cancel_helpers.dart`), because *"the daemon only sends a `bond-slashed` notice after a waiting-timeout slash, never after a voluntary cancel"*. `hadBond` is a precondition, and no order in this app can carry a bond yet — so under v1's rule this would fire never, where the status proxy here fires on every non-`Pending` cancel. If it comes back, the node's `bond_enabled` tag (kind 38385, already parsed on the Dart side, and Rust already fetches that event for pow / `protocol_version` / escrow mode) is the cheap way to make it a no-op when no bond can exist.

**Where the work goes:**

- The bond-specific half returns with the anti-abuse bond epic, which is being deferred until the order flows are solid. Nothing about it can be verified before then — including #197's acceptance criterion, which this PR correctly declined to claim.
- The two bond-independent fixes are captured in **#335**: `install_session` (a confirmed retake currently keeps the stale session, because `take_order` discards `create_session`'s `SessionAlreadyExists`) and the generation gate against messages from a superseded take.

Also relevant to whoever picks #335 up: **#334**. `create_session` is called only in `take_order`, so a maker never has a session at all, and the counterparty pubkey the daemon reveals is written only into that session and never onto the trade row. Verified against a live daemon. It's the same code, and the session model's real gap is at the creation end rather than the deletion end.

Two things here hold up and are worth carrying forward: skipping v1's restart reconciliation is correctly argued — sessions are in-memory and the `Storage` trait has no session methods, so there is nothing to reconcile — and the 60 s window matches v1's `_bondCancelGraceWindow` exactly.

Issue #202, opened 2026-07-22
open_dispute persists a local dispute optimistically — daemon CantDo rejection is never reconciled


AndreaDiazCorreia:
**Type:** bug
**Priority:** medium
**Area:** rust-core / protocol

### Summary

`open_dispute` treats a successful relay publish as a successful dispute. It persists a local `Dispute { status: Open }` immediately after publishing the `Dispute` action, **without waiting for the daemon's reply**. When the daemon rejects the request with `CantDo`, the rejection is never reconciled: the optimistic local dispute stays `Open`, so the action appears to have succeeded when it did not.

### Current behavior

In `rust/src/api/disputes.rs`, `open_dispute`:

```rust
crate::api::orders::publish_event(&event_json).await ...; // publish to relay
// Publish succeeded — persist the dispute record.
let dispute = Dispute { status: DisputeStatus::Open, initiated_by_me: true, ... };
dispute_store().try_insert_if_absent_or_resolved(dispute).await
```

Unlike `create_order` / `take_order` / `send_invoice`, `open_dispute` does **not** register a pending-request waiter for the daemon reply. The request/reply correlation and `CantDo` handling live in the dispatcher in `rust/src/api/orders.rs` (`Action::CantDo` arm), which notifies a waiting caller and rolls back optimistic state on rejection. Because `open_dispute` never registers as a waiter, an incoming `CantDo` for a dispute matches no pending request and is dropped:

> `CantDo: reason=… — no matching pending request, ignoring event`

On the Dart side, `_openDispute` in `lib/features/trades/screens/trade_detail_screen.dart` then upserts the dispute into `disputeNotifierProvider` and navigates to the dispute detail, showing it as `Open` — a false success.

### Steps to reproduce

1. Reach a state where the daemon rejects the dispute (the trade is not yet disputable).
2. Tap **Open dispute**.
3. The app navigates to the dispute detail and shows the dispute as `Open`.
4. The daemon replies `CantDo`; the local dispute is never reverted.

### Expected behavior

`open_dispute` should wait for the daemon's reply using the same request/reply correlation as `create_order` / `take_order` / `send_invoice`, persist the dispute **only** on a successful reply, and surface `CantDo` as an error. The Dart handler already catches errors from `openDispute`, shows a localized SnackBar, and rethrows — so propagating the error is enough to remove the false-success navigation.

### Related

- #143 Wire dispute system end-to-end (list + admin chat send/attach)
[...]

PR #271, opened 2026-08-05
fix(#203): stop the public order status from driving trade actions

### LINKED ISSUE #203: Trade Detail offers Open dispute before the trade is disputable (daemon rejects with CantDo)
**Type:** bug
**Priority:** medium
**Area:** ui / protocol

### Summary

Trade Detail presents the **Open dispute** action while the trade is not actually disputable. Attempting the dispute is then rejected by the daemon with `CantDo`.

### Steps to reproduce

1. Create an order and have the counterpart pay only the bond (without completing the trade funding).
2. Open Trade Detail: the status shows **Active** and the **Open dispute** button is offered.
3. Tap **Open dispute** -> the daemon rejects with `CantDo`.

### Observations

- The client gates the dispute action on `canDispute = status == active || status == fiatSent` (`_buildSecondaryActionRow` in `lib/features/trades/screens/trade_detail_screen.dart`).
- The displayed status is derived from the polled order status (`tradeStatusProvider` -> `_mapOrderStatus`, where `OrderStatus.active | inProgress -> TradeStatus.active`).
- The daemon's rejection implies its view of the trade is not yet disputable at this point. Either:
  - the displayed **Active** status is ahead of the daemon's real state (status desync), or
  - the client `canDispute` gating does not match the daemon's actual dispute preconditions (which, with the anti-abuse bond flow, may require the trade to be fully funded before a dispute is valid).

### Expected behavior

The client should not present an action the daemon will reject. `canDispute` should be gated on the real disputable state so **Open dispute** is only offered when the daemon would accept it.

### Notes

Needs confirmation against the daemon of the exact precondition for a valid dispute at this stage (relevant to the anti-abuse bond flow). Marked `needs verification` for the root-cause determination (status desync vs. gating mismatch).

### Related

- #145 [Epic] Anti-abuse bond — full protocol support
- #191 Phase 5: maker anti-abuse bond on order creation / #192 / #193 (bond flow phases)
- #114 Migrate per-trade UI state from polling to a push event stream (Rust -> Dart)
- #143 Wire dispute system end-to-end
- #202 — the daemon's `CantDo` rejection for a dispute is currently not reconciled locally

AndreaDiazCorreia:
Closes #203.

## Root cause

The `s` tag of a Kind 38383 event is NIP-69's four-bucket public view (`pending`, `in-progress`, `success`, `canceled`), and mostrod stops publishing once a trade turns private — `active`, `fiat-sent` and `dispute` never reach the wire. Both ingest paths wrote that bucket straight into the trade row and the order book, dragging a trade that daemon messages had already advanced to `Active` back to `InProgress`, which the UI presented as an active trade.

That is what offered **Open dispute** (and **Mark fiat sent**) on a trade the daemon rejects with `CantDo`: it only accepts either on an `Active` or `FiatSent` order. So the gating rule was right; the status feeding it was not — status desync, not a gating mismatch (answers the issue's `needs verification`).

## What changed

- **Rust** — `BuyerTookOrder` / `HoldInvoicePaymentAccepted` now imply `Active` in `status_for_action`; a wire status only applies when no finer local one is known or when it is terminal (`wire_status_applies`); `open_dispute` checks the status it already holds before publishing and fails with the `TradeNotDisputable` marker.
- **Dart** — `in-progress` gets its own `TradeStatus`: waiting presentation, no dispute and no fiat-sent, cancel still available since
  the daemon accepts it there.
- **Docs** — the public-vs-trade status rule in the orders contract, `OrderStatus`, the state machine, the `open_dispute` precondition and CLAUDE.md.

## Verification

`cargo test` and `cargo clippy` (24 warnings, same as main), `flutter analyze` clean, `flutter test` 197 passed. Three Rust tests and two widget tests added.

## Notes

- Overlaps with #213 (bond taker) in `trade_detail_screen.dart` and the status maps; branched off `main`, so whichever merges second rebases.
- #202 stays open: the daemon's `CantDo` reply is still not reconciled locally. This PR removes the path that triggered it.

## Summary by CodeRabbit

- **New Features**
  - Added a clear “Trade setup in progress” state with waiting messaging and appropriate actions.
  - Trade status now remains accurate instead of being shown as active prematurely.
  - Dispute actions are unavailable until the trade reaches an eligible state.

- **Bug Fixes**
  - Improved handling of trade status updates to prevent incorrect state changes.
  - Added a specific message when disputes are attempted too early.
  - Improved status updates for buyer and seller actions.

- **Localization**
  - Added corresponding trade-status and dispute messages in English, German, Spanish, French, and Italian.

- **Documentation & Tests**
  - Updated status behavior documentation and expanded coverage for in-progress trades.


Catrya (approved):
tACK

PR #275, opened 2026-08-05
fix(#202): wait for the daemon reply before persisting a dispute

### LINKED ISSUE #202: open_dispute persists a local dispute optimistically — daemon CantDo rejection is never reconciled

AndreaDiazCorreia:
Closes #202

## Problem

`open_dispute` treated a successful relay publish as a successful dispute: it persisted `Dispute { status: Open }` right after publishing and never registered a pending-request waiter. The `Dispute` message also carried no `request_id`, so the daemon's `CantDo` rejection matched nothing in the dispatcher and was dropped (`no matching pending request, ignoring event`). The dispute stayed locally Open and the UI navigated to the dispute detail — a false success.

## Fix

The Dispute message now carries a `request_id` nonce, which the daemon echoes in both `DisputeInitiatedByYou` and `CantDo` (verified in mostro `src/app/dispute.rs` and `src/app.rs`). `open_dispute` registers a pending record before publishing and waits up to 10 s, exactly like `create_order` / `take_order` / `send_invoice`. It persists only on a correlated acceptance; rejection and timeout return an error and write nothing.

The acceptance also carries the daemon's dispute UUID — the id its Kind 38386 event and the solver refer to — so the record is stored under it instead of a locally minted one.

No Dart changes: `_openDispute` already showed a localized SnackBar and rethrew without navigating, so propagating the error was enough.

## Notes

The daemon replies `CantDo` only for `MostroCantDo` causes. A duplicate dispute or a daemon-side DB failure is an internal error it merely logs, so those surface as `NoDaemonResponse` rather than a precise reason.

## Verification

`cargo build`, `cargo clippy --locked -- -D warnings`, `cargo test` (238 pass), `flutter analyze`, `flutter test` (195 pass). `frb-generate.sh` produces no drift — the public bridge surface is unchanged.

Tests cover the correlation layer, where the bug was: the kind/nonce gates of `take_matching_dispute`, a `CantDo` reaching the waiting caller through the dispatcher's lookup, and the dispute-id extraction. The end-to-end path needs a live daemon — pending manual check with the #203 repro.

## Follow-ups (not in this PR)

- A record created for a peer-opened dispute still gets a local UUID, so the two sides know the same dispute under different ids. Documented in `data-model.md`.
- `fiat_sent`, `release` and `cancel` still send no nonce and await no reply — the same false-success shape as this issue, in other actions.

## Summary by CodeRabbit

* **New Features**
  * Dispute requests now wait for daemon confirmation and use the daemon-provided dispute ID.
  * Concurrent dispute attempts for the same trade are rejected.
  * Late acceptance responses can reconcile disputes after a timeout.

* **Bug Fixes**
  * Improved handling of rejections, timeouts, malformed replies, publish failures, and duplicate disputes.
  * Prevented disputes from being recorded unless acceptance is confirmed.
  * Improved matching of dispute responses to their originating requests.

* **Documentation**
  * Clarified request correlation, timeouts, dispute identifiers, and peer-opened disputes.

### DIFF
--- a/lib/core/daemon_errors.dart
+++ b/lib/core/daemon_errors.dart
@@ -44,5 +44,10 @@ String localizedDaemonError(
   if (raw.contains('TradeNotDisputable')) {
     return l10n.tradeNotDisputable;
   }
+  // A dispute for this trade already exists, or one is still in flight: the
+  // open is a duplicate either way, and retrying it changes nothing.
+  if (raw.contains('DisputeAlreadyOpen')) {
+    return l10n.disputeAlreadyOpen;
+  }
   return fallback;
 }
--- a/lib/l10n/app_de.arb
+++ b/lib/l10n/app_de.arb
@@ -345,6 +345,7 @@
   "openDisputeFailed": "Streit konnte nicht eröffnet werden. Bitte erneut versuchen.",
   "openDisputeTitle": "Streitfall eröffnen",
   "openDisputeConfirmation": "Möchtest du wirklich einen Streitfall eröffnen? Dies eskaliert den Handel an einen Administrator und kann nicht rückgängig gemacht werden.",
+  "disputeAlreadyOpen": "Für diesen Handel ist bereits ein Streit eröffnet.",
   "tradeNotDisputable": "Ein Streit kann erst eröffnet werden, wenn die Gelder für diesen Handel gesperrt sind.",
   "tradeWaitingInvoiceBuyerInstruction": "Sende deine Lightning-Rechnung, damit der Verkäufer die Gelder sperren kann.",
   "tradeWaitingInvoiceSellerInstruction": "Warte auf die Lightning-Rechnung des Käufers.",
--- a/lib/l10n/app_en.arb
+++ b/lib/l10n/app_en.arb
@@ -705,6 +705,8 @@
   "@openDisputeTitle": {"description": "Title of the open-dispute confirmation dialog"},
   "openDisputeConfirmation": "Are you sure you want to open a dispute? This escalates the trade to an admin and cannot be undone.",
   "@openDisputeConfirmation": {"description": "Body of the open-dispute confirmation dialog"},
+  "disputeAlreadyOpen": "A dispute for this trade is already open.",
+  "@disputeAlreadyOpen": {"description": "Snackbar shown when a dispute is opened for a trade that already has one, or while a previous attempt is still in flight"},
   "tradeNotDisputable": "A dispute can only be opened once the funds are locked for this trade.",
   "@tradeNotDisputable": {"description": "Snackbar shown when a dispute is attempted before the trade reaches a disputable state"},
   "tradeWaitingInvoiceBuyerInstruction": "Submit your Lightning invoice so the seller can lock the funds.",
--- a/lib/l10n/app_es.arb
+++ b/lib/l10n/app_es.arb
@@ -345,6 +345,7 @@
   "openDisputeFailed": "No se pudo abrir la disputa. Por favor, inténtelo de nuevo.",
   "openDisputeTitle": "Abrir disputa",
   "openDisputeConfirmation": "¿Seguro que quieres abrir una disputa? Esto escala la operación a un administrador y no se puede deshacer.",
+  "disputeAlreadyOpen": "Ya hay una disputa abierta para este intercambio.",
   "tradeNotDisputable": "Solo se puede abrir una disputa cuando los fondos ya están bloqueados para este intercambio.",
   "tradeWaitingInvoiceBuyerInstruction": "Envía tu factura Lightning para que el vendedor pueda bloquear los fondos.",
   "tradeWaitingInvoiceSellerInstruction": "Esperando a que el comprador envíe su factura Lightning.",
--- a/lib/l10n/app_fr.arb
+++ b/lib/l10n/app_fr.arb
@@ -345,6 +345,7 @@
   "openDisputeFailed": "Impossible d'ouvrir le litige. Veuillez réessayer.",
   "openDisputeTitle": "Ouvrir un litige",
   "openDisputeConfirmation": "Êtes-vous sûr de vouloir ouvrir un litige ? Cela transmet l'échange à un administrateur et ne peut pas être annulé.",
+  "disputeAlreadyOpen": "Un litige est déjà ouvert pour cet échange.",
   "tradeNotDisputable": "Un litige ne peut être ouvert qu'une fois les fonds bloqués pour cet échange.",
   "tradeWaitingInvoiceBuyerInstruction": "Soumettez votre facture Lightning pour que le vendeur puisse bloquer les fonds.",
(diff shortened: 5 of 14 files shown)


Catrya (commented):
@AndreaDiazCorreia please do rebase to fix the conflicts

Catrya (changes requested):
Reviewed at `e139591`, merged locally against current `main` (`3641db9` — the branch is 14 commits behind and merges clean). On that merge: **305 Rust tests** pass, `cargo clippy --locked -- -D warnings` and `cargo check --locked --target wasm32-unknown-unknown` — the exact CI commands — are clean. No Dart changes, and no FRB surface change (`open_dispute` keeps its signature, `record_late_acceptance` is `pub(crate)`, `dispute_id_from_payload` is private).

**Nothing here is broken.** I verified the whole path end to end by writing the tests that are missing, and they pass. Both asks below are small, and one of them is a single sentence.

## Blocking: the seam this PR creates is not covered by any test that runs the real code

The two central tests assert a *copy* of the logic rather than the logic:

- `a_cantdo_rejection_reaches_the_waiting_open_dispute` re-implements the `Action::CantDo` arm inside the test itself — `take_matching_restore(key).or_else(|| take_matching_request(key, Some(73)))`, under the comment "Exactly what the `Action::CantDo` arm does to find its caller" — and then sends the `Wake` by hand. It never calls `dispatch_mostro_message`. Add a `kind` gate to that arm and the test stays green while **#202 comes back exactly as reported**: the rejection matches no pending request and is dropped.
- `take_matching_dispute_only_consumes_dispute_records` covers the matcher, not the dispatcher's `if kind.action == Action::DisputeInitiatedByYou` arm, and not the claim that the acceptance still falls through to the status arm.

This is the one place the PR asks to be trusted on, so it should be the one place a test drives the real dispatcher. The module already has that harness: `a_late_cancel_for_a_superseded_generation_is_dropped` builds an `UnwrappedMessage` with `sender = active_mostro_pubkey()` and calls `dispatch_mostro_message` directly.

Two tests are enough, and I confirmed on your branch that both pass as written, so this is missing coverage rather than a defect:

1. **Acceptance.** Register a dispute request, seed the order as `Active`, dispatch a `DisputeInitiatedByYou` carrying that nonce and a `Payload::Dispute(uuid, None)`, then assert the waiter receives `DisputeAccepted` with the daemon's id **and** that the order ends up `Dispute` — that second assertion is what pins the "falls through to the status arm" claim the comment makes.
2. **Rejection.** Register a dispute request, dispatch a `CantDo` carrying that nonce with `Payload::CantDo(Some(NotAllowedByStatus))`, and assert the waiting caller receives `Rejected`. That is #202 itself, driven through the arm that was dropping it.

Two notes that cost me a compile each: `CantDoReason` is re-exported from `mostro_core::error`, not `::message`, and `Wake` has no `Debug`, so the non-matching arm can't format it.

## Blocking: "the whole record is purged when the trade's subscription ends" is not true in the common case

That sentence appears twice — in the doc-comment on `PendingRequestKind::Dispute { superseded }` and in `contracts/disputes.md` — and it is the justification for deliberately **not** capping the superseded list after the cap was pushed back on.

`purge_pending_request` has exactly **one production caller**: the exit of `subscribe_daemon_messages` (`orders.rs:1299`). That watcher only exists for trades created, taken or restored **in this process**. A dispute opened on a trade loaded from the database after an app restart has no per-trade watcher — its daemon messages arrive over the global `mostro-dm` feed — so its pending record, superseded list included, lives for the life of the process. Nothing purges it.

The decision itself still holds: the list only grows through user-driven retries, each gated by a 10-second wait, at eight bytes each. But the stated reason is wrong, and in `specs/` — a living contract — it reads as a bound that does not exist. One sentence fixes it in both places.

## Minor

- **`DisputeAlreadyOpen` is not mapped in `lib/core/daemon_errors.dart`.** The new single-flight refusal (`"an open_dispute for trade … is already in flight"`) falls through to the generic `openDisputeFailed`, as does the pre-existing "a dispute already exists". Not a regression — the marker was already missing — but this PR adds a second trigger for it. One line in the mapper plus a key in the five `.arb` files.
- **The `CantDo` reason is lost at the UI.** `bail!("{message}")` produces the daemon's English prose, which matches no marker, so Dart shows the generic fallback. The user learns the dispute failed but not that it was `NotAllowedByStatus` — which is precisely the #203 case this is meant to make legible. That is the debt CLAUDE.md already records ("some `CantDo` errors still return English prose directly — should become markers"), so a follow-up rather than a change here, but worth an issue.
- **A superseded nonce is never removed once it has been reconciled as `Late`.** Harmless, but the list then claims attempts are still outstanding when they are not. One line if you want it honest.

Catrya (approved):
tACK

Issue #277, opened 2026-08-05
fix(orders): order status gets out of sync (stale InProgress, dead subscription, missing kind-14 catch-up)


Catrya:
Order/trade statuses shown by the app drift from reality: a maker order
displays **Active** while it is actually in `waiting-buyer-invoice`; takes
and turn changes arrive late or never; after a reopen the status jumps
around. Reproduced on device with an instrumented build (every status write
logged with its source channel).

## Evidence — one order hits all three causes (logcat, order cd1433f7)

    12:58:07  38383-sync:  Pending→InProgress          taken
    12:58:08  kind14-sync: →WaitingBuyerInvoice        precise DM corrects, OK
    12:59:21  NewOrder (taker timed out, republished)
    ───────── 22 minutes of TOTAL silence — no events at all ─────────
    13:21:38  app restarted by hand
    13:21:39  38383-sync:  WaitingBuyerInvoice→InProgress   ← coarse clobbers precise
    13:22:21  38383-sync:  InProgress→Canceled         seller-side timeout: the
                                                       PayInvoice DM (our turn) never arrived

During the silent window the order was re-taken and the buyer sent their
invoice; the app received none of it, the seller never saw their turn to pay
the hold invoice, and the daemon canceled the order.

## Causes identified

1. **Coarse 38383 overwrites the precise status.** mostrod masks every taken
   order as `in-progress` publicly; the own-order ingest sync writes it over
   the kind-14-derived status with only a `!= Pending` guard
   (`orders.rs` `ingest_order_event`). Whichever channel lands last wins
   (~0.5–1s apart, no ordering guarantee), and on every restart the replayed
   38383 clobbers again. Observed live three times.
2. **Dead subscription is never detected.** The websocket went silent for
   22 min (no events, not even 38383 replays) with no reconnect and no
   error; only an app restart recovered. Needs a liveness mechanism
   (relay ping / silence watchdog / resubscribe on app resume).
3. **kind-14 offline catch-up delivers nothing.** The DM subscription
   deliberately has no `since`, yet on restart no missed kind-14 was
   replayed (WaitingBuyerInvoice, PayInvoice, Canceled all absent).
   Suspects: relay retention policy for DMs, or the `#p` filter with 41
   trade pubkeys exceeding relay filter limits — needs investigation.

## Proposed fixes

- (1) Guard the sync: never downgrade a more specific local status to
  `InProgress` (it only informs when the row says `Pending`); skip no-op
  writes. Small, independent, can land first.
[...]

PR #279, opened 2026-08-06
feat(logging): instrument core flows with an end-to-end debug trace

### LINKED ISSUE #241: Instrument core flows with debug logs (relay, transport, orders, db, queue)
**Type:** enhancement
**Priority:** medium

`rust/src/` has 137 log statements across ~12.5k lines (generated code excluded), and their distribution makes them close to useless for following what the app is doing:

| level | sites |
| --- | --- |
| `trace!` | 0 |
| `debug!` | 13 |
| `info!` | 32 |
| `warn!` | 79 |
| `error!` | 13 |

67% are `warn!`/`error!`, which only fire on anomalies. There is no instrumentation of the happy path, so a normal run emits a handful of lines and a trade that misbehaves leaves no trace to reconstruct after the fact.

**Scope**

Add `debug!` instrumentation to the flows that are currently opaque, with an explicit and consistent `target:` per subsystem (`relay`, `transport`, `orders`, `db`, `queue`, `nwc`) rather than relying on the module path, which today yields tags from the last `::` segment.

- `relay`: connect, disconnect, retry with backoff, pool state transitions.
- `transport`: NIP-44 wrap/unwrap and NIP-59 gift wrap/unwrap — kind, truncated event id, outcome. Never the plaintext.
- `orders`: order state transitions, message sent to the daemon, reply correlation, timeouts.
- `db`: migrations, writes, and which backend is active (SQLite vs IndexedDB).
- `queue`: enqueue, retry, backoff, drop.
- `nwc`: request/response lifecycle, excluding preimages and any secret material.

**Redaction at the source** (moved here from #240)

Instrumentation is where redaction helpers earn their place, so this issue owns the invariant:

- Helpers for truncated ids and key redaction, applied by the new statements.
- A hard rule, enforced by review: nsec, mnemonic, preimages and chat plaintext never enter a log record.
- Focused tests proving forbidden material never reaches `recent_logs()`.

**Acceptance**

- A full take-order flow leaves a readable end-to-end trace at `Debug` level.
- No secret material in any new statement, following the redaction helpers introduced by the infrastructure issue.
- `Debug` is off by default in release, so the added volume costs nothing unless the user opts in.

Depends on #240.

Catrya:
Closes #241 

## Scope

Four additive commits — log statements over existing code paths, no logic
changes:

1. **Redaction helpers + tests**: `short_id()` (8-char ids) and
   `scrub_secrets()` (masks `nsec1…`), applied once at the logger choke
   point so no sink — console, logcat, retained buffer — can leak key
   material. Tests prove forbidden material never reaches `recent_logs()`.
2. **Relay & subscription lifecycle**: relay status transitions from the
   pool monitor (gain/loss at info, retry churn at debug), subscription
   create/replace with filter shape, and the notification loop no longer
   swallows relay control messages: EOSE/auth at debug, CLOSED/NOTICE at
   warn.
3. **Wire boundaries**: per-relay OK/FAIL for every published event
   (protocol + chat envelope, metadata only), outbox
   enqueue/flush/retry/give-up, raw inbound kind-14/1059 frames at debug
   (before the SDK's first-seen dedup), and the two previously silent
   drops (unmatched p-tag → warn, duplicate → debug).
4. **Trade state transitions**: every wire→status sync logs its decision
   (`wire`, `local`, `applies`, source channel — real transitions at
   info, blocked/no-op at debug), uniform status lines for the kind-14
   arms, session create/remove, and the request-correlation story
   completed (created/resolved/timeout; latency by subtracting log
   timestamps — no struct changes).

## Out of scope

- **Dart-side log unification** (routing Dart logs into the Rust ring so
  the in-app Logs screen and `adb logcat -s mostro` carry the whole
  story): deferred until the Logs screen work happens; for development,
  `adb logcat -s mostro flutter` already shows both halves.
- A latency field on `PendingRequest`: timestamps on the created/resolved
  lines make the duration derivable without touching the struct.
- Fixes for anything the new logs reveal (e.g. the #277 kind-14 catch-up
  findings) — those go in their own PRs.

## Cost

Info-level flow lines are a handful per protocol action; everything
chatty (raw frames, EOSE, dedup, book updates) sits at debug, which is
off in release builds and costs a branch when disabled.

## Manual testing

- Desktop: `flutter run` shows the full trace (Rust via stderr).
- Android: `adb logcat -s mostro flutter -v time` (or `-s mostro:I` for
  info+ only). Create → take → invoice → release reads end-to-end.

Plus one behavior fix requested in review: publish_event_json now rejects total publish failure

## Summary by CodeRabbit

* **Bug Fixes**
  * Sensitive `nsec1` key material is now removed from live and retained logs.
  * Log messages and relay-provided text are sanitized before display.
  * Relay identifiers are shortened or filtered to avoid exposing sensitive details.

* **Improvements**
  * Added clearer diagnostics for order processing, relay activity, sessions, queue retries, and connection status.
  * Relay publishing now reports individual successes and failures.
  * Queue and wallet connection logs include safer retry, error, and request details.

### DIFF
--- a/lib/core/daemon_errors.dart
+++ b/lib/core/daemon_errors.dart
@@ -31,6 +31,11 @@ String localizedDaemonError(
   if (raw.contains('NoDaemonResponse')) {
     return l10n.sessionTimeoutMessage;
   }
+  // No relay accepted the event — it never left the device. Same remedy as
+  // a daemon timeout: check the connection and retry.
+  if (raw.contains('NoRelayAccepted')) {
+    return l10n.sessionTimeoutMessage;
+  }
   // No durable storage: no trade key can be derived (issue #249).
   if (raw.contains('StorageUnavailable')) {
     return l10n.storageUnavailable;
--- a/rust/src/api/logging.rs
+++ b/rust/src/api/logging.rs
@@ -140,8 +140,11 @@ impl log::Log for BridgeLogger {
             return;
         };
 
-        platform_console(record);
-        forward_log(record.level(), record.target(), &record.args().to_string());
+        // Format once and scrub before ANY sink sees the text, so a leaked
+        // secret can reach neither the console/logcat nor the retained buffer.
+        let message = scrub_secrets(&record.args().to_string()).into_owned();
+        platform_console(record.level(), record.target(), &message);
+        forward_log(record.level(), record.target(), &message);
     }
 
     fn flush(&self) {}
@@ -151,9 +154,10 @@ impl log::Log for BridgeLogger {
 
 /// Android discards a process's stderr, so records go to logcat. The fixed tag
 /// keeps `adb logcat -s mostro` useful; `android_logger` prepends the module
-/// path to the message.
+/// path to the message. Takes the pre-scrubbed message text (not the raw
+/// record) so redaction happens once, upstream, for every sink.
 #[cfg(target_os = "android")]
-fn platform_console(record: &log::Record) {
+fn platform_console(level: log::Level, target: &str, message: &str) {
     use log::Log as _;
 
     static ANDROID: OnceLock<android_logger::AndroidLogger> = OnceLock::new();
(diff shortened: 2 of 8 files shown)


Catrya:
Keeping Closes #241: with the nwc commit, the bulk of the issue's scope (relay, transport, orders, queue, nwc + redaction) is done. A follow-up issue will track what's deliberately deferred: db instrumentation, unifying Dart logs into the Rust ring once the in-app Logs screen work happens, and the chat publish path, left for later on purpose since the chat flow is about to change.

PR #300, opened 2026-08-13
fix(#244): show waiting spinner after NWC payment instead of re-arming the pay button

### LINKED ISSUE #244: Pay-invoice screen gives no feedback after a successful NWC payment and re-arms the pay button
`lib/features/order/screens/pay_lightning_invoice_screen.dart:137` returns a screen holding only `NwcPaymentWidget` whenever an NWC wallet is connected. The `_waiting` flag set by the payment-success callback is only read in the manual/QR branch (`:326`), so it has no effect in the NWC branch.

`NwcPaymentWidget` resets its own `_paying` flag in its `finally` block (`lib/shared/widgets/nwc_payment_widget.dart:47-49`), so right after a successful payment the user is shown an enabled "Pay with wallet" button and no confirmation of any kind.

### Impact

The user has no way to tell the payment went through, and tapping again re-sends the same bolt11. The wallet rejects the duplicate, which triggers `onFallbackToManual` and drops the user onto the QR of an already-settled invoice. No funds are lost, but the flow is confusing enough to look broken.

### Fix

Render the waiting indicator instead of the payment widget while `_waiting` is set, mirroring the manual branch. The same defect was fixed for the bond screen in #213 (`pay_bond_invoice_screen.dart`) — that patch is a direct model.

The NWC branch is also missing the cancel button, likewise fixed for the bond screen in #213. Worth deciding separately whether the escrow flow wants one: the semantics differ, since cancelling there is a real trade cancel rather than a bond back-out.

codaMW:
## Problem

On the pay-Lightning-invoice screen, after an NWC (auto-pay) payment succeeded, the pay controls stayed live instead of showing a waiting state the user could re-send an already-settled bolt11 while the daemon was still confirming. `_onPaymentDetected` set `_waiting = true`, but the NWC branch never read `_waiting`, so nothing on screen changed.

## Fix

When a payment is detected (`_onPaymentDetected`, and the manual-success path at `:143`), `_waiting` is set and the build **returns early with a full-screen waiting `Scaffold`** the `waitingForPaymentConfirmation` spinner under the standard `AppBar` **before** either the NWC or the manual-invoice/QR branch is reached (`pay_lightning_invoice_screen.dart:230`).

So once payment is detected, **no invoice-submission control survives on either path**: the QR, the NWC widget, the copy button and the pay button all disappear, replaced by the waiting spinner. The added widget test pins exactly this it asserts the NWC widget, the external pay button and the copy button are all gone.

### Scope note the manual path changes too

Previously the manual branch rendered the spinner *in context* (amount and invoice still visible alongside it). With the early return the manual path now also goes to the full-screen spinner and loses that context. This is a deliberate consequence of unifying both paths on one waiting state, and it matches the screen's other states (loading / error / invoice-guard), which all use the same full-screen `Scaffold(AppBar(...))` shape.

## Known limitations (tracked, not addressed here)

- `_waiting` is never cleared: if the daemon never flips the status, the user is left on the spinner with only the back arrow. A timeout or a guidance line would help worth a follow-up.
- Whether a cancel control belongs on this screen is a separate question (cancelling here is a trade cancel, not a payment abort) and belongs in an issue, not this PR.

## Testing

- Widget test: after an NWC success the QR, NWC, copy and pay controls are all absent and the waiting spinner is shown. Removing the `if (_waiting) return Scaffold(...)` block fails the test exactly where expected.
- `flutter analyze` clean; NWC + reputation tests green.
- A real end-to-end NWC payment isn't reproducible on the current regtest rig; the widget test covering the spinner path is the best available verification.

Closes #244.


Catrya (commented):
Hi @codaMW please rebase to fix the conflicts

codaMW:
Rebased onto current main. The only conflict was in \`test/support/fake_trades.dart\` main added reputation params (\`peerRating\` / \`peerReviews\` / \`peerDays\`) and switched \`amountSats\` to nullable-with-default; resolved to main's fixture signature. \`flutter analyze\` clean, NWC waiting/reputation tests green. No code change to the fix itself.

Catrya (changes requested):
**Changes requested.** The fix is correct and the test genuinely pins it. What needs to change is a modification to a shared test helper that is provably unnecessary, and a description that documents a different implementation from the one shipped.

## Blocking: the `fake_trades.dart` change is unnecessary

The diff touches a helper shared by four suites:

```dart
-    amountSats: amountSats,
+    amountSats: amountSats ?? BigInt.zero,
```

I reverted it and **all 313 tests still pass**. That follows: the new test already passes the value explicitly (`amountSats: BigInt.from(1000)`), so nothing needs the default.

And it is not harmless. It changes every ca `BigInt.zero`, and there is code that
distinguishes the two — `trade_state_headerats != null ? ... : ...`. No test renders
that widget with a `fakeTrade` today, so noe that does will take the wrong branch
silently.
I reverted it and **all 313 tests still passt already passes the value explicitly
(`amountSats: BigInt.from(1000)`), so nothi

And it is not harmless. It changes every ca `BigInt.zero`, and there is code that
distinguishes the two — `trade_state_headerats != null ? ... : ...`. No test renders
that widget with a `fakeTrade` today, so noe that does will take the wrong branch
silently.

The justification in the description does n **gains** an optional `amountSats` param …
(existing callers default to zero, unchange **already exists** (`fake_trades.dart:14`),
and existing callers did **not** default toll`. Both halves of that sentence are
wrong.

**Ask: drop that hunk.** The test works wit

## Blocking: the description does not descr

The body says:

> Gate the NWC branch on `_waiting` as wellalMode && !_waiting`. Once a payment is
detected, the NWC branch is skipped and **tinvoice/QR layout, which already renders the
`waitingForPaymentConfirmation` spinner in h is unchanged.**

The code does not do that. Commit `6aaf68b`is now an early `return Scaffold(...)` with
a full-screen spinner, **before both branchdy does not mention:

- It does not fall through to the QR layoutscreen.
- **The manual path does change.** Previousranch rendered the spinner *in context*
(`:456`), keeping the amount and the invoic the cancel button. Now the manual path
loses that context too.

I am not arguing the new approach is worse e no resend control survives at all, QR andcopy included. I am arguing the body describes the first attempt and claims a path is untouched when it is not. That text is what stays in the record when #244 is closed.

## What I verified and what holds up

**The test is load-bearing.** I removed thefold(...) }` block — reverting the fix — and it fails exactly where it should:

```
Expected: no matching candidates
  Actual: _TypeWidgetFinder:<Found 1 widget with type "NwcPaymentWidget">
```

And it does not stop at the NWC widget: it external button and the copy button are
gone. That covers the whole resend surface,used the report.

**The diagnosis is right.** `_onPaymentDete=> _waiting = true)` (`:52-53`), and on`main` the NWC branch (`:224`) never read `_waiting`. The bug was real.

**The new screen is consistent with the rest.** All four states of this screen (loading, error, invoice guard, and
now waiting) use the same `AppBar(title: ..either a different pattern nor a dead end
the other states did not already have.

## Minor

1. **`_waiting` is never cleared.** No pathh the spinner now taking the whole screen, a payment that succeeds while the daemon never flips the status leaves the user on a screen with no action but the back arrow — previously, on the manual path, they at least still saw the invoice. Not a regression introduced here
(the manual branch's `_waiting` already rep the stakes are higher now: a timeout, or aline explaining what to do if it takes long, would help.

2. **The open question about the cancel button should be closed, not left in the PR.** I agree with leaving it out —
cancelling here is a trade cancel, not a boon belongs in an issue, not in a paragraphof a PR that is about to merge.

## What I could not verify

- **A real end-to-end NWC payment.** The bo with the limitation: it is not reproducible
on the current regtest rig. The widget testinner path, which is the best available.
- **What happens if `onFallbackToManual` fi* By construction it should not (success and
failure callbacks are exclusive), but I did

With those two resolved — drop the `fake_tr description up to date — this goes in asis.

Catrya:
Superseded by https://github.com/MostroP2P/app/pull/563

PR #303, opened 2026-08-18
feat(#272): push-first trade status, replacing the 2 s poll

### LINKED ISSUE #272: refactor(trades): migrate trade status UI from 2s polling to the on_trade_updated push stream
Trade status in the UI is poll-driven: every visible My Trades row chip and
the trade detail screen run `tradeStatusProvider`, which calls the Rust
bridge every 2 s per subscriber (`trade_state_provider.dart:34`). This has
two costs:

- **Latency ceiling**: any status change (e.g. an order returning to pending
  after a taker timeout) can take up to 2 extra seconds to paint even though
  the event is already in the order book.
- **Overhead**: N visible rows × one FRB call every 2 s, noticeable on
  phones — pure waste when nothing changed.

The push infrastructure already exists: `on_trade_updated()` (added for the
timeout-cleanup work) emits `TradeUpdate { order_id, status }`, but only the
`Canceled` handler and the stale-state sweep emit today. The 004 contract
already envisioned this as `on_order_status_changed()` — defined in
`contracts/orders.md`, never implemented.

## Scope

- Rust: extend `emit_trade_update` to every status transition — the
  status-sync gift-wrap arms, `PayInvoice`, the peer-pubkey/active arm, and
  the Kind 38383 ingest sync for own orders.
- Dart: rework `tradeStatusProvider` to be push-first (initial fetch +
  stream updates filtered by `order_id`), keeping a low-frequency poll only
  as a reconnection fallback — consumers (row chips, trade detail,
  pay-invoice listener, badge counter) should not need changes.
- Update `contracts/orders.md`: fold `on_order_status_changed` into the
  implemented `on_trade_updated` (one stream, filter client-side).

codaMW:
## Problem
`tradeStatusProvider` polled `getOrder()`/`listTrades()` every 2 seconds to surface a trade's live status. The `on_trade_updated` push channel already existed and the Kind 14 dispatch arms already emitted on it, so the poll added latency for no benefit and it masked a gap: a **client-initiated cancel** updates local state optimistically and never comes back through a dispatch arm, so nothing was emitted.

## Dart push-first `tradeStatusProvider`
- Bridges the shared `tradeUpdatesProvider` (one relay subscription feeds every watched trade) plus a periodic reconnection-fallback tick into a **single** event stream, drained by one `await for`. A push carries the new status directly; a tick triggers a reconciliation fetch. The fallback interval drops from 2 s to 30 s since pushes now carry the real-time signal.
- Preserves the immediate first emission, terminal short-circuit, and DB fallback for orders wiped from the in-memory book. The provider signature is unchanged, so consumers (`my_order_screen`, `pay_lightning_invoice_screen`, the trades list) need no changes.
- The reconciliation fetch is now failure-tolerant: a bridge/DB error yields `null` rather than tearing down the whole status stream; the next push or tick recovers.

## Rust one client-initiated emit gap
`cancel_order()` optimistically writes `Canceled` and removes the order from the book but never emitted, so a push-first listener would miss it (the order is gone nothing to reconcile against and the daemon's gift-wrap confirmation may arrive much later or never). It now emits `Canceled` right after the optimistic update. `release_order()`/`send_fiat_sent()` publish and wait for the daemon, whose reply already emits through the dispatch arms, so there is no gap there. The daemon-driven arms and the stale-sweep were already emitting; this is the only addition.

## Docs
Folds `on_order_status_changed` into `on_trade_updated` in the orders contract a separate per-order status stream isn't implemented; `tradeStatusProvider` consumes `on_trade_updated` filtered by `order_id`.

## Testing
- `flutter analyze` clean; the existing `trade_action_listener` and `trade_detail_screen` suites (which consume these providers) pass.
- **Device (Nokia C31):** cancelling a trade emits a push that flips the UI to Canceled immediately, not on the next poll. Idle fallback cadence is one tick per watched order every 30 s, confirming the merged stream doesn't busy-loop.
- No automated provider test was added: `tradeStatusProvider` is an infinite-loop stream that resists unit testing without `fakeAsync` scaffolding the codebase doesn't currently use for its sibling polling providers (`tradeAmountProvider`, `tradeHoldInvoiceProvider`). Verification is on-device, matching the existing convention for these providers.

## Note for reviewers
This touches `trade_state_provider.dart`, which my open #299 also builds on (that PR buckets the trades list by live status via `tradeStatusProvider`). The two compose #272 changes how the provider sources its data, #299 how the list consumes it but a rebase may be needed depending on merge order.

Closes #272.

## Summary by CodeRabbit

* **New Features**
  * Trade status updates now appear immediately through push notifications.
  * Status changes are deduplicated and stop automatically once an order reaches a final state.
  * Order cancellations notify connected clients immediately.

* **Bug Fixes**
  * Added a low-frequency fallback to reconcile status when push updates are unavailable.

* **Documentation**
  * Updated guidance to use the trade update stream for order status changes.


codaMW:
@coderabbitai full review

codaMW:
Thanks addressed:

Doc (optimistic cancel): updated the on_trade_updated contract to note that cancel_order emits an optimistic Canceled immediately after the client action before daemon confirmation, and even if the local DB write fails and that consumers must not treat every Canceled as daemon-confirmed.
Rust test: added client_cancel_emits_canceled_update, verifying the emit contract cancel_order relies on. cancel_order itself isn't unit-testable wholesale (it needs trade keys, identity, and the relay), so the test targets the emit_trade_update call it makes after its optimistic update. It filters the process-wide broadcast by a unique order id to stay robust to concurrent tests' emits passes in isolation and in the full suite, clippy clean.

Dart provider test: I've left tradeStatusProvider without a unit test, consistent with its three sibling polling providers in the same file (tradeAmountProvider, tradeHoldInvoiceProvider, tradeInfoStreamProvider), which are likewise untested they're infinite-loop StreamProviders that resist unit testing without fakeAsync scaffolding the project doesn't currently use. I verified this provider on-device instead, which caught two real issues a unit test would have missed (a busy-loop in an earlier StreamIterator-based draft, and the cancel_order emit gap this PR fixes). Happy to add a fakeAsync-based test if you'd prefer just flagging the tradeoff and the existing convention first.

Catrya (changes requested):
Changes requested — two small things, neither a defect

The design is sound and three of its decisions are worth calling out, because each closes a trap that's easy to leave open:

- The `StreamController` is created **before** the first status fetch, so a push arriving while the initial status resolves is queued rather than lost.
- Pushes and fallback ticks are merged into **one** stream drained by a single `await for` — no abandoned `moveNext()` futures.
- The reconciliation fetch swallows errors into `null` instead of tearing down the stream, which also lets the provider run without `RustLib.init()`.

Two things before it lands. Neither is a defect.

## 1. Cancel the ticker before the terminal `return`

When the status is terminal the generator returns, but the `Timer.periodic` and the `ref.listen` callback keep pushing into a controller nobody drains until the provider is disposed. Nothing accumulates — the events are discarded with no subscriber — but the timer keeps waking every 30 s for nothing.

It also compounds with #299, which watches every trade in the list including terminal ones: each would hold a pointless timer for the life of the process. One line.

## 2. One automated test for the push path

The reason given for not adding one — an infinite-loop stream that resists testing without `fakeAsync` — was true of the *previous* provider, and this redesign removed it, as a side effect of changes made for other reasons:

- The provider no longer polls the bridge in a loop; it listens to `tradeUpdatesProvider`, and any Riverpod provider can be replaced in a test with `overrideWith`. That seam was added for the shared relay subscription, but it makes the push path injectable.
- `_currentStatus` now returns `null` on failure instead of propagating, which was done so a transient bridge error wouldn't kill the stream — and which also lets the provider run without `RustLib.init()`.

So a test can override the updates provider, push one `TradeUpdate`, and assert the emitted status, never reaching the 30 s tick. That covers exactly what this PR introduces.

This replaces the mechanism behind every trade-status surface in the app and its only verification today is one device, so one push-to-emission test is worth having before it lands.

Two further steps, **not** requested here, for whenever the testability of these providers gets its own pass — ideally across all three siblings at once, since `tradeAmountProvider` and `tradeHoldInvoiceProvider` have the same shape:

- Make the fetch injectable so the first emission, the `status != last` dedup and the terminal short-circuit become testable too. #266 set the pattern in this repo: constructor params defaulting to the real bridge functions (`getConfirmed ?? identity_api.getBackupConfirmed`). Here the equivalent would be a small `statusFetcherProvider` returning `Future<OrderStatus?> Function(String)`.
- Make `_reconnectPoll` injectable so a test can set it to milliseconds and exercise the reconciliation tick without `fakeAsync`.

## On the optimistic `Canceled` emit

Worth a line in the description so it isn't read as "the cancel is confirmed". `cancel_order` already wrote `Canceled` to the DB optimistically and removed the order from the book, without distinguishing the cooperative case — even though its own doc says *"Both parties must cancel for it to take effect"*. So on a cooperative cancel the local state claims canceled while a counterparty confirmation is still pending.

That state already reached the UI before this PR (the old poll fell through to the DB and showed `Canceled` within 2 s); this just makes it instant. Not a regression, but it makes something wrong more visible. #125 (*Cooperative cancel UX*) is where it belongs — it describes this exact gap: *"the Trade Detail UI only maps to terminal status. While a coop cancel is in flight, the UI does not show whose confirmation is pending"*.

## Merge order with #299

Worth deciding now: **#299 benefits considerably from this landing first.** My main objection there is that it pins `tradeStatusProvider` for every trade in the list, which under the 2 s poll means N bridge calls every two seconds, permanently. With this merged, the same fan-out becomes one tick per order every 30 s and that objection largely dissolves.

codaMW:
Both done.

**1. Ticker cancelled before the terminal return.** Extracted the teardown (close the listener, cancel the ticker, close the controller) into an idempotent \`stop()\`, used by \`onDispose\` and called at both terminal returns. A terminal trade no longer leaves the \`Timer.periodic\` waking every 30s until disposal which, as you noted, compounds with #299 watching every trade in the list.

**2. Push-path test added.** Overrides \`tradeUpdatesProvider\`, pushes one \`TradeUpdate\`, and asserts \`tradeStatusProvider\` emits the mapped status no \`RustLib.init()\` (\`_currentStatus\` swallows the bridge error into null, so the immediate emission is skipped) and never reaching the 30s tick. Added a second case asserting a different order's update doesn't surface here (the \`orderId\` filter in the \`ref.listen\` callback).

The two further steps you flagged (injectable fetch + injectable \`_reconnectPoll\`) I've left for the dedicated testability pass across all three sibling providers, as you suggested. The optimistic-\`Canceled\` doc note is already in from the earlier round.

Rebased onto current main (clean). \`flutter analyze\` clean, provider tests pass.

Catrya (changes requested):
Re-reviewed at `6ed6cf34` against current `main` (`7cf1f0d`). No new commits since my last round — the head is unchanged; what moved is the base (#376 landed, bumping mostro-core 0.14.1 → 0.14.6), which is why GitHub is showing `mergeable: unknown`. I re-ran everything on the new base in case the bump moved something: merge is clean, `cargo test --locked` **340 passed**, `cargo clippy --locked -- -D warnings` clean, `cargo check --locked --target wasm32-unknown-unknown` clean, `flutter analyze` clean, `flutter test` **315 passed**.

## Both of my previous asks are done

| Ask | Status |
|---|---|
| Cancel the ticker before the terminal `return` | ✅ `stop()` is idempotent, called before both returns, and registered with `ref.onDispose` |
| One automated test for the push path | ✅ and they are load-bearing — verified by mutation: dropping the `u.orderId == orderId` filter fails the negative test; not bridging pushes fails the positive one |

CodeRabbit's third point (document that a `Canceled` may be optimistic and not daemon-confirmed) is covered in `contracts/orders.md`.

**Changes requested**, for the Rust half only.

## Blocking: the Kind 38383 ingest does not emit, and the issue asks for it

#272's scope says:

> **Rust**: extend `emit_trade_update` to every status transition — the status-sync gift-wrap arms, `PayInvoice`, the peer-pubkey/active arm, **and the Kind 38383 ingest sync for own orders**.

The body answers:

> The daemon-driven arms and the stale-sweep were already emitting; **this is the only addition**.

The first three were indeed already there — `HoldInvoicePaymentAccepted` (`:2042`), `AddInvoice` (`:2106`), `PayInvoice` (`:2184`), plus the `PaymentFailed` arm (`:2242`). **The fourth is not.** On the merged tree there are 14 `emit_trade_update` call sites in `orders.rs` and **zero** inside the *"Sync trade status in DB for own orders"* block: it does `log_wire_status_sync` and `update_trade_fields`, and never emits.

The consequence is a latency regression, not data loss: a transition that arrives only as a public 38383 event, with no accompanying kind-14, used to reach the UI within the 2 s poll and now waits for the 30 s reconciliation tick. The fallback does recover it — `_currentStatus` reads the book — it just takes fifteen times longer in the worst case.

The gap may well be narrow in practice: a take and a timeout republish both produce kind-14s that do emit (`:2876`, `:2893`). But then **that is what should be written down** — why scope item four is unnecessary. Closing #272 with it unimplemented and unmentioned is a scope claim that closes an issue while part of it stays open.

## Blocking: the one Rust line this PR adds is uncovered

I removed `emit_trade_update(&order_id, OrderStatus::Canceled)` from `cancel_order` — reverting the entire Rust half — and **all 340 tests still pass**.

`client_cancel_emits_canceled_update` calls `emit_trade_update` **directly**, not `cancel_order`. Its docstring is honest about why (*"cancel_order itself needs trade keys, identity and the relay"*), and I have no problem with the limitation. What I do have a problem with is the body listing it under the review round as "add emit test": it tests the helper, not the wiring, and the wiring is the only thing this PR adds in Rust.

Either extract the decision into something testable, or say plainly in the body that this line rests on the on-device verification alone. As written it implies coverage that does not exist.

## Minor

1. **The `stop()` from the last round is also uncovered.** I removed it from the terminal `return` and every Flutter test passes. I am not asking for a test — proving a `Timer` stopped waking needs `fakeAsync`, which the PR correctly notes this codebase does not use — but it is worth knowing that line can vanish in a refactor unnoticed, and it is the one that keeps the problem from multiplying with #299.

2. **No monotonicity guard in the merged stream.** The drain emits anything satisfying `status != last`, in both directions. Two narrow but new exposures: the `StreamController` is created before the first fetch (good — no lost push), but if **two** pushes land in that window the drain emits the older one first even though the fetch already returned the newer; and a reconciliation tick reads the book, whose status can be coarser than one a push carried. The Rust side has `wire_status_applies` and `status_sync_blocked_by_terminal` precisely because walking a status backwards is a known hazard (#203); this provider applies pushes blindly. The terminal case is protected (it returns). A "never go backwards" comparison would be cheap.

3. **The title promises more than it delivers**: *"replacing the 2 s poll"*. `tradeStatusProvider`'s poll is replaced, but `tradeHoldInvoiceProvider` and `tradeInfoStreamProvider` still run **two full `listTrades()` per second** on the pay-invoice screen (`:210`, `:232`), untouched. That is outside #272's scope and fine to leave, but the title reads as if polling is gone.

codaMW:
Both Rust blockers addressed.

38383 emit (scope item 4). You're right the sync block updated the DB but never emitted, so a 38383-only transition waited for the 30s tick. Added emit_trade_update(&info.id, info.status) after the DB sync, gated on the same wire_status_applies forward-only guard the write uses (already unit-tested), so it only emits own-order forward transitions and never walks backwards. Corrected the body scope item 4 is now actually implemented, not claimed.

Uncovered Rust line. Fair. The cancel_order emit and this new 38383 emit both run inside handlers that need a live relay, so the wiring rests on device verification I've said that plainly in the body rather than implying client_cancel_emits_canceled_update covers it (it tests the helper). The forward-only decision the 38383 emit is gated on is unit-covered via wire_status_applies.

Monotonicity (minor). Acknowledged. Dart has no wire_status_applies equivalent, and OrderStatus's enum order isn't a valid lifecycle rank (inProgress/dispute sit after the terminals), so a correct "never go backwards" guard needs a hand-built rank map not the cheap comparison it first looks like. The terminal backward-walk (the real hazard) is already blocked by the return on terminal. I'd rather track the full non-terminal ordering than ship a fragile enum-order guard happy to open a follow-up.

Title. Fair scoped the description to note tradeStatusProvider's poll is replaced while tradeHoldInvoiceProvider/tradeInfoStreamProvider still poll (out of #272's scope).

Catrya:
Closing this as **superseded**: `main` (PR #492 ) now drives the trade status from a push,
and the 2 s poll this PR set out to remove is gone. 

## Why

`lib/features/order/providers/trade_state_provider.dart` no longer polls at all
— zero occurrences of `Duration(seconds: 2)`. `tradeStatusProvider` goes through
`_followTrade` (`:54`), which re-reads the status when Rust rings a doorbell:

```dart
/// The trade doorbell pushed from Rust (`api::trade_touch`): "this order's
/// entry or row was written — read it again".
final tradeTouchProvider = StreamProvider.autoDispose<TradeTouch>(…);   // :28
```
with a safety poll every 30 s (tradeSafetyPollInterval, :22) instead of
every 2.

The Rust half of #272 is covered as well, including the item the last review
round blocked on — an emit from the Kind 38383 ingest for own orders. That path
writes through sync_trade_fields_if_changed (rust/src/api/orders.rs:5841),
which ends in crate::api::trade_touch::touch_trade(order_id), so a
38383-only transition reaches the UI without waiting for a fallback tick.

It landed as part of the push series (#488, #492, and the docs in #486), not as
this branch.

State of the branch

613 commits behind main, conflicting in trade_state_provider.dart,
rust/src/api/orders.rs and test/features/order/providers/trade_state_provider_test.dart.
The provider it rewrites has itself been rewritten since.

Nothing is lost

The two findings from the review rounds — stopping the fallback ticker on a
terminal status, and the optimistic Canceled that cancel_order writes before
the daemon confirms — are documented in contracts/orders.md and are part of
how the push path behaves today. I am closing #272 alongside this.

@codaMW thanks for the diagnosis and for the rounds on it; the shape main
converged on is the same one, only pushed from Rust rather than pulled from Dart.

Issue #270, opened 2026-08-05
feat(order): waiting-state countdown driven by the node's real expiration_seconds


Catrya:
The waiting-state countdown is cosmetic and wrong on two counts:
- `trade_detail_screen.dart:41` hardcodes 900 s instead of using
  `expiration_seconds` from the node's Kind 38385 instance event. That value is
  already parsed in Dart but only displayed on the About screen;
  `MostroNodeInfo` in Rust (`types.rs:376`) is documented as unused.
- The countdown source is `OrderInfo.expiresAt` — the 24 h pending-order
  expiration tag — not the waiting-state deadline. v1 computes it as
  *timestamp of the state-change message + expiration_seconds*
  (see `.specify/v1-reference/TIMEOUT_DETECTION_AND_SESSION_CLEANUP.md`).
  `TradeInfo.timeout_at` is written on take (`orders.rs:1113`) but never read.

## Scope

- Feed `expiration_seconds` from the instance event into the countdown
  (fallback 900 when the daemon omits it).
- Base waiting-state countdowns on the state-change deadline (`timeout_at` or
  message timestamp + expiration_seconds), keeping `expires_at` for pending.
- Show the remaining time on Take Order / trade detail so users know how long
  they have before the taken order is auto-canceled.
- UI only informs — the daemon remains the authority on expiry (no local
  cancellation at zero).

codaMW:
Hi @Catrya happy to take on this one if not assigned to anyone yet. Thank you

PR #306, opened 2026-08-20
feat(#270): drive waiting-state countdown from expiration_seconds + timeout_at

codaMW:
## Problem

The waiting-state countdown was cosmetic and wrong on two counts:

1. It hardcoded `900 s` (`trade_detail_screen.dart`) instead of the node's
   advertised `expiration_seconds` from the Kind 38385 instance event a value
   already parsed in Dart but only shown on the About screen.
2. It counted down to `OrderInfo.expiresAt` (the 24 h pending-order expiry), not
   the waiting-state deadline. `TradeInfo.timeout_at` is written on take but was
   never read.

## Fix

Deadline selection is now a shared pure helper, `waitingCountdownDeadline`
(`lib/features/order/utils/waiting_countdown.dart`), so every surface agrees:

- **Pending** -> counts to the 24 h pending expiry (`expiresAt`), unchanged.
- **Waiting states** (buyer-invoice / payment) -> counts to the state-change
  deadline: the trade's `timeoutAt` when the daemon persisted one, else
  `now + expiration_seconds`, falling back to `900 s` only when the node omits it.
- **Other states** -> no countdown.

Both data sources were already available in Dart, so there's **no FRB regen**:
`expirationSeconds` on `MostroInstance` (via `mostroNodeProvider`) and
`timeoutAt` on `TradeInfo` (via `tradeInfoProvider`).

UI only informs the daemon stays the authority on expiry (no local
cancellation at zero).

## Scope note

The issue names Take Order / trade detail. While verifying, I found the **chat
trade-state header** (`trade_state_header.dart`) had the identical bug it fed
raw `order.expiresAt` to its countdown chip and showed ~30 days (`719:55:19`)
for a waiting-payment order. Since it's the same defect and a half-fix would be
inconsistent, I extracted the logic into the shared helper and fixed both
surfaces. Happy to split the chat-header change out if you'd prefer it narrower.

## Testing

Device-verified on hardware (Nokia C31):
- About advertises `expiration_seconds = 900`.
- A waiting-for-payment order shows **15:00** counting down on **both** the
  trade-detail screen and the counterpart chat header matching, and matching
  the node's advertised window.
- Pending orders still reflect the 24 h pending expiry on both surfaces.

`flutter analyze` clean.

## Summary by CodeRabbit

* **New Features**
  * Added accurate waiting countdowns for pending orders, invoices, and payment states.
  * Countdown deadlines now reflect live trade status, timeout data, node settings, and pending expiry.
  * Countdown indicators appear only when a valid deadline is available.

* **Bug Fixes**
  * Improved countdown updates when trade deadlines change.
  * Added a fallback duration when expiration settings are unavailable.


codaMW:
@coderabbitai full review

codaMW:
Thanks, all three addressed in the latest commit.

**Fallback drift (major):** good catch. Anchored the fallback on \`TradeInfo.startedAt\` + window instead of \`now + window\`, so the helper is now a pure function of its inputs the deadline no longer slides forward on the per-second rebuild. Added a regression test asserting stability across repeated resolutions. Returns null when neither \`timeout_at\` nor a start anchor is present.

**Snapshot status (minor):** the chat header now passes \`liveStatus ?? order.status\`, mirroring the pill, so the countdown shows during the provider's loading frame.

**Clear on non-countdown transition (major):** trade detail now clears the countdown when the resolved state returns null (active / fiat-sent / terminal), so no stale countdown lingers.

Added \`waiting_countdown_test.dart\` covering all branches. \`flutter analyze\` + \`flutter test\` green."

Catrya (changes requested):
Reviewed at `f2bd61e`. The branch is 45 commits behind and **conflicts** with `main` in `trade_detail_screen.dart` — mechanically, because #320 added a `ref.watch(tradeInfoProvider(...))` at the same anchor. I resolved it locally by keeping both sides and unifying the variable name; on that merge `flutter analyze` is clean and **250 Dart tests** pass.

## What holds up

The helper is well built: pure, no clock read, and anchored on a fixed timestamp so the deadline doesn't slide forward on every per-second rebuild — with a test pinning exactly that. Eight tests cover its branches. `_applyDeadline` only resets the timer when the target actually changes. And the premise is right: both inputs are already in Dart, so no FRB regen.

## Blocking: neither input carries what the helper assumes

**`timeoutAt` is a hardcoded 900, now in Rust.** It is written in exactly one place, `orders.rs:774`, inside `take_order`:

```rust
timeout_at: Some(now + 900),
```

So on the helper's primary branch — "the trade's `timeoutAt` when the daemon persisted one" — the node's `expiration_seconds` is **not used at all**. The headline only holds on the fallback branch. Two consequences:

- The field is never rewritten, so when the trade moves from `waiting-buyer-invoice` to `waiting-payment` the target is still `take + 900`: the second window opens with a countdown that has already run out.
- It isn't persisted by the daemon either — the client writes it on take. The doc comment says otherwise.

**The fallback anchor is the wrong timestamp.** With `timeoutAt` null the helper uses `waitingSinceEpoch + window`, and the caller passes `TradeInfo.startedAt`. For a maker that is the **order creation time** (`orders.rs:546`, alongside `timeout_at: None`), not the moment the trade entered a waiting state. A pending order that sits in the book for three hours before being taken yields a deadline `creation + 900 s` — over two hours in the past. The countdown is born at zero.

That is the central case of this PR: on a sell order the maker created, once it is taken the **maker** is the one watching the `waiting-payment` countdown to pay the hold invoice. The path with `timeout_at: None` and a creation-time `startedAt` is the user this was meant to help.

## How v1 does it

`trade_detail_screen.dart:1037-1126` in `mobile` resolves all three points explicitly.

**The anchor is the message that produced the state**, not a field on the trade:

```dart
final stateMessage = _findMessageForState(messages, status);
if (stateMessage?.timestamp == null) return null;      // no anchor, no countdown
final expSecs = mostroInstance?.expirationSeconds ?? 900;
final messageTime = DateTime.fromMillisecondsSinceEpoch(stateMessage!.timestamp!);
final expiration = messageTime.add(Duration(seconds: expSecs));
```

It also guards the edges: timestamp ≤ 0, a message more than an hour in the future, and for pending an expiry more than an hour in the past — each returns no countdown.

Two things to take from it:

1. **The anchor isn't portable as-is, which is why the fix belongs in Rust.** v1 recomputes it at render time from persisted message history; this app doesn't have that — the `messages` table holds chat (`MessageType.peer`) and daemon kind-14s are dispatched without being stored. The equivalent here is to **stamp `timeout_at` when the trade enters a waiting state**, using the node's advertised `expiration_seconds` instead of a fixed 900. That is the same value v1 recomputes, recorded once in the layer that receives it. The Dart helper then collapses to "use `timeoutAt`", and both the hardcoded window and the anchor problem disappear.

2. **"No anchor, no countdown" is portable today.** Rather than falling back to `startedAt`, return null. It costs a countdown on that path and removes the maker bug immediately — and once Rust stamps the field, the path stops existing. A missing countdown is better than one that starts at zero.

## Minor

**`totalWindowSeconds` for `pending`** returns the waiting window (900 or the node's value) while the deadline points 24 h out, so the progress ring is meaningless in that state. It was equally wrong before, but v1 shows the shape: it uses a *different* widget for pending, `DynamicCountdownWidget(expiration, createdAt)`, whose total is `expiration - createdAt`. Now that one helper decides both, returning `deadline - createdAt` for pending is nearly free.

**Keep the chat-header fix in.** It's the same defect, and fixing one surface would leave the two disagreeing. The offer to split it out isn't needed.

codaMW:
Addressed, and thank you for the precise diagnosis, the maker case was exactly the hole.

**Fallback anchor (the central bug).** You're right that \`startedAt\` is order-creation time, so for a maker whose order sat in the book before being taken it produced a deadline already in the past the countdown born at zero, and that maker watching the waiting-payment countdown is precisely who #270 was for. I took the portable half you pointed to: with no \`timeout_at\` the helper now returns null (no anchor, no countdown) rather than falling back to \`startedAt\`. A missing countdown beats one that starts at zero, and the path disappears once the daemon stamps the field.

**The real fix belongs in Rust.** \`timeout_at\` is a hardcoded 900 written client-side on take (orders.rs:774), never rewritten and not persisted by the daemon so on the primary branch \`expiration_seconds\` isn't used, and the buyer-invoice -> payment transition opens an already-expired window. The correct fix stamp \`timeout_at\` on waiting-state entry from the node's \`expiration_seconds\` is a daemon change; I'll open a follow-up on the daemon repo and link it here so the Dart helper can collapse to \"use timeoutAt\" once it lands.

**Pending ring.** Now spans creation -> 24 h expiry (\`deadline - createdAt\`) instead of the waiting window, so the ring is meaningful.

**Doc comment** corrected (fixed 900, client-written on take, not daemon-persisted). **Chat-header fix kept** same defect on both surfaces.

Rebased onto current main (the trade_detail conflict with #320's \`tradeInfoProvider\` watch resolved to a single shared watch). Tests rewritten: the born-at-zero cases became a null-return regression guard for the maker path; the stability test now pins the \`timeout_at\` branch. \`flutter analyze\` clean, full suite green.

Catrya:
Superseded by https://github.com/MostroP2P/app/pull/564

PR #307, opened 2026-08-23
feat: Mortsom automation contract

grunch:
## What this is

The automation contract that lets [Mortsom](https://github.com/MostroP2P/mortsom) — the
end-to-end harness — drive this app the way a user does: through stable identifiers,
never through visible copy. The classic app has carried it since MostroP2P/mobile#650;
this is the same mechanism, applied to v2's own screens.

An automated driver cannot search for "Continue" or "Skip": the text changes with the
locale, with a redesign, with a copy review. It needs identifiers the app promises to
keep. That promise is what this branch adds, plus a test that keeps it honest.

## What it adds

- **`AutomationIds`** — one registry of the identifiers the harness may use, and
  `.withAutomationId()`, which attaches them as `Semantics.identifier` (the Android
  `resource-id`). An identifier is a property of a control, so it applies at the end of
  the expression rather than wrapping it: naming a control does not re-indent its
  subtree. Applied to onboarding, navigation and the drawer, key management, settings,
  the node selector, relays, NWC, the order book, order creation, take order, trade
  actions, invoices and payments.
- **A test-environment entry point** (`lib/main_mortsom.dart`) and a visible banner, so a
  build under test is never mistaken for a real one — by a person or by the harness,
  which refuses to run against a build without the marker.
- **State readouts** the harness asserts on (`order.status`, `order.id`,
  `keys.public_key`, `wallet.connection`, the invoices) rather than inferring them from
  screenshots.
- **A contract test** (`test/core/automation/automation_contract_test.dart`) that fails
  the build when the contract rots.
- **`docs/automation-contract.md`**, the document the contract lives in.

## Three shapes, and picking the wrong one is how this rots

| Shape | How | Why |
|---|---|---|
| A control | `.withAutomationId(id)` | The id, the visible label, the enabled flag and the tap action merge onto one node — what the accessibility bridge exposes. |
| A state readout | `.withAutomationId(id, label: machineValue)` | The descendants stop contributing semantics, so `order.status` reads `waiting-payment` and not that value with the localized pill copy appended. |
| A composite row | `.withAutomationId(id, merge: false, label: ...)` | A relay row has a toggle and a delete button; merging would collapse them and automation could no longer choose which relay to remove. |

The contract test proves each of these. Writing it caught two real bugs before this
branch left my machine: a readout whose label came out as `waiting-payment\nEsperando
pago…`, and an identifier declared and attached to nothing.

## The test environment

```sh
flutter build apk -t lib/main_mortsom.dart \
  --dart-define=MORTSOM_TEST_ENV=true \
  --dart-define=MOSTRO_PUB_KEY=<daemon pubkey> \
  --dart-define=MORTSOM_RELAYS=ws://10.0.2.2:7000
```

It is active only when arming and the compile-time define agree. `lib/main_mortsom.dart`
is the only caller of `arm()`; production never arms it and the release pipeline never
passes the define. `MORTSOM_RELAYS` **replaces** the relay defaults compiled into the
Rust core rather than extending them — a run whose local relay is unreachable must fail,
not quietly pass against a public relay — and a build without it fails at startup naming
the define. The add-relay dialog accepts `ws://` there, and only there.

Startup moved to `lib/core/app_bootstrap.dart` so both entry points share it. A second
copy of 180 lines of startup would drift, and the two builds would then start the app
differently — exactly what an end-to-end harness must not have.

## The one product change

The account screen gains a public-key row. Mortsom reads the identity it is driving from
there, and a user could not otherwise tell which account was active without revealing the
secret words. There is deliberately **no** identifier for the mnemonic itself: a stable
readout would put the seed phrase in the accessibility tree, where any accessibility
service on the device can read it, and no scenario needs it.

## Behaviour

Otherwise none changed. The identifiers make existing controls addressable; the entry
point, the relay seed and the banner apply only to a build that carries the define.

## Test plan

- [x] `flutter analyze` clean on `lib/`
- [x] `flutter test` — 223 tests, all passing
- [x] The automation contract test passes (12 tests)
- [x] Mortsom's `selectors_match_the_v2_app_contract` passes against this checkout: every
      identifier the harness uses exists here
- [ ] A complete trade end to end on two emulators (needs the matching Mortsom branch,
      MostroP2P/mortsom#1)

https://claude.ai/code/session_01EycMBjFyEwZdbTyZQneY1x

## Summary by CodeRabbit

- **New Features**
  - Added stable accessibility identifiers across navigation, onboarding, account, wallet, order, payment, and trade workflows.
  - Added machine-readable readouts for invoices, order details, wallet status, and trade statuses.
  - Added a visible test-environment indicator and dedicated test startup configuration.
  - Centralized application startup for consistent production and test launches.

- **Documentation**
  - Added an automation contract covering identifiers, semantics, test setup, and security requirements.

- **Tests**
  - Added comprehensive validation for automation identifiers, semantics, dynamic values, and test-environment behavior.


grunch:
Thanks both — seven findings, six fixed, one skipped with a reason. Pushed as bbf1016.

### Codex, `MOSTRO_PUB_KEY` (P1) — fixed, and it was the important one

Correct, and it would have broken the first real run. The build command this branch
documents advertises the define and nothing read it, so every subscription before the
harness reached settings was addressed to the compiled-in production node, which cannot
decrypt them. The app would have sat there looking idle for a reason no log names.

`TestEnvironment.mostroPubkey` now reads it, and `bootstrapAndRun` applies it through
`settings_api.setActiveMostroNode` **before** the relay pool starts, so the very first
subscription already lands on the daemon under test. Two constraints on top of the
suggestion:

- **Only when no node was ever chosen.** A scenario may switch nodes through the UI, and
  a restart mid-scenario must not silently undo that.
- **A malformed define is ignored**, not forwarded. Handing a bad key to the bridge
  surfaces as an opaque bridge failure rather than the build mistake it is.

### CodeRabbit

**`_bolt11` not cleared on failure** — fixed, and this was a real bug I introduced. If
submitting throws after generation, nothing reached the daemon, yet the readout still
named an invoice; automation would have gone looking for a payment nobody was going to
make. Cleared in the catch, *and* the error branch now precedes the readout, so a future
path that forgets to clear it still cannot render a success that did not happen.

**Relay URL canonicalization** — fixed. Real hazard, and mine: the identifier normalizes
a trailing slash but the duplicate check did not, so a list holding both spellings gave
two rows one identifier. Canonicalized before comparing and before storing, through one
`canonicalRelayUrl` the identifier helper matches. I deliberately did *not* dedupe the
list returned by `getRelays()` — if the Rust pool holds both spellings that is a
relay-pool issue, and papering over it in the card would hide it.

**Unguarded mirror write** — fixed. Worth noting this code is *moved*, not new: the bug
predates the PR in `main.dart`. But the doc comment I carried over promised the loop
survives a failure, and it did not.

**Banner text** — fixed, `· Mortsom`.

**Tests for public-key loading** — fixed by making it testable first. `_loadPublicKey`
calls the Rust bridge, so the screen itself cannot be pumped in a unit test. The card is
now `PublicKeyCard` in its own file and `test/features/account/widgets/public_key_card_test.dart`
covers no identity, a key, and a replaced key — including that the placeholder never
leaks into the readout, since an empty readout is how the harness knows to generate an
identity. The three lines of async loading in the screen remain uncovered; that needs a
bridge fake, which is a wider change than this PR should carry.

**Test for the NWC submission transition** — fixed, and it found something. The widget
had no seam (`nwc_api.makeInvoice` called straight from `initState`), so it takes its
generator as a parameter now. `test/shared/widgets/nwc_invoice_widget_test.dart` covers
the invoice being readable *while the send is still in flight* — that is the window
automation reads it in — plus both failure paths. I checked the third test fails against
the pre-fix widget, so it is a regression test and not a decorative one.

### One thing I left alone

`flutter analyze` on this branch is clean on the Flutter version CI pins (3.38.2). On a
newer local toolchain it reports `containsSemantics` as deprecated in favour of
`isSemantics` — but `isSemantics` does not exist on 3.38.2, so migrating now would break
CI. Same trap as MostroP2P/mobile#650's "build the contract test on the Flutter version
CI pins", from the other side. Worth doing in the same PR that bumps `FLUTTER_VERSION`.

`flutter test`: 229 passing.

PR #312, opened 2026-08-25
refactor(orders): move the request registry and status rules off the bridge

grunch:
First step of #120, which I rewrote before starting: the problem there was never the line count, it is that `rust/src/api/` is the **FRB bridge surface** and `api/orders.rs` had accumulated the protocol internals as well. Nothing moved in this PR is reachable from Dart.

## What moved

**`mostro::pending`** — the correlation registry that decides which inbound daemon reply belongs to which in-flight request: `DaemonReply`, `PendingRequestKind`, `PendingRequest`, the `PENDING_REQUESTS` static and its matchers (`take_matching_*`, `remove`/`purge`/`detach`), plus `classify_take_reply`, `may_reconcile_stored_id` and `order_content_key`. Self-contained bookkeeping over its own state.

**`mostro::status`** — the pure mapping and reconciliation rules: `status_for_action`, `map_core_status`, `is_hard_terminal`, `is_terminal_status`, `wire_status_applies`, `cancellation_wipes_history`, `add_invoice_sync`.

The second module was not optional scope creep: `classify_take_reply` calls `map_core_status` and `status_for_action`, so extracting the registry alone would have left a back-dependency from `mostro` into `api` — the exact inversion this is meant to remove.

`api/orders.rs`: **5014 → 3911 lines.**

## Behaviour

Preserved. Every function moved verbatim; the only edits are visibility and the imports at both ends. No call site changed.

## Tests

Tests follow their subjects — all of them. The eleven that exercise the registry and the status rules moved out of `api/orders.rs` along with the code they test, so the bridge file no longer owns their coverage:

- to `mostro::pending`: the `take_matching_restore` / `take_matching_request` / `take_matching_take` / `take_matching_add_invoice` correlation tests, `late_genuine_reply_matches_after_timeout`, `concurrent_requests_do_not_cross_consume`, `overlapping_same_key_attempts_do_not_cross_detach`, both `classify_take_reply_*`, `stored_id_reconciles_only_when_owned_by_the_pending_create`, `content_key_lookup_skips_live_waiters`
- to `mostro::status`: `the_public_status_never_replaces_a_finer_local_one`, `hard_terminal_matches_protocol_finality`, `cancellation_wipes_history_only_for_never_active_trades`, `add_invoice_sync_maps_payloads`, `escrow_locked_actions_imply_active`, `status_for_action_maps_add_invoice`

`small_order_with` is now needed by both modules, so it lives in a shared `mostro::test_fixtures` rather than being spelled out twice.

Both modules also gained coverage that was awkward to justify while they were buried in a 5000-line file. The new ones pin rules the existing tests did not:

- only this trade key's own pending local UUID may be rebound
- the content key separates range from fixed orders, and buys from sells, and stays inside the `content:` prefix that keeps it out of the UUID keyspace it shares a map with
- `SettledHoldInvoice` is terminal but **not** hard-terminal — conflating the two would drop a trade whose payout is still in flight
- the bond statuses map to `None` rather than to something wrong, and the match stays exhaustive so a new upstream variant breaks the build
- actions that carry no status change return `None` rather than a guess

## Review round 1

- the eleven tests above actually moved; the first pass had moved one and left the rest behind
- three tests written in the first pass restated rules the moved ones pin more thoroughly (the rebind rule, the wire-status precedence, the cancellation wipe) — dropped rather than duplicated
- `PENDING_REQUESTS`, `request_id_matches` and `is_terminal_status` have no caller outside their own module and are private again; widening them worked against the encapsulation this refactor exists to create. Both modules are `pub(crate) mod`, matching their contents
- rustfmt on the new files — the `pub(crate)` prefix had pushed several lines past 100 columns
- the `map_core_status` doc comment was dangling on `status_for_action` (a pre-existing bug in `main`); reattached
- a module doc saying which of `mostro::status` and `mostro::fsm` owns what
- rebased on `main`, which carried the gift-wrap → transport rename (#313) through the moved code: the debug tag is `daemon-msg` and the docs no longer say gift-wrap

## Verification

- `cargo test --lib` — 257 passed, 0 failed (260 on `main` minus the 3 duplicates dropped above)
- `cargo clippy --lib --all-targets` — no warning in any touched file
- `flutter analyze` — 7 issues, identical to `main` (all pre-existing)
- `flutter test` — 229 passed
- `./scripts/frb-generate.sh --check` — clean; the bridge surface is unchanged, which is the point

## Next

Steps 2–5 in #120, in order of value: `dispatch_mostro_message` (798 lines in one function) is the big one, then the subscription plumbing and `ingest_order_event`. Each is its own PR.


grunch:
@coderabbitai review

grunch:
@coderabbitai review

PR #315, opened 2026-08-25
docs(readme): add Linux desktop toolchain to prerequisites

Forte11Cuba:
The base Linux desktop toolchain (clang/cmake/ninja/GTK) required by Flutter was not documented.

The native-assets build drives the linker through \`clang\` and looks for it inside clang's own toolchain directory. When the default \`clang\` is a versioned LLVM install whose \`bin\` dir has no linker, the build fails with \`Failed to find any of [ld.lld, ld]\`. Adding \`lld\` covers that case.

## Summary by CodeRabbit

* **Documentation**
  * Added Linux desktop development prerequisites to the setup guide, including required build tools and libraries.


Catrya (dismissed):
ACK

Catrya (approved):
ACK

grunch (approved):
LGTM

PR #323, opened 2026-08-26
fix(order): prevent fixed price on range orders

### LINKED ISSUE #322: Range orders can be created with a fixed price, which Mostro rejects
On the Create Order screen, the range toggle (min/max fiat amount) and the price type toggle (Market / Fixed) are fully independent. Nothing prevents the user from enabling a range order and selecting a fixed sats price at the same time.

A range order has a variable amount, so it cannot carry a fixed sats price Mostro prices it at market with a premium. When both are set, createOrder sends fiatAmountMin/fiatAmountMax and amountSats together, and the daemon rejects the order with a CantDo error. The user only finds out at submit time, via a generic error, with no explanation of what went wrong.

Steps to reproduce

1. Open Create Order (buy or sell).
2. Turn on the Range order switch and enter a min and max amount.
3. In the price section, switch to Fixed and enter an amount in sats.
4. Fill payment method and submit.

Expected: The UI should not allow combining a range order with a fixed price. Range orders should be forced to market price, matching mobile v1.

Actual: The order is submitted and rejected by the daemon with a CantDo error; the combination is never prevented in the UI.

Forte11Cuba:
Closes #322

Range orders have a variable amount and must be priced at market, so combining a range with a fixed sats price made the daemon reject the order with a CantDo error at submit time.

Mirror mobile v1: enabling range now forces Market price, clears any entered fixed sats, and disables the Fixed toggle with an explanatory note. Adds the localized string across en/es/fr/de/it.

## Summary by CodeRabbit

- **New Features**
  - Range orders now automatically use Market pricing.
  - Fixed pricing is disabled while range mode is active.
  - Enabling range mode clears any entered fixed-price value.
  - Added localized guidance explaining how to use fixed pricing with range orders in English, German, Spanish, French, and Italian.

### DIFF
--- a/lib/features/order/screens/add_order_screen.dart
+++ b/lib/features/order/screens/add_order_screen.dart
@@ -51,6 +51,7 @@ class _AddOrderScreenState extends ConsumerState<AddOrderScreen> {
           ref.read(settingsProvider).defaultFiatCode ?? 'USD';
       ref.read(selectedFiatCodeProvider.notifier).state = defaultFiat;
       ref.read(isMarketPriceProvider.notifier).state = true;
+      ref.read(isRangeOrderProvider.notifier).state = false;
       ref.read(premiumValueProvider.notifier).state = 0.0;
       ref.read(fixedSatsProvider.notifier).state = '';
       ref.read(selectedOrderPresetProvider.notifier).state =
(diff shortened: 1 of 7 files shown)


Catrya (approved):
tACK

Issue #327, opened 2026-08-27
Rating screen has no reachable "rated" state, the rate prompt persists after a successful rating


Forte11Cuba:
### Root cause
`TradeStatus.rated` exists but is **unreachable**: `_mapOrderStatus` (`lib/features/trades/screens/trade_detail_screen.dart`) maps `settledHoldInvoice`/`success` → `TradeStatus.pendingRating` unconditionally, and nothing consumes the local rating state (`reputation_api.getRatingForTrade`, which already returns `is_mine=true` after `submitRating`) or the daemon's `RateReceived` (which carries the score, e.g. `RatingUser(5)`) to flip to `rated`. The bridge methods `getRatingForTrade` / `onRatingReceived` are exposed but unused in the UI.

### Reproduction
1. Complete a trade end-to-end and submit a rating for the counterpart.
2. The daemon confirms with `RateReceived` (e.g. `RatingUser(5)`), but the trade detail screen still shows the "Rate" step/prompt indefinitely.

### Proposed fix (options)
- Overlay the local rating state: when the trade is settled/success **and** a local `is_mine` rating exists, show `TradeStatus.rated` instead of `pendingRating`; invalidate the provider after `submitRating`.
- And/or drive it from the daemon's `RateReceived` confirmation.

### Notes
v1 `mobile` does not distinguish a "rated" state either (it keeps the RATE button), so this is a UX improvement rather than a protocol correctness bug. Separate from the trade-key index bug (#326 ).

Catrya:
done in #330

PR #330, opened 2026-08-27
fix(#327): make the rated trade state reachable

AndreaDiazCorreia:
`TradeStatus.rated` had a label, a headline, an instruction, a timeline step and its own CTA branch, but nothing could ever produce it: `_mapOrderStatus` buckets `settledHoldInvoice` / `success` / `completedByAdmin` / `settledByAdmin` as `pendingRating` unconditionally, so the rate prompt persisted after a successful rating. The bridge methods that hold the answer (`getRatingForTrade`) were exposed and unused — they appeared only in the generated `lib/src/rust/api/reputation.dart`, with zero call sites in feature code.

## What changed

- `lib/features/rate/providers/rating_providers.dart` (new)
  - `tradeRatingProvider` — the rating the Rust store holds for a trade.
  - `ratedByMeProvider` — whether the local user rated their counterpart.
- `trade_detail_screen.dart` — overlays that predicate on the mapped order status: `pendingRating` + rated-by-me resolves to `rated`.
- `rate_counterpart_screen.dart` — invalidates `tradeRatingProvider` after a  successful `submitRating`, so the screen underneath resolves on pop instead  of on the next rebuild.

No Rust changes, no bridge surface change, no new l10n keys — every string and widget the `rated` state needs already existed.

## Why `isMine` matters

`RatingStore::get` falls back to the counterpart's rating when the local user has not submitted one, so a non-null `RatingInfo` does not mean "I rated" — it can mean "they rated me". Only `is_mine` resolves the prompt. That is why the predicate lives in its own provider and has a dedicated test.

## Tests

Three cases in `test/features/trades/trade_detail_screen_test.dart`, driven through the real provider chain:

- settled, no rating → prompt stays
- settled, rated by me → `Rated` + `CLOSE`, prompt gone
- settled, rated by the counterpart only → prompt stays

`flutter analyze` clean; `flutter test` 246 passing.

## Summary by CodeRabbit

* **New Features**
  * Trade details now indicate when you have already rated a settled trade.
  * Rating status distinguishes between ratings you submitted and ratings received from the other party.
  * The rating prompt is replaced with a close action after you submit your rating.

* **Bug Fixes**
  * Trade details now refresh rating status immediately after a rating is submitted.

### DIFF
--- a/lib/features/rate/providers/rating_providers.dart
+++ b/lib/features/rate/providers/rating_providers.dart
@@ -3,18 +3,23 @@ import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:mostro/src/rust/api/reputation.dart' as reputation_api;
 import 'package:mostro/src/rust/api/types.dart';
 
-/// The rating the local user submitted for [tradeId], or `null` when they
-/// have not rated that trade yet.
+/// The rating held for [tradeId], or `null` when neither side has rated.
+///
+/// The Rust store is in-memory by design (ratings live in the daemon's kind
+/// 38383 tags, not in the local DB), so this resolves to `null` again after a
+/// restart and the rate prompt comes back.
+final tradeRatingProvider = FutureProvider.autoDispose
+    .family<RatingInfo?, String>((ref, tradeId) async {
+  return reputation_api.getRatingForTrade(tradeId: tradeId);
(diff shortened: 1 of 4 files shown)


Catrya (approved):
tACK

PR #331, opened 2026-08-28
fix(#259): serialize daemon message dispatch and take persistence per order id

### LINKED ISSUE #259: Serialize gift-wrap handling per order id
Split out of #197 / PR #258.

## Problem
`dispatch_mostro_message` validates the recipient trade key
(`is_current_generation`) and then mutates the order book, the persisted trade
and the session across several awaits. Validation and mutation are not one
operation, so a delivery that passes the gate can be overtaken by a retake of
the same order id while suspended:

1. `Canceled(A)` passes the gate while session A is current, then yields;
2. retake B is accepted and persists its order-book / DB state;
3. `Canceled(A)` resumes, removes B's book entry and writes `Canceled` into
   B's trade row.

The session half is already covered: #258 made every session mutation
generation-conditional, and `install_session` guarantees an accepted take keeps
its session whatever the ordering. What remains is the order-book and DB writes.

## Scope
Serialize per `order_id` (a keyed mutex around gate plus mutations), or make the
order-book and DB writes generation-conditional the way the session ones are.
This is dispatcher-wide infrastructure, not bond-specific: the `Canceled` arm
has mutated book and DB with no generation check since long before #258.

## Notes
Raised by review bots on #258. A regression test has to drive the real
`dispatch_mostro_message` (building an `UnwrappedMessage`), which no test does
today — that scaffolding is part of this work.

AndreaDiazCorreia:
`dispatch_mostro_message` checks local state (terminal-status gate, local-status lookup) and mutates the order book, the trade row and the session several `await`s later. Validation and mutation are not one operation, so a delivery that passed the gate can be overtaken by a retake of the same order id while suspended: the retake persists its state, the suspended handler resumes and writes the previous generation's outcome over it.

Both halves now run under a per-order mutex, keyed by `order_id` so one stalled handler cannot stop every other trade.

- `dispatch_mostro_message` takes the lock once the message kind is parsed, covering the local→daemon id reconcile, the waiter  interception and every per-action arm. Messages with no order id own no order state and take no lock.
- `take_order` takes it around its persistence block only — after the  daemon reply, never around the wait for it. That reply is delivered by  `dispatch_mostro_message`, which takes the same lock, so holding it across the wait would deadlock the take until its 10 s timeout. The  constraint is documented at `lock_order` and in the contract.
- The registry drops entries no handler holds any more, so it tracks live work rather than every order ever dispatched.

## Verification

The regression test drives the real `dispatch_mostro_message` with a `Canceled` while the lock is held on a retake's behalf, and asserts the two effects land in order. With the guard removed it fails with `[Canceled, Active]` instead of `[Active, Canceled]` — the previous generation's outcome written over the new one.

`cargo clippy --locked -- -D warnings`, `cargo test --lib` (284 pass) and `cargo check --target wasm32-unknown-unknown` are clean. No Dart changes: `./scripts/frb-generate.sh` leaves `lib/src/rust/` untouched, since everything added is private.

## Out of scope

This closes the interleaving window (steps 1-3 of the issue). A `Canceled` for a superseded generation that arrives genuinely late,
overlapping nothing, is still applied — catching that needs the generation check on `Session.trade_key_index` that #258 introduces on the session side. The two are complementary.

Closes #259

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved order processing reliability by preventing conflicting actions on the same order from running simultaneously.
  - Ensured cancellations and retakes are handled safely during concurrent activity.
  - Preserved parallel processing for unrelated orders.

- **Documentation**
  - Documented order-processing serialization, lock behavior, and cleanup guarantees.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -830,6 +830,16 @@ pub async fn take_order(
         peer_days: None,
     };
 
+    // The other side of the race guarded in `dispatch_mostro_message`: this
+    // block is the "retake is accepted and persists its state" step. Taking the
+    // same per-order lock keeps it from landing between a daemon handler's
+    // check and its write, and keeps that handler from landing between ours
+    // (#259).
+    //
+    // Acquired here, *after* the daemon reply above, and never around the wait
+    // itself: that reply is delivered by `dispatch_mostro_message`, which takes
+    // this very lock — holding it while waiting would deadlock the take.
+    let _order_guard = lock_order(&order_id).await;
     store_trade_key_index(&order_id, trade_index).await;
     if status.is_some() || amount_sats.is_some() {
         // Keep the public order book in sync with the reply so the order
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -4660,6 +4660,69 @@ mod tests {
         assert!(map.contains_key(&live));
     }
 
+    /// The #259 race, driven through the real dispatcher: a `Canceled` for a
+    /// generation that is being replaced must not land in the middle of the
+    /// retake persisting its own state.
+    ///
+    /// The retake side is represented by the lock `take_order` holds around its
+    /// persistence block, because `take_order` itself needs a relay pool and a
+    /// live daemon. The dispatcher is the code under test and runs unmodified,
+    /// against a real `UnwrappedMessage`.
+    ///
+    /// The assertion is on the *order* of the two effects rather than on a
+    /// timeout: unserialized, the dispatcher reaches `emit_trade_update` during
+    /// the sleep below and its Canceled is observed before the retake's write.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn a_cancel_cannot_land_inside_a_concurrent_retake() {
+        use crate::api::types::OrderStatus;
+        use crate::rt::time::{sleep, Duration};
+        use mostro_core::message::{Action, Message};
+
+        let order_uuid = uuid::Uuid::new_v4();
+        let order_id = order_uuid.to_string();
+        // Pending, so the terminal-status gate lets the Canceled through and
+        // the arm runs its full sequence.
+        order_book().upsert_order(dummy_order_info(&order_id)).await;
+
+        let mut rx = trade_updates_tx().subscribe();
+
+        let sender =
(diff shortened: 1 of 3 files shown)


Catrya (approved):
## Approved

Concurrency changes are where subtle bugs hide, so I went looking for the usual four and all four are covered:

- **The registry prune is safe.** `retain(|_, lock| Arc::strong_count(lock) > 1)` cannot drop a lock in use: `OwnedMutexGuard` holds its own `Arc`, so a held entry always counts ≥ 2. The dangerous interleaving — removing an entry while another task is about to acquire it, leaving two tasks on different mutexes for the same key — can't happen.
- **The guard hand-off is correct.** The dispatcher does `order_guard.take()` into the `Wake` and returns immediately after, so it never continues working without the lock. The FIFO rationale in the comment is right: releasing it there would let an already-queued message beat the woken task and run its arm against a trade row and session that don't exist yet.
- **The deadlock this design invites is identified and avoided** — `take_order` takes the lock around its persistence block only, never around the wait, since the reply is delivered by the dispatcher under the same lock.
- **No network I/O under the lock.** `on_peer_pubkey_received` derives keys and spawns its subscription in `crate::rt::spawn`, outside the critical section.

The tests cover both failure modes I'd have gone hunting for by hand: `a_dead_take_waiter_releases_the_handed_lock` (the guard travelling to a receiver that's already gone) and `the_registry_drops_locks_no_handler_holds`. I did not re-run the regression test with the guard removed to confirm the `[Canceled, Active]` ordering it reports.

## One correction to the description, before merging

The **Out of scope** section says a late `Canceled` for a superseded generation "is still applied" and that catching it needs the `Session.trade_key_index` check from #258. Neither half is true any more:

- The PR **does** drop those messages. The generation gate is in the dispatcher, read under the same lock, keyed on the trade-key binding: `if trade_index < bound { return; }` — with `a_late_cancel_for_a_superseded_generation_is_dropped` and `a_message_for_a_newer_generation_passes_the_gate` covering it. That gate isn't described anywhere in the body, so a substantive part of the change has its rationale only in code comments.
- #258 is closed. The session-side work that remains — a confirmed retake keeping the stale session and its old `trade_key_index` — is tracked in **#335**, and is complementary to this at the session layer rather than the dispatch layer.

Worth fixing because the PR page is what someone reads months from now to decide whether that case is covered, and because whoever picks up #335 could otherwise re-implement this gate a layer down.

PR #333, opened 2026-08-28
feat(#328): source restore trade-key resync from Action::LastTradeIndex

### LINKED ISSUE #328: Implement Action::LastTradeIndex and source the restore trade-key resync from it
Carries the remaining scope of #217, closed as partially completed by PR #239. Sub-issue of #142.

PR #239 delivered the resync machinery — `ensure_trade_key_index_at_least`: a monotonic, idempotent bump that persists with an in-memory rollback if the save fails. That part is correct and is reused here unchanged. What it got wrong is **where the floor comes from**, and the daemon already exposes the right number.

## The problem #217 set out to fix is still open

#217's stated problem was: after a restore, the next order the user creates reuses a trade key already bound to a recovered trade. #239 raises the counter to the maximum index across the trades the restore returns, which satisfies #217's acceptance criteria literally but does not achieve that goal.

Per `restore_session.md` in `MostroP2P/protocol`, the daemon returns **only non-finalized orders**:

> "Mostro will respond with a message containing all non-finalized orders (e.g., statuses such as `pending`, `active`, `fiat-sent`, `waiting-buyer-invoice`, `waiting-payment`, `settled-hold-invoice`) and any active disputes."

So the maximum over that payload is a **lower bound** of the daemon's counter. When the user's highest-index trade is already finished, the floor comes out short and the next legitimate increment lands on an index the daemon has already seen.

## Reproduced against a live regtest mostrod
Setup: order **X** open at index 1, order **Y** *canceled* at index 2 — the finalized trade holding the higher index. Then a fresh install, counter 0, going through the app's own code path:

```
restore_session()      -> ACCEPTED, 1 order
  order=d7ea25d4…  index=1  status=pending     <- only the non-finalized one
counter after resync   -> 1                     <- PR #239
first order created    -> REJECTED: InvalidTradeIndex
```

The daemon's own payload:

```json
"restore_data":{"orders":[{"order_id":"d7ea25d4…","trade_index":1,"status":"pending"}],"disputes":[]}
```
and its refusal:

```json
{"cant-do":{...,"payload":{"cant_do":"invalid_trade_index"}}}
```

mostrod refuses an index it has already seen for that identity **even though that order was canceled**. The user restores successfully and their very first new order fails. It only takes the most recent trade being the one that finished first — not an edge case.

## The fix: the protocol already answers this

`mostro-core` has `Action::LastTradeIndex`:

> Client asks Mostro for its last known trade index. Payload must be `None`.

The value comes back in the reply's `trade_index` field. Asked against the same identity as above, it returns **2** — the real high-water mark, including the canceled trade — where the restore payload reported 1.

`mostro-cli` implements it in `src/cli/last_trade_index.rs` and, on the reply, performs exactly the resync #217 wanted (`user.set_last_trade_index(...)`). **This app does not implement it anywhere** — `LastTradeIndex` appears nowhere in `rust/src/`.

Two properties that matter:

- **Account-scoped, signed with the identity key for both seal and rumor.** From mostro-cli: *"Mostro looks that up by the sender pubkey. Sign both seal and rumor with `identity_keys` so the request resolves to the account, not to a (possibly unregistered) trade key."*
- **It consumes no trade index**, which also closes the second window below.

## Scope

1. Send `Message::Restore(MessageKind::new(None, None, None, Action::LastTradeIndex, None))`, wrapped with the identity keys as both seal and rumor signer.
2. Handle the inbound `Action::LastTradeIndex` reply and read the counter from its `trade_index` field.
3. Feed it into `ensure_trade_key_index_at_least(floor)` from PR #239 — unchanged.
4. Keep `recovered_max_trade_index` (also from #239) as a **fallback** for when the daemon does not answer, not as the primary source.

## Second, minor window — closed by the same change

`restore_session` calls `derive_trade_key()` as its first statement, purely to obtain a reply address: `Message::new_restore(None)` carries no `trade_index`, and the probe confirmed the daemon accepts a restore whose sender trade key is already bound to an existing trade. So it burns a local counter tick rather than breaking the restore, and no ephemeral sender key is needed — PR #239's scope note describes this window as more severe than it is. Asking `LastTradeIndex` first, identity-signed and deriving nothing, removes even that tick.

## Acceptance criteria
- [ ] `Action::LastTradeIndex` is sent (identity-signed,
- [ ] `Action::LastTradeIndex` is sent (identity-signed, no trade key derived) and its reply handled
- [ ] The resync floor is sourced from that reply, with the restore payload maximum kept only as a fallback when the daemon does not answer
- [ ] After a restore on a fresh install, the first order created is accepted — **including when the user's highest-index trade is already finalized**
- [ ] The counter is never lowered, and re-asking is idempotent
- [ ] Covered on the regtest e2e path, alongside `trade_then_restore_recovers_order`

## Related

- #217 (closed) / PR #239 — the machinery this reuses; its floor source is what this replaces
- #218 — applies the restore snapshot and leans on this counter; its step 4 points at #217 and should point here
- #142, #215, #216, #219 — the restore flow this blocks
- Not #326, which is the create-time daemon-UUID → index **mapping**: a separate bug in the same file

Forte11Cuba:
Close #328 

The restore payload lists only non-finalized orders, so its maximum trade index is a lower bound of the daemon's real counter. When the user's highest-index trade is already finalized, the resync floor came out short and the first order created after a restore was rejected with InvalidTradeIndex.

 Implement Action::LastTradeIndex: an account-scoped request whose reply carries the daemon's authoritative counter. The key split follows the protocol spec (key_management.html): the rumor is authored by the trade key restore_session already derived  so the outer kind-14 never publishes the master identity pubkey while the account resolves from the encrypted identity proof (event.identity) and the reply comes back to the trade key (event.sender), which is exactly what the daemon implements. This deliberately diverges from mostro-cli, which departs from the spec by signing both seal and rumor with the identity keys identity-authored events would leak a permanent identity→Mostro link on every relay. The privacy property is pinned by last_trade_index_payload_none_rumor_by_trade_key, mutation-verified.

Validated end-to-end against a live daemon: with an open order at index 1 and a canceled order at index 2, LastTradeIndex returns 2 while the restore payload maximum is 1.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale or replayed responses from corrupting order state during recovery.
  * Preserved trade-key associations during fingerprint recovery.
  * Improved trade-session restoration by selecting the highest valid trade index.
  * Ensured the first order after restoration is accepted when the daemon has advanced its trade counter.

* **Reliability**
  * Improved recovery when live trade-counter data is unavailable, delayed, or refused.
  * Trade-counter synchronization now safely handles duplicate and unmatched responses.
  * Recovery remains safe and repeatable without lowering an existing trade counter.


Forte11Cuba:
@coderabbitai review

Catrya (changes requested):
Reviewed at `17d3a82`, merged locally against current `main`. On the merged result: **283 Rust tests** pass, `cargo clippy --locked -- -D warnings` is clean, and `cargo check --locked --target wasm32-unknown-unknown` passes.

## Verified against a live daemon: it does what #328 asked

I re-ran the scenario from the issue — order X open at index 1, order Y **canceled** at index 2, so the finalized trade holds the higher index — then a fresh install:

```
counter before          = 0
restore_session()       -> Ok, 1 order        (the payload still carries only X)
counter after restore   = Some(2)             (it was 1 before this PR)
first order created     -> ACCEPTED, index 4
```

The acceptance criterion holds: after a restore, the first new order is accepted even though the highest index belongs to a finalized trade. On a second run the daemon's counter had advanced and the resync followed it correctly.

## Blocking: the subscription is never closed

`last_trade_index()` calls `client.subscribe(filter, None)` with an auto-generated id and never unsubscribes — not on the success `return`, not on the timeout, not on the `Shutdown`/`Closed` breaks. Every restore leaves a live relay-side subscription (`author=mostro`, `#p=identity_pk`) for the rest of the process. That's the leak class #182 describes and #255 is currently fixing in the other two loops.

Both reference clients solve this, in opposite ways, and each is consistent with the shape of what it subscribes to:

- **mostro-cli** — `wait_for_dm`, which is exactly what its `execute_last_trade_index` uses — subscribes with `SubscribeAutoCloseOptions::default().exit_policy(ReqExitPolicy::WaitForEventsAfterEOSE(1))` and never unsubscribes; there is not a single `unsubscribe` call in that codebase.
- **mostrix** has no auto-close at all: 41 explicit `unsubscribe` calls across five files, plus an id registry and teardown on reload/reconnect — all of them for long-lived listeners.

`last_trade_index()` is mostro-cli's shape: one request, one reply, discarded after ten seconds. So the fix is the library's own auto-close rather than manual bookkeeping:

```rust
let opts = SubscribeAutoCloseOptions::default()
    .exit_policy(ReqExitPolicy::WaitForEventsAfterEOSE(1))
    .timeout(Some(Duration::from_secs(10)));   // also close when nothing arrives
client.subscribe(filter, Some(opts)).await
```

The relay-side CLOSE is then the library's job on every path, including the one where the daemon never answers, and the existing outer wait loop stays as it is.

Worth noting the distinction so this doesn't look like it contradicts the direction of #255: auto-closing subscriptions are deliberately excluded from re-subscription on reconnect (`should_resubscribe` returns `false` for them), which is why they are wrong for a 30-minute per-trade watcher and right here — if the socket drops mid-request, the request fails and the code falls back to the payload maximum, which is the designed behaviour.

## Minor

**The worst-case restore latency doubles.** `restore_session` already waits 10s for its own reply; this chains another 10s for `LastTradeIndex`. Against a silent daemon a restore now takes 20s before falling back. Since this query — unlike the restore itself — *has* a fallback, a shorter timeout (5s) costs little and halves the worst case.

**One improvement that isn't visible in the diff, worth a line in the description:** the request is made unconditionally, so a user whose trades are *all* finalized — the case where the payload contributes nothing — now resyncs correctly. Before this PR that restore did nothing at all.

## Also verified

- **The reply is authenticated properly**: it checks `event.pubkey == mostro_pubkey`, that the p-tag is ours, and — the one that matters — `unwrapped.sender != mostro_pubkey` after decryption, so a third party can't inject a forged counter.
- **`rx = client.notifications()` is taken before subscribing**, so the reply can't land in the gap between subscribe and the first `recv`. `limit(0)` carries the same anti-replay rationale as `subscribe_daemon_messages`.
- **No trade key is derived**, which was half the point of #328: it also closes the minor window where the restore itself consumed an index.
- **Privacy mode**: `get_active_keys()` is the master key, which is correct here because recovery only exists in reputation mode (`import_from_mnemonic` states it), not because the toggle is ignored.
- **`resync_floor` takes the max of both sources** rather than trusting the daemon blindly, and its comment explains both why that is a no-op against a consistent daemon and why it is kept anyway. Good call.
- **`sanitize_trade_index`** correctly drops negatives and `u32::MAX` (the reserved terminal index that would overflow the next `+1`), and is now shared by both the payload path and the reply path so the two can't drift.

## One observation, not attributed to this PR

On the **first** run of the scenario, the post-restore `create_order` returned `NoDaemonResponse` — but the daemon had in fact created the order (`New order saved Id: b51c39c8…`, with its reply sent). So the client reported failure for an order that exists. It did not reproduce on the second run, so I can't tie it to this change; it points at `create_order`'s 10s correlation window over a public relay. Flagging it here only so it isn't lost — it's the mirror image of the phantom-order problem and deserves its own issue if it shows up again.

rust/src/api/orders.rs
```diff
@@ -3604,6 +3611,196 @@ fn recovered_max_trade_index(
...
+/// This is a self-contained request/reply (own subscription + inline wait, like
+/// mostro-cli's `wait_for_dm`) rather than a reuse of the per-trade
+/// subscription/dispatch path: that path derives its recipient keys from a
+/// trade index and is coupled to trade keys, but this reply is encrypted to the
+/// identity key.
+async fn last_trade_index() -> Result<Option<u32>> {
+    use nostr_sdk::RelayPoolNotification;
+    use crate::rt::time::{timeout, Duration};
+
+    let identity_keys = crate::api::identity::get_active_keys().await?;
```
grunch: **Major — publishes an event authored by the master identity key, and bypasses the privacy toggle.**

`wrap_message_nip44` signs the outer kind-14 with its `trade_keys` argument, so passing `identity_keys` for both here makes `event.pubkey` the user's long-lived identity pubkey, `p`-tagged to Mostro, on every relay — permanently linking the master key to Mostro. Every other daemon-bound event in this client is authored by an ephemeral trade key.

It isn't required: the daemon resolves the account from `event.identity` (`mostro/src/app/last_trade_index.rs`: `let requester_pubkey = event.identity.to_string();`) and only *replies* to `event.sender`. That's the same shape `restore_session` already uses correctly with a fresh trade key + identity proof.

Suggestion: take the `sender_keys` `restore_session()` already derived as the rumor key and `get_transport_identity_keys(&sender_keys)` as the identity key, and subscribe/filter on that trade pubkey. It's already in the global DM coverage, which also kills the `no-matching-p-tag` warn this currently triggers.

Also: `get_active_keys()` skips `get_transport_identity_keys()`, this codebase's single gate for privacy mode (`identity.rs:640-645`). Nothing stops `restore_session()` from running with privacy mode on, so today this signs and publishes with the real identity key even then.

rust/src/api/orders.rs
```diff
@@ -3604,6 +3611,196 @@ fn recovered_max_trade_index(
...
+    // (mostro-cli's wait_for_dm shape), not a long-lived watcher, so the library
+    // issues the CLOSE on every path: after the reply (WaitForEventsAfterEOSE(1))
+    // and, if the daemon never answers, on the timeout. Leaving it to manual
+    // bookkeeping is the leak class #182/#255 address. Auto-close subs are
+    // deliberately excluded from reconnect re-subscription (correct here: if the
+    // socket drops mid-request we fall back to the payload maximum by design).
+    let close_opts = nostr_sdk::prelude::SubscribeAutoCloseOptions::default()
+        .exit_policy(nostr_sdk::prelude::ReqExitPolicy::WaitForEventsAfterEOSE(1))
+        .timeout(Some(REPLY_TIMEOUT));
+    if let Err(e) = client.subscribe(filter, Some(close_opts)).await {
```
grunch: **Major — the relay-side auto-close timer starts here, but the client-side deadline starts only after the PoW and the publish.**

Between this `subscribe` and `let start = Instant::now()` (l. 3741) sit `first_contact_pow_for()` (awaits the capability watch, up to `CAPABILITY_WAIT` = 10s) and the synchronous PoW mining inside `EventBuilder::pow()`, plus the publish round-trip. So the effective window is `5s − (pow + publish)`, and the two budgets are skewed by that amount — the `REPLY_TIMEOUT` comment ("so the two can't drift") and the wait-loop comment ("both give up together") describe a property this ordering doesn't have.

At a high node-advertised `pow_first_contact` on a slow device, the relay can CLOSE before the request is even published: the reply is then never deliverable and the feature silently degrades to the payload max — the exact #328 bug — with only the routine fallback warn in the log.

Suggested ordering: build the event (paying the PoW) → `client.notifications()` → `subscribe` → `publish` → start the deadline at subscribe time. The receiver-before-subscribe guarantee is preserved.

rust/src/mostro/actions.rs
```diff
@@ -419,6 +419,32 @@ pub async fn restore_session(
     wrap_message_first_contact(identity_keys, trade_keys, mostro_pubkey, &msg).await
 }
 
+/// Build and wrap a `LastTradeIndex` request (#328).
+///
+/// Account-scoped: `identity_keys` sign BOTH the Seal and the rumor, so the
+/// daemon resolves the account by sender pubkey and NO trade key is derived.
```
grunch: **Doc inaccuracy** (mirrored from `mostro-cli`, but it's wrong there too): the daemon does **not** resolve the account by the sender pubkey. `mostro/src/app/last_trade_index.rs` reads

```rust
let requester_pubkey = event.identity.to_string(); // account lookup
let trade_key = event.sender;                      // reply address only
```

`event.identity` is the proven identity pubkey from the encrypted identity proof (or the rumor author when no proof is attached, i.e. privacy mode). Signing the rumor with the identity key is therefore a *choice*, not a protocol requirement — see the main review comment on `orders.rs:3679`.

rust/src/api/orders.rs
```diff
@@ -3604,6 +3611,196 @@ fn recovered_max_trade_index(
...
+                            continue;
+                        }
+                        let idx = kind.trade_index.and_then(sanitize_trade_index);
+                        crate::api::logging::blog_info("restore", format!(
+                            "LastTradeIndex reply: trade_index={:?} -> floor={idx:?}",
+                            kind.trade_index
+                        ));
+                        return Ok(idx);
+                    }
+                    Ok(None) => continue,
```
grunch, line 4219: **Minor — `CantDo` replies fall through here and cost the full 5s.**

The daemon answers this action with `MostroCantDo(CantDoReason::NotFound)` (user absent) or `MostroCantDo(CantDoReason::InvalidTradeIndex)` (`last_trade_index == 0`), echoing the same `request_id`. `is_matching_last_trade_index_reply` rejects them on the action check, so the loop `continue`s and burns the whole timeout before falling back.

An early `return Ok(None)` on `Action::CantDo` with a matching `request_id` (with a `blog_warn` naming the reason) turns a 5s stall into an immediate, diagnosable fallback.

rust/src/api/orders.rs
```diff
@@ -3604,6 +3611,196 @@ fn recovered_max_trade_index(
...
+/// when neither source has a usable index.
+fn resync_floor(
+    daemon_counter: Option<u32>,
+    info: &mostro_core::message::RestoreSessionInfo,
+) -> Option<u32> {
+    let payload_max = recovered_max_trade_index(info);
+    match (daemon_counter, payload_max) {
+        (Some(daemon), Some(payload)) => Some(daemon.max(payload)),
+        (daemon, payload) => daemon.or(payload),
+    }
```
grunch, line 4017: **Minor — this silently swallows the case the doc above calls impossible.**

If `payload > daemon`, the doc's own argument says the daemon is stale, partial, or buggy. `recovered_max_trade_index` already warns when it drops an index for the same class of reason ("the daemon sent something this client's model does not cover"). A `blog_warn` on that branch would make a real inconsistency visible rather than invisible-but-handled — the behaviour is right, the silence isn't.

rust/src/api/orders.rs
```diff
@@ -3604,6 +3611,196 @@ fn recovered_max_trade_index(
...
+}
+
+/// True only for the reply to THIS `LastTradeIndex` request: the action
+/// matches and the daemon echoed our correlation nonce
+/// (`mostro/src/app/last_trade_index.rs` copies `request_id` into the reply).
+///
+/// A replayed reply from an earlier request carries a different nonce — or
+/// none, since this client's own pre-hardening requests sent no id — so
```
grunch: **Nit — factual accuracy of the comment.** No released build of this client has ever sent a `LastTradeIndex` request; the id-less version exists only in an earlier commit of this PR (`17d3a82`), so there are no "stored replies" from it in the wild. Worth rewording so a false history doesn't get frozen into a comment — strict matching is justified without it (and `limit(0)` already makes historical replays undeliverable on this subscription).

grunch (changes requested):
**Review — `feat(#328): source restore trade-key resync from Action::LastTradeIndex`**

The diagnosis is right and the fix is the right shape: the restore payload only lists non-finalized orders, so its max index is a lower bound, and `Action::LastTradeIndex` is the authoritative counter. `resync_floor` taking the max of both sources is correct and well argued, the `sanitize_trade_index` extraction removes a real duplication risk, and the new unit tests cover the interesting cases. `cargo clippy --all-targets` is clean and `cargo test --lib` passes locally (45/45 in `api::orders`).

Two issues need addressing before merge, plus some smaller items.

---

### 🔴 Major

**1. The request is published as a kind-14 event authored by the master identity key — a public, permanent link between the user's long-lived pubkey and Mostro. And the daemon does not need it.**

`mostro-core::transport::wrap_message_nip44` signs the *outer* event with the `trade_keys` argument:

```rust
EventBuilder::new(Kind::PrivateDirectMessage, encrypted)
    .tags(tags)
    .pow(opts.pow)
    .sign_with_keys(trade_keys)   // <- outer event author
```

`actions::last_trade_index` passes `identity_keys` for both, so `event.pubkey` on the wire **is the master identity pubkey**, `p`-tagged to the Mostro node, and the daemon's reply is `p`-tagged back to it. Every relay now sees `identity_pubkey → mostro_pubkey`, permanently. This is the first place in this client that publishes a relay event authored by the identity key — every other daemon-bound event is authored by an ephemeral trade key, with the identity key carried only inside the *encrypted* identity proof.

It is also unnecessary. The daemon resolves the account from `event.identity` (the identity proof), not from the author:

```rust
// mostro/src/app/last_trade_index.rs
let requester_pubkey = event.identity.to_string();   // account lookup
let trade_key = event.sender;                        // reply address
...
send_dm(trade_key, my_keys, &message_json, None).await
```

That is byte-for-byte the same shape as `restore_session.rs` (`let master_key = event.identity;`), which this client already speaks correctly using a fresh trade key + identity proof. A trade-key-authored rumor resolves to the same account and the reply comes back to the trade key.

Suggested fix: have `last_trade_index()` take the trade keys `restore_session()` already derived (`sender_keys`) as the rumor key and `get_transport_identity_keys(&sender_keys)` as the identity key, then subscribe/filter on the trade pubkey instead of the identity pubkey. That trade key is already in the global DM coverage (`ensure_global_dm_coverage` runs a few lines earlier), which also resolves finding **3** below.

Secondary point on the same line: `get_active_keys()` bypasses `get_transport_identity_keys()`, which is this codebase's single gate for the privacy toggle (`api/identity.rs:640-645`). Nothing in `restore_session()` refuses to run in privacy mode, so with privacy mode ON this call still signs with — and publishes — the real identity key. Whatever is decided about the transport shape, the identity key should be resolved through that helper, as `api/reputation.rs:162` explicitly does.

**2. The relay-side auto-close timer and the client-side deadline do not "give up together" — the relay's starts first.**

Current ordering in `last_trade_index()`:

1. `client.subscribe(filter, close_opts.timeout(5s))` — **relay-side 5s starts here**
2. `actions::last_trade_index(...)` → `wrap_message_first_contact` → `first_contact_pow_for()` (awaits the capability watch, up to `CAPABILITY_WAIT` = 10s) **and mines the PoW synchronously in `EventBuilder::pow()`**
3. `publish_event_json(...)`
4. `let start = Instant::now()` — **client-side 5s starts here**

So the usable window is `5s − (pow + publish)`, not 5s, and the two budgets are skewed by exactly that amount. The comment on `REPLY_TIMEOUT` ("Shared by the relay-side auto-close and the outer wait loop so the two can't drift") and on the wait loop ("the same budget the relay-side auto-close uses, so both give up together") describe a property the code does not have.

The failure mode is silent: at a high node-advertised `pow_first_contact` on a slow device, the relay CLOSEs the subscription before or shortly after the request is published, the reply is never deliverable, and the feature degrades to the payload max — i.e. exactly the #328 bug — with nothing in the logs but the routine "no usable daemon reply" fallback warn.

Cheap fix: build the event first (paying the PoW cost), *then* take `client.notifications()`, subscribe, publish, and start the deadline at subscribe time. The receiver-before-subscribe ordering that the existing comment protects is preserved.

---

### 🟡 Minor

**3. The reply trips the global handler's "anomaly" warning on every restore.** `build_trade_key_map()` only covers trade indexes `1..=trade_key_index`, so the identity pubkey is never in the map. The global notification loop is not filtered by subscription id, so it sees this reply, finds no matching `p` tag, and emits `blog_warn("daemon-msg", "drop ev=… reason=no-matching-p-tag …")` — a warning whose comment says it exists to flag "stale filter after regenerate? key map gap?". Making it fire on a normal happy path erodes it. Resolved for free by finding **1**.

**4. `CantDo` replies are ignored, costing a full 5s stall on every failure path.** The daemon answers this action with `MostroCantDo(CantDoReason::NotFound)` (user absent) or `MostroCantDo(CantDoReason::InvalidTradeIndex)` (`last_trade_index == 0`), and those replies carry the same echoed `request_id`. The loop `continue`s past them and burns the whole timeout before falling back. Early-returning `Ok(None)` on `Action::CantDo` with a matching `request_id` turns a 5s stall into an immediate, well-logged fallback.

**5. `resync_floor` silently swallows the case its own doc calls impossible.** The doc argues a consistent daemon can never answer below an index it still returns in the restore payload, so `payload_max > daemon_counter` means a stale/partial reply or a daemon bug. `recovered_max_trade_index` already warns when it drops an index for exactly that "the daemon sent something this client's model does not cover" reason; the same `blog_warn` here would make a real inconsistency visible instead of invisible-but-handled.

---

### 🔵 Nits

**6.** The `is_matching_last_trade_index_reply` doc says "this client's own pre-hardening requests sent no id — so their stored replies are exactly the replay material to reject". No released build of this client has ever sent a `LastTradeIndex` request; the id-less version exists only in an earlier commit of this same PR. Worth rewording so a false history doesn't get codified in a comment (the argument for strict matching stands on its own).

**7. Spec not updated.** `specs/004-mostro-p2p-client/contracts/identity.md` still describes the recovery flow as steps 1–6 with a bare "5. Sync trade key index", with no mention of the new `LastTradeIndex` round-trip or the fallback. Per `CLAUDE.md` ("Specs are a living artifact — update the matching spec/contract as part of any behavior/contract change") this should move with the code.

---

### ✅ Verified

- `Message::Restore(MessageKind::new(None, Some(request_id), None, Action::LastTradeIndex, None))` matches what the daemon builds and what `MessageKind::verify` requires (`LastTradeIndex | RestoreSession => self.payload.is_none()`). ✔
- The daemon copies `request_id` into the reply and embeds the counter in `trade_index`, so the correlation guard and `kind.trade_index.and_then(sanitize_trade_index)` are correct. ✔
- Floor semantics line up on both sides: `user.last_trade_index` and `recovered_max_trade_index` are both "last index used", and `ensure_trade_key_index_at_least` raises monotonically to that. ✔
- `crate::rt::time` shims keep the new `timeout`/`Instant` usage wasm-safe. ✔

Requesting changes on **1** and **2**.

Catrya (changes requested):
**Changes requested.** The production code is right — I checked the non-obvious claims against the daemon source instead of taking them on trust, and I ran the end-to-end test against a live daemon including the counterfactual. What has to change is the evidence and the record: the e2e test the PR presents as its proof cannot run as shipped, and the description describes a design this PR already abandoned.

## Blocking: the e2e test cannot run as shipped

I ran `restore_with_finalized_top_index_resyncs_from_last_trade_index` against a live daemon and relay. First attempt, unmodified:

```
[test] creating order A (index 1)...
panicked: create order A: StorageUnavailable: deriving a trade key requires durable storage
```

**The test never calls `init_db`**, and `derive_trade_key` refuses without durable storage, so it dies on the first order before reaching any assertion. Its sibling `restore_session_roundtrip` does not call it either, so nothing initialises the `OnceCell` by another route. One line fixes it:

```rust
let dbp = std::env::temp_dir().join(format!("e2e-333-{}.db", uuid::Uuid::new_v4()));
crate::db::app_db::init_db(dbp.to_str().unwrap()).await.expect("init_db");
```

With that in, it passes:

```
[test] creating order A (index 1)...  order A id=bf8d543d…
[test] creating order B (index 2)...  order B id=6a6f86ac…
[test] calling restore_session()...
[test]   order_id=bf8d543d… status=pending index=1     ← payload carries only A
[test] ✓ post-restore order accepted id=665e187e…
test result: ok. 1 passed
```

And the counterfactual — cutting the `last_trade_index` call out of `restore_session`, i.e. the pre-#333 state — fails on exactly the right assertion:

```
panicked: counter (1) must be >= 2 — the LastTradeIndex floor, not the payload bound
```

So the test does prove what the PR claims; it just needs that line, plus the node pubkey lifted out of the hardcoded `bae71ea2…` (the author's regtest) into an env var so anyone can run it. This is blocking because — see the last section — that test is the *only* possible evidence for this change: there is no user-reachable path to exercise it.

## Blocking: the description describes a design this PR abandoned

It says the request is *"signed with the identity keys for both the Seal and the rumor (deriving no trade key)"*. Commit `88e2616` ("review round 1 — trade-key-authored request") changed that, and the current code signs the rumor with a trade key — which is this PR's insight and its deliberate divergence from mostro-cli. As written, the body describes precisely the variant that leaks the identity, and the body is what the next reader will believe.

## The design claim is correct, and the divergence from mostro-cli is the right call

The load-bearing claim is *who signs the rumor*. `mostro/src/app/last_trade_index.rs` says exactly what the docstring says:

```rust
let requester_pubkey = event.identity.to_string();   // account resolution
let trade_key = event.sender;                        // reply address only
let user = is_user_present(pool, requester_pubkey).await
...
send_dm(trade_key, my_keys, &message_json, None)
```

So authoring the rumor with an ephemeral trade key resolves the account correctly **and** avoids publishing a permanent identity↔Mostro link as a kind-14 author on every relay. And the divergence the docstring flags is real: `mostro-cli/src/cli/last_trade_index.rs:32-33` signs seal and rumor with `identity_keys`, with a comment claiming the daemon resolves by sender pubkey. The daemon source says otherwise.

The rest checks out too:

- **Reply shape** — `Message::Restore(MessageKind::new(None, request_id, Some(last_trade_index), Action::LastTradeIndex, None))`: the counter travels in `trade_index`, the nonce is echoed, payload `None`.
- **Refusals** — `CantDo(NotFound)` for an unknown account and `CantDo(InvalidTradeIndex)` for a zero counter are both `MostroCantDo`, so they *do* produce a reply (unlike internal errors), and `manage_errors` copies the `request_id`. The CantDo fast path is right.
- **No version boundary** — `last_trade_index.rs` has existed since **v0.15.0** (Oct 2025), well before the v0.18.0 that introduced transport v2. Any daemon this client can talk to already has it.
- **`check_trade_index` skips this action** (it only covers `NewOrder`/`TakeBuy`/`TakeSell`), so sending `trade_index: None` is correct.
- **Privacy mode** degrades exactly as the docstring says: no identity proof, no account to look up, `CantDo(NotFound)`, fall back to the payload bound.

Also worth recording: the `WaitDurationAfterEOSE` vs `WaitForEventsAfterEOSE(1)` reasoning is subtle and correct — the restore reply goes to the same trade key and would consume a one-event budget.

Catrya (approved):
tACK

PR #336, opened 2026-08-28
docs(readme): list supported NIPs and BUDs

### LINKED ISSUE #123: README: explicitly list supported NIPs and BUDs
**Type:** docs / dx
**Priority:** low

The README mentions `nostr-sdk 0.44` and NIP-44 / NIP-59 in passing. A short bullet list of NIPs (and Blossom BUDs) the app actually implements would help contributors and integrators understand the protocol surface immediately.

**Proposed list (verify before merging)**
- NIP-44 (encrypted payloads — Mostro transport v2)
- NIP-47 (Nostr Wallet Connect)
- NIP-59 (Gift Wrap — peer / admin chat)
- NIP-17 (DM transport semantics)
- NIP-40 (event expiration)
- BUD-01 / BUD-02 (Blossom upload + Kind-24242 auth)

Place under "The Mostro Protocol" or "Architecture & Fundamentals".

Matobi98:
## Summary

Adds a consolidated "Supported NIPs & BUDs" section to the README, under **Protocol Reference**. The README mentioned NIP-44/NIP-59 in passing but had no single place listing what the app actually implements.

Each entry was verified against the current implementation, not just against code comments:

- **NIP-06** — mnemonic to key derivation (`crypto/keys.rs`, real BIP-32 DerivationPath)
- **NIP-13** — proof-of-work mining on outgoing events (`mostro/pow.rs`, real EventBuilder pow call)
- **NIP-44** — encrypted payloads for daemon messages and the chat envelope (`crypto/ecdh.rs`, `nostr/transport.rs`, real nip44 encrypt/decrypt calls)
- **NIP-47** — Nostr Wallet Connect (`nwc/client.rs`, real WalletConnectResponse handling)
- **NIP-69** — public order-book status via the `s` tag on Kind 38383
- **BUD-01 / BUD-02** — encrypted file attachments to Blossom servers (`nostr/blossom.rs`, real reqwest PUT upload)

Also corrects an outdated claim: NIP-59 (gift wrap) is no longer used by peer or dispute chat (dropped in #246/#254, this client neither reads nor writes Kind 1059). Chat instead rides a custom Kind-14 envelope, which is explicitly not NIP-17 despite sharing its kind number, noted to avoid a common point of confusion.

NIP-17 and NIP-40 were considered but left out of the list: NIP-17 is not actually implemented (kind 14 is reu
sed, but not per the NIP-17 spec), and I could not confirm real NIP-40 tag usage in the codebase (only an internal expiration field), happy to add it back if someone can confirm.

Docs-only change, no code touched.

Closes #123

## Test plan

- Rendered the README locally to confirm the section displays correctly
- Each NIP/BUD claim cross-checked against actual code (not just comments), see file references above
- N/A, no Rust/Dart code changed, so cargo fmt/clippy/test and flutter analyze/test do not apply

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,9 +154,11 @@ Both parties rate each other (optional)
 | [NIP-44](https://github.com/nostr-protocol/nips/blob/master/44.md) | Encrypted payloads — daemon messages and the peer/dispute chat envelope (`rust/src/crypto/ecdh.rs`, `rust/src/nostr/transport.rs`) |
 | [NIP-47](https://github.com/nostr-protocol/nips/blob/master/47.md) | Nostr Wallet Connect, for paying invoices from a connected wallet (`rust/src/nwc/client.rs`) |
 | [NIP-69](https://github.com/nostr-protocol/nips/blob/master/69.md) | Public order-book status (`s` tag on Kind 38383) |
-| [BUD-01 / BUD-02](https://github.com/hzrd149/blossom) | Encrypted file attachment upload/auth to Blossom servers (`rust/src/nostr/blossom.rs`) |
+| [BUD-01](https://github.com/hzrd149/blossom/blob/master/buds/01.md) | Blob retrieval (`GET /<sha256>`) for encrypted file attachments (`rust/src/nostr/blossom.rs`) |
 
 **Not used:** NIP-59 (gift wrap) was dropped from peer and dispute chat in #246/#254 — this client neither reads nor writes Kind 1059. Chat instead rides a custom Kind-14 envelope (see "P2P Chat" above), which is *not* NIP-17 despite sharing its kind number.
+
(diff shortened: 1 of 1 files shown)


Matobi98:
@ermeme Both are valid, fixed the first and tracking the second separately:

1. Fixed — the Architecture diagram's `nostr-sdk 0.44 → relay pool, NIP-44/59` bullet was stale, same claim as the one this PR already corrected in the Protocol section. Updated to match.

2. The upload path doesn't match BUD-02 as written — verified against the current spec (https://github.com/hzrd149/blossom/blob/master/buds/02.md) and its commit history: `PUT /upload` has been the defined endpoint since May 2024, well before this code was written in #91, so it's not a case of the spec moving after the fact. The code's own comment even misattributes it to BUD-01, which defines no PUT endpoint at all.

Since this PR is docs-only, I didn't want to fold a Rust behavior change in here per CONTRIBUTING's one-PR-one-change guidance, so I corrected the README to describe what the code actually does today (no BUD-02 upload compliance claim) and opened #341 to track fixing `try_upload` to use `PUT /upload` per spec. Encryption is confirmed application-layer (ChaCha20-Poly1305 before upload, not part of either BUD) — added a note on that too.

One more thing while verifying #2: the Kind-24242 authorization event itself isn't BUD-02 either — that's [BUD-11](https://github.com/hzrd149/blossom/blob/master/buds/11.md) (Nostr Authorization), a separate spec. The code comment mislabels this too. Noted in the README and added as a follow-up on #341.

Catrya (approved):
ACK

PR #111, opened 2026-06-22
feat(transport): switch Mostro protocol traffic to NIP-44 direct (kind 14)

Catrya:
Switches this app's Mostro-protocol transport from NIP-59 gift wrap (kind 1059) to protocol v2 — NIP-44 direct, signed kind-14 events authored by the trade key. PR number 2 of 2 for the transport migration (spec `005-transport-v2-migration`); follows the `mostro-core` 0.13.1 bump (#110).

Protocol v2 only (no dual support): the target node speaks v2 and there are no users to bridge. Peer/dispute chat is out of scope and stays on gift wrap.

  ### Changes

  **1. Wire format (`gift_wrap.rs`)**
  - `wrap_mostro_message` → `wrap_message_with(Transport::Nip44Direct, …)` (`WrapOptions` unchanged: PoW preserved, `expiration: None` — daemon fills its own, FR-005)
  - `unwrap_mostro_message` → `unwrap_incoming` (dispatches by event kind)
  - Tests assert kind 14 + author = trade key; v2 doc refresh

  **2. Wiring (`relay_pool.rs`, `orders.rs`)**
  - Three subscription filters → kind 14 + `.author(mostro_pubkey)` (relay_pool `dm_filter`, orders per-trade + global bulk)
  - Three receive handlers → kind-14 check + reject `event.pubkey != mostro_pubkey`, disambiguating Mostro replies from NIP-17 peer chat (also kind 14, FR-003)
  - Remove now-unused `KIND_GIFT_WRAP` const; refresh stale docs/logs/comments

  #### Out of scope (unchanged)

  - NIP-17 peer chat and dispute-admin chat stay on gift wrap kind 1059 (`gift_wrap.rs` `wrap`/`unwrap`, `messages.rs::subscribe_incoming_chat`) — FR-004
  - Mostro message logic, action set, payloads, key derivation
  - Internal identifier names (`subscribe_gift_wraps`, etc.) kept to keep the diff focused

  ### Verification

  - `cargo build` / `cargo test` (81 passed) / `cargo clippy` — all green, no new warnings
  - Dart untouched; no FRB regen (changes internal to `crate::nostr` / `crate::api`)
  - Manual E2E against a v2 node still pending — full order lifecycle (new-order → take → pay → fiat-sent → release)

  ### Spec

  `specs/005-transport-v2-migration/` · reference:

## Release Notes

**Refactor**
[...]

PR #340, opened 2026-08-28
docs(specs): sync and reorganize tasks.md for 004 and 005

### LINKED ISSUE #122: docs: sync and reorganize tasks.md for specs 004 and 005
**Type:** docs
**Priority:** low

Two small docs chores in one issue:

**`specs/005-transport-v2-migration/tasks.md`**
- All 12 tasks are still `[ ]` even though the code was merged in #111 and verified end-to-end on a v2 node.
- Update to `[x]` (or add a "completed" note) so the doc reflects reality.

**`specs/004-mostro-p2p-client/tasks.md`**
- Phase 18/19/20 are completed historical work mixed with the active backlog.
- Separate "completed phases" from "active backlog" so contributors can tell what's still open at a glance.
- Add a short note at the top: "Mostro transport migrated to NIP-44 direct (Kind 14) in spec 005 — earlier tasks describing NIP-59 wrap for Mostro messages reflect the pre-v2 design."
- **Do not** delete the historical tasks themselves — they remain valuable as record.

No behavior change.

Matobi98:
## Summary

Two docs chores from #122, both covering `specs/*/tasks.md` files that had drifted from reality.

### specs/005-transport-v2-migration/tasks.md

All 12 tasks were still `[ ]` even though the transport migration merged in #111 and was verified end-to-end. Marked all 12 done, and added a status note at the top: the plan is now a historical record, not a map of the current code. Verified against current code, it's been superseded further than the issue assumed — mostro-core is now 0.14.1 (past the 0.13.1 target), gift_wrap.rs no longer exists (wrap_mostro_message/unwrap_mostro_message now live in rust/src/nostr/transport.rs), and T011's "leave gift wrap on peer/dispute chat untouched" no longer holds, since that chat also moved off gift wrap entirely in #246/#254.

### specs/004-mostro-p2p-client/tasks.md

Three changes:

- Added an "Open work at a glance" summary near the top. Almost the entire file (T001-T145) is done, so phases 18/19/20 aren't really historical work mixed into an active backlog, they're just appended sequentially at the end. The real open work is two small clusters of partial tasks: file attachments (T079, T080, T081) and dispute admin chat (T088, T090-T093), both cases where the Rust side is ready but nothing in Dart calls it yet.
- Corrected an existing transport-v2 note that had gone stale in the same way as the 005 file: it claimed peer chat still uses NIP-59 gift wrap, which is no longer true after #246/#254.
- Added a pointer task (T148, Phase 21) for the mnemonic-restore flow: it has a contract in contracts/identity.md (import_from_mnemonic with recover=true) but no task ever tracked it, per @grunch's note on #142. Scoped as a pointer to the epic (#142 and its sub-issues), not a new task breakdown, so it doesn't duplicate work already planned there.

Docs-only change, no code touched.

Closes #122

## Test plan

- Read both files fully before editing to confirm the actual state (checked boxes, phase structure, existing notes) rather than assuming the issue's description was still accurate
- Cross-checked the corrected claims against the current code (mostro-core version, file layout, chat transport) the same way as in #123
- N/A, no Rust/Dart code changed, so cargo fmt/clippy/test and flutter analyze/test do not apply

### DIFF
--- a/specs/004-mostro-p2p-client/tasks.md
+++ b/specs/004-mostro-p2p-client/tasks.md
@@ -19,6 +19,26 @@
 - `[~]` — Partial: code exists but blocked on missing infrastructure (noted in description)
 - `[ ]` — Not started
 
+## Open work at a glance
+
+Most of what follows (T001–T145) is `[x]` — the phases read as a
+completed build log, not a live backlog. What's actually still open lives
+in three clusters of `[~]` partial tasks (code exists, but the backend or
+the UI wiring doesn't):
+
+- **IndexedDB / web storage** (Phase 1 — Setup): T010 — only messages,
+  settings and the active Mostro pubkey are implemented; orders, trades,
+  relays, identity, queued messages and trade keys return "IndexedDB not
+  yet implemented", so the web target cannot persist a trade (#233).
+- **File attachments** (Phase 10 — P2P Chat): T079, T080, T081 — Blossom
+  upload/download (T070) has no Dart caller yet.
+- **Dispute admin chat** (Phase 12 — Dispute System): T088, T090, T091,
+  T092, T093 — the Rust side (T085/T086) is ready; the screens are stubs
+  with no bridge subscription.
+
+Plus one task with no entry at all: **restore from mnemonic** — see T148
+below.
+
 ---
 
 ## Phase 1: Setup (Project Initialization)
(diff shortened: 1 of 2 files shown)


Catrya (commented):
@Matobi98 please fix the conflicts and address the hermemes review, thanks!

Matobi98:
Thanks for the review — both blockers addressed in the rebased head.

1. Conflicts: rebased onto main (179 commits). The only clash was the Transport v2 note, which upstream had already rewritten in 61fa304 — I dropped my edit of that line and kept upstream's.

2. T010: now [~] with the real scope spelled out — only messages, settings and the active Mostro pubkey are backed by real object stores; orders, trades, relays, identity, queued messages and trade keys still return IndexedDB not yet implemented (#233). Added as the first bullet in "Open work at a glance" (phase order), and fixed the surrounding claims: "Almost everything below (T001–T145) is [x]" → "Most of what follows…", and "two clusters" → "three clusters".

3. Superseded tasks: added _(Superseded — see the Transport v2 note above.)_ inline on 004's T011 and T012, the two that name Kind 1059 explicitly; the rest of the historical set stays covered by the block note. 005's T011 already carried that annotation.

PR #343, opened 2026-08-29
feat(#337): validate market-price orders against the node's sats limits

### LINKED ISSUE #337: Validate market-price orders against the node's sats limits (needs a fiat↔sats rate)
Follow-up to #282 / PR #302, which validated **fixed-sats** orders against the node's `min_order_amount` / `max_order_amount`. Market-price orders are still unchecked: the user submits, and the daemon rejects with `OutOfRangeSatsAmount` after the fact.

The blocker is that checking a market-price order needs a fiat↔sats conversion, and **this app has no exchange rate at all** — there is no rate provider, service or price fetch anywhere in `lib/`.

## How v1 does it

v1 implements this fully, and its decisions are worth copying because most of them are about rounding and wording rather than the check itself.

**Rate source** (`lib/services/nostr_exchange_service.dart`): the rate comes from the **Mostro node over Nostr**, a NIP-33 addressable **kind 30078** event, with an in-memory cache and a fallback chain — Nostr → Yadio HTTP → cache — surfacing `"Failed to fetch exchange rates from all sources (Nostr, HTTP, cache)"` when all three fail. `exchangeServiceProvider` wires `NostrExchangeService`, not the HTTP one, so the normal path adds no external dependency and leaks nothing about which currency the user is interested in.

**It blocks rather than failing open** (`add_order_screen.dart`, `_validateSatsRange`): no rate → `exchangeRateNotAvailable`; no node info → `mostroInstanceNotAvailable`. Note this is the opposite of PR #302's fail-open choice for fixed sats, and deliberately so: without a rate there is nothing to compare, so submitting is a coin flip.

**Conversion and rounding** (`lib/shared/utils/order_amount_limits.dart`):

```dart
int _calculateSatsFromFiat(double fiat, double rate) => (fiat / rate * 100000000).round();

final minFiat = max(1, (minSats / _satsPerBtc * exchangeRate).ceil());
final maxFiat = (maxSats / _satsPerBtc * exchangeRate).floor();
```

The min rounds **up** and the max rounds **down**, so every value inside the range shown to the user is guaranteed to be inside the real sats range. The min is floored at 1 because the field rejects 0 and decimals.

**It shows the limits in fiat, not in sats**, with the reason written into the code: *"the sats bounds mean nothing to most users"*. It falls back to raw sats when the fiat range collapses — `isDisplayable => minFiat >= 1 && maxFiat >= minFiat`, i.e. when the whole valid range is under one unit of fiat and no whole number is enterable. That yields four messages: too low / too high × fiat-range / sats-only.

## Scope

1. **A rate source.** Per Principle I the kind-30078 fetch is protocol work and belongs in Rust, alongside `fetch_mostro_instance_tags` — v1 does it in Dart only because v1 is all-Dart. Dart consumes the resulting rate.
2. **The conversion and check**, mirroring v1's rounding directions so a shown bound is never itself invalid.
3. **The messages**, in fiat with the sats fallback, localized in the five locales.
4. Decide whether to keep v1's block-on-missing-rate or PR #302's fail-open. They can legitimately differ: a fixed-sats amount is directly comparable, a market-price one is not.

## Acceptance criteria

- [ ] A market-price order whose fiat amount converts outside the node's sats range is rejected before submitting, with the range shown in the user's fiat currency
- [ ] Every whole fiat value inside the shown range is accepted by the daemon (rounding never advertises an invalid bound)
- [ ] The sats fallback message appears when the fiat range collapses below one unit
- [ ] Behaviour when no rate is available is explicit and documented, not incidental

## Related

- #282 / PR #302 — the fixed-sats half, and where this was split off
- While verifying #302 it also turned out that node capabilities are read **twice** from the same kind-38385 event: Rust fetches it for pow / `protocol_version` / escrow mode, and Dart fetches it again for min/max (and for `bond_enabled`, noted in #258). If a rate fetch lands in Rust, that is the moment to unify: expose the node's capabilities from the core and let Dart read them from there.

AndreaDiazCorreia:
## Problem
PR #302 checked fixed-sats amounts against the node's `min_order_amount` / `max_order_amount`. Market-price orders stayed unchecked: they carry no sats amount, so nothing on this side could tell whether the daemon would price them inside the range — the user submitted and got `OutOfRangeSatsAmount` after the fact. The blocker was that the check needs a fiat↔sats rate, and this app had no rate source at all.

## Fix
Read the rate the Mostro node itself publishes (kind 30078, NIP-33, `d` tag `mostro-rates`) and use it to check the fiat amount before submitting.

**Rust — the rate source** (`mostro/rates.rs`, `api::nostr::fetch_exchange_rate`):
- Same NIP-33 query shape as `fetch_mostro_instance_tags`, re-verifying kind, author and d-tag on whatever the relay returns.
- Discards an event that has expired per its NIP-40 `expiration` tag, so a relay that ignores NIP-40 cannot serve a zombie price.
- The rate table is cached per node — never served back to a different one, like `pow` and `escrow_mode` — bounded by the event's own expiration and clamped to one hour, so a range order's amount fields cost a single relay query.
- Per Principle I this is protocol work and belongs in Rust; v1 does it in Dart only because v1 is all-Dart.

**Dart — the check** (`shared/utils/order_amount_limits.dart`, `add_order_screen.dart`):
- `marketAmountsOutOfNodeRange(...)` takes every amount the daemon will price — both ends for a range order — disables submit, and shows an inline message naming the accepted range, with the usual defence-in-depth re-check in `_submit`.
- The range is shown in the user's own currency, since the sats bounds mean nothing to most people, falling back to sats when the whole valid range is under one fiat unit and no whole number is enterable:
  > Amount must be between {min} and {max} {currency} for this Mostro node
- **Fails open**: no rate, no advertised bounds, or an amount that is not yet a number blocks nothing, and the daemon stays the authority. This is the opposite of v1's choice, deliberately — publishing rates is optional for an operator (`publish_to_nostr`), so blocking would make market-price orders unusable against every node that leaves it off. Written into `contracts/orders.md`.

## Decisions taken from the daemon, not from v1
Verified against `MostroP2P/mostro` while implementing; v1 differs on the first point:
- **Truncation, not rounding.** The daemon computes `(fiat_amount / price * 1E8) as i64` (`src/app/order.rs`). v1 rounds, which can accept an amount one sat below the node's floor and still see it rejected — the exact surprise this check exists to remove.
- **The premium is not part of the check.** The daemon prices the raw fiat amount.
- **Both ends of a range order are checked**, because the daemon prices every amount in `amount_vec` and rejects the order if any one of them is out of range.
- The `min_order_amount` tag is the very value the check compares against (`src/nip33.rs` publishes `mostro_settings.min_payment_amount`), and the published rate comes from the same aggregate as `get_bitcoin_price` — so this reproduces the node's own arithmetic rather than approximating it.

## Scope
- No Yadio/HTTP fallback. The Nostr path adds no external dependency and does not disclose which currency the user is interested in; a fallback can be its own issue.
- The kind-38385 double read noted in the issue (Rust for pow / `protocol_version` / escrow, Dart for min/max) is left alone — that is a refactor, and deserves a separate issue.
- The preview footer still shows a sats figure in Fixed price mode only. Extending it to market price is now possible, but out of scope here.

## Testing
Automated only so far — **not yet verified against a live node**:
- `cargo test`: 304 passed (14 new in `mostro::rates`); `cargo clippy --all-targets` clean on the touched files.
- `flutter test`: 281 passed (24 new); `flutter analyze` clean.
- The acceptance criterion "every whole fiat value in the shown range is accepted by the daemon" is covered by a test sweeping bounds and midpoints across four limit/rate combinations, including a sub-1 rate and a 1,000,000 rate, against the same conversion the daemon performs.

Closes #337.

## Summary by CodeRabbit

* **New Features**
  * Added exchange-rate retrieval for supported fiat currencies.
  * Market-price orders now validate entered amounts against node limits using current exchange rates.
  * Out-of-range warnings display the applicable fiat range when available.
  * Added localized validation messages in English, German, Spanish, French, and Italian.

* **Bug Fixes**
  * Improved handling of missing, expired, invalid, or unavailable exchange-rate data without blocking order submission.

### DIFF
--- a/lib/features/order/providers/exchange_rate_provider.dart
+++ b/lib/features/order/providers/exchange_rate_provider.dart
@@ -1,0 +1,41 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+import 'package:mostro/core/mostro_defaults.dart';
+import 'package:mostro/features/settings/widgets/mostro_node_selector.dart';
+import 'package:mostro/src/rust/api/nostr.dart' as nostr_api;
+
+/// Price of one BTC in [fiatCode], as published by the active Mostro node in
+/// its Kind 30078 (`d` = `mostro-rates`) event.
+///
+/// Exists so a market-price order can be checked against the node's sats
+/// limits before it is submitted (#337): the daemon prices such an order from
+/// this same rate, so it is the number its range check will use.
+///
+/// Reads the node pubkey from [mostroPubkeyProvider], like
+/// `mostroNodeProvider`, so the rate always belongs to the node the order will
+/// be sent to.
+///
+/// Resolves to `null` whenever the node has no usable rate to give — it
+/// publishes none (publishing is optional), the event has expired, or it
+/// quotes no such currency — and an unreachable relay surfaces as an error.
+/// Callers must treat both as "not checkable" and submit anyway, leaving the
+/// daemon as the authority, which is what PR #302 chose for fixed-sats
+/// amounts.
+///
+/// `autoDispose` and keyed by currency: switching currency starts a fetch for
+/// the new one, which the Rust-side cache usually answers without another
+/// relay query.
+final exchangeRateProvider =
+    FutureProvider.autoDispose.family<double?, String>((ref, fiatCode) async {
+  final code = fiatCode.trim();
+  if (code.isEmpty) return null;
+
+  final pubkey = ref.watch(mostroPubkeyProvider);
+  final resolvedPubkey =
+      pubkey.trim().isEmpty ? defaultMostroPubkey : pubkey.trim();
+
+  return nostr_api.fetchExchangeRate(
+    mostroPubkeyHex: resolvedPubkey,
+    fiatCode: code,
+  );
+});
--- a/lib/features/order/screens/add_order_screen.dart
+++ b/lib/features/order/screens/add_order_screen.dart
@@ -60,6 +60,19 @@ class AddOrderScreen extends ConsumerStatefulWidget {
   return null;
 }
 
+/// The amount [text] holds, or null when it is not one the form can submit.
+///
+/// `Infinity`, `-Infinity` and `NaN` all parse as doubles and would pass a
+/// bare positivity check, only to throw in the sats conversion further down —
+/// while the screen is building. Nothing filters the amount fields' input, so
+/// a pasted value can be any of them.
+@visibleForTesting
+double? enteredAmount(String text) {
+  final value = double.tryParse(text.trim());
+  if (value == null || !value.isFinite || value <= 0) return null;
+  return value;
+}
+
 /// Returns the node's accepted `(min, max)` sats range, and that range in
 /// fiat, when a market-price order's amount prices outside it, otherwise null.
 ///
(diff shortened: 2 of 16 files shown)


Catrya (approved):
tACK

PR #345, opened 2026-08-31
fix: persist counterparty pubkey and create the maker's session at peer reveal

### LINKED ISSUE #334: Peer chat: the counterparty pubkey is never persisted, so the maker has no chat room and nobody can send after a restart
## Root cause

When the daemon reveals the counterparty's trade pubkey, the app writes it **only into the in-memory session** (`orders.rs:2169`, inside `on_peer_pubkey_received`) — and only if a session already exists. It never reaches the trade row, which is the only durable store.

Two facts make that fatal:

1. **`create_session` is called in exactly one place**: `take_order` (`orders.rs:812`). A maker therefore never has a session. When `on_peer_pubkey_received` fires for them, it takes the `else` branch, whose own comment states the intent it doesn't carry out:

   ```rust
   } else {
       // Session may not exist … Create it now best-effort.
       log::warn!("… session not found for order={order_id}, skipping session update …");
   }
   ```

2. **The maker's trade row is created with an empty counterparty** (`orders.rs:539`, `counterparty_pubkey: String::new()`) and no code path ever fills it. The taker's row gets it at take time (`orders.rs:768`).

So for a maker the peer's identity is held nowhere at all, and for a taker it is held only in RAM.

## What breaks

1. **The maker's trade never appears in the chat rooms list.** `tradeInfoToChatRoom` bails on an empty peer (`lib/features/chat/providers/chat_providers.dart:190`: `if (peerPubkey.isEmpty) return null;`). The room is still reachable from the trade detail (`trade_detail_screen.dart:1230`), where sending then fails as below.
2. **The maker can never send.** With no session, `send_message` takes its local-only branch: it stores the message and returns `Ok` **without publishing**. The user sees the message in the chat; the counterparty never receives it. No error, no retry, no "not delivered" marker.
3. **After a restart the maker stops receiving too**, because `resubscribe_active_chats` skips trades whose `counterparty_pubkey` is empty.
4. **After a restart nobody can send** — taker included — because nothing rebuilds sessions at startup.

`send_file` and `download_attachment` don't degrade silently; they fail with `SessionNotFound`.

None of this affects the trade itself: `send_invoice`, `send_fiat_sent`, `release_order` and `cancel_order` all read the index from the persisted `trade_keys` table and re-derive their keys, so orders, statuses and actions survive a restart normally. Only the peer channel is affected, which is why this went unnoticed.

## Evidence

Verified against a live regtest daemon with a real maker and taker in separate processes. Maker creates an order, a second identity takes it:

```
[taker] TOOK ok status=WaitingBuyerInvoice session=true

[maker] t=30s  session=NONE  book_status=Some(Pending)
[maker] t=40s  session=NONE  book_status=Some(WaitingBuyerInvoice)   <- taken
[maker] t=100s session=NONE  book_status=Some(WaitingBuyerInvoice)
[maker] send_message sender_pubkey=""    <- empty == stored locally, never published
```

The maker's order book tracked the status change in real time, so this is not connectivity: daemon messages arrive fine (they don't use sessions), the session simply never exists.

**Scope of the evidence.** The send above happens at `waiting-buyer-invoice`, where the chat is not expected to work anyway, so treat it as showing the mechanism rather than the user-facing failure. Reaching `active` needs the full Lightning flow. The code path is unambiguous either way — nothing creates a session for the maker at any status, and nothing writes `counterparty_pubkey` to the trade row — so `active` changes none of the four consequences, but that step is unverified.

## How the reference clients avoid it

- **v1 (`mobile`)** creates a session on **both** sides: `newSession` is called when taking a buy, when taking a sell, and when *creating* an order (`add_order_notifier.dart:117`, keyed by `requestId` until the daemon returns the order id, then promoted). Its `Session` holds the actual key pairs, so it persists them and reconstructs sessions on restart.
- **mostrix** has no session concept: the chat state lives on the persisted order row — `counterparty_pubkey`, `order_chat_shared_key_hex` — and is derived on demand when a field is missing. Symmetric for both roles, restart-safe by construction.

## Proposed fix

Keep this app's model — sessions in memory, keys re-derived from the mnemonic, no secret material on disk — and close the gap the model requires:

1. **Persist `counterparty_pubkey` on the trade row** when the daemon reveals it, for both roles. It is a public key, so this does not change the security posture (the local SQLite is unencrypted and today holds no key material).
2. **Have the three session-dependent functions derive from the trade row** when no session exists, reusing what `resubscribe_active_chats` already does: `get_active_trade_keys(trade.trade_key_index)` + `trade.counterparty_pubkey` → `derive_chat_keys`. The session then becomes a pure cache, which is what the rest of the codebase already assumes.

Everything else in the core already rehydrates this way — `ORDER_BOOK` and `RATING_STORE` from Kind 38383, `GLOBAL_DM_KEYS` by re-deriving `1..=max_index`, `TRADE_KEY_MAP` by falling back to `db.get_trade_key`, chat listeners from the `trades` table, messages from the `messages` table. Sessions are the exception, not the rule.

## Related

- **#256** fixes the identical problem on the dispute channel — `dispute.admin_pubkey` also arrives once and is kept only in memory — by persisting it to the settings KV under `dispute_admin:<order_id>`. Same bug shape, different store, for a legitimate reason: a dispute has no durable row of its own, a trade does. Worth noting so the two don't look like divergent patterns.
- The dispute chat itself does **not** depend on the session: it derives from `dispute.admin_pubkey` + the persisted trade key index. Only the peer chat does.
- Cleanup while in here: `SessionManager::set_admin_shared_key` (`session.rs:147`) and the `Session::admin_shared_key` field are dead code — nothing in the repo calls them.
- **#197 / PR #258** discuss *when to delete* sessions. This issue is about the fact that half of them are never created.
- **#308** — this is a concrete instance of the class that issue sets out to prevent
  ("two writers, one reader"; "wiring instead of a principle"), except it happens
  without any background/resume cycle: a maker never has a session even in the
  foreground. Whatever hydration principle #308 lands should cover this.
- **#140** — wiring the attachments UI will hit the same root cause from the other
  side: `send_file` and `download_attachment` fail outright with `SessionNotFound`.

## Acceptance criteria

- [ ] A maker's trade appears in the chat rooms list once the counterparty is known
- [ ] Both roles can send peer messages in every state where the chat is available
- [ ] Both roles can still send after closing and reopening the app
- [ ] A maker still receives peer messages after a restart
- [ ] `send_message` never reports success for a message it did not publish — either it publishes, or it surfaces the failure

Forte11Cuba:
Close #334 

## Problem
- `on_peer_pubkey_received` wrote the revealed peer only into the in-memory session, and only if one existed. `create_session` has a single call site (`take_order`), so a **maker never had one**: no chat room, sends stored locally as "sent" but never published, nothing to resubscribe after a restart.
- The taker's persisted `counterparty_pubkey` was seeded with `order.creator_pubkey` — which on a Kind 38383 book order is **the Mostro node's own pubkey** (`order_events.rs`), not the maker's trade key. After a restart, `resubscribe_active_chats` derived chat keys from it and claimed the single-owner subscription guard with garbage, silently blocking the correct subscription when a replayed reveal arrived.

## Fix (mirrors mostrix, the reference client)
- **Symmetric peer capture**: any daemon message whose payload carries a `SmallOrder` naming both trade pubkeys reveals the counterparty —match our own trade key against the two, take the other. No per-action role table; every replayed reveal self-heals a row that missed it. Runs before the take-waiter interception so a take's first (consumed) reply also counts; `BondSlashed` is exempt for the same reason it skips the generation gate.
- **Durable record**: new `update_trade_counterparty` persists the peer on the trade row (`json_set` by `order.id`, refuses empty). `take_order` no longer seeds `creator_pubkey`.
- **Maker session**: `on_peer_pubkey_received` now creates the session when none exists the maker's normal case, and on web the only chat-identity store (trades table stubbed, #233).
- **Invariant guard**: `chat_still_relevant` refuses a counterparty equal to the active mostro pubkey, so a pre-fix row can never claim the subscription guard with garbage keys.

## Tests
- `resolve_peer_side_is_symmetric` — both roles + stranger payload
- `update_trade_counterparty_round_trips_by_order_id` — scoping, poisoned-row overwrite, empty refused, survives reopen
- `chat_still_relevant_selects_only_live_trades` — extended with the daemon-pubkey rejection

Manually verified against a live daemon: maker's chat room appears at reveal, messages flow both directions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trade recovery by correctly identifying and restoring peer connections.
  * Prevented invalid chat subscriptions for legacy trades involving the active daemon or trade creator.
  * Preserved counterparty details across trade updates and database restarts.
  * Prevented valid counterparty information from being replaced with empty values.
  * Improved handling of incomplete or missing peer-session information during trade setup.
  * Improved reliability of order-book and daemon-message updates, including duplicate and delayed messages.
  * Improved dispute handling when responses arrive late or maintenance mode is active.
  * Added safeguards for stale trades and missing trade information.


rust/src/api/messages.rs
```diff
@@ -1436,10 +1436,18 @@ pub(crate) async fn resubscribe_active_chats() {
...
+/// "counterparty" is the Mostro node itself (rows written before the fix
+/// seeded `creator_pubkey` = the 38383 event author) would derive garbage
+/// chat keys AND claim the single-owner subscription guard with them,
+/// silently blocking the correct subscription when a replayed reveal
+/// arrives.
 fn chat_still_relevant(trade: &crate::api::types::TradeInfo) -> bool {
     use crate::api::types::OrderStatus::*;
     trade.outcome.is_none()
         && !trade.counterparty_pubkey.is_empty()
+        && trade.counterparty_pubkey != crate::config::active_mostro_pubkey()
```
grunch, line 1455: **Moderate — this guard is node-scoped, but the rows it guards are not.**

`trades` has no node column, and `resubscribe_active_chats` iterates `db.list_trades()` — every row, from every node the user has pointed at (switching nodes clears the order book, not trades). So a pre-fix row poisoned with a *different* node's pubkey passes this check while the current node is active, derives garbage chat keys, and claims `ACTIVE_CHATS`. Since `subscribe_incoming_chat` has no idle timeout, that claim is held for the process lifetime and the self-healing reveal's spawn becomes a silent no-op — exactly the failure this guard was added to prevent.

Node-agnostic and exact, using the same row:

```suggestion
        && trade.counterparty_pubkey != trade.order.creator_pubkey
```

The poison *is* `creator_pubkey` (`take_order`'s old seed), so this catches it whichever node published the 38383 event, needs no config read, and can never reject a legitimate peer — a peer's trade pubkey is never the event author. Keep the `active_mostro_pubkey()` check alongside it if you like, but this is the load-bearing half.

rust/src/api/orders.rs
```diff
@@ -2254,28 +2266,130 @@ pub(crate) async fn local_trade_status(order_id: &str) -> Option<OrderStatus> {
...
+    let peer_hex = peer_pk.to_hex();
+    log::info!(
+        "[orders] peer-reveal {action:?}: order={order_id} role={role:?} peer={peer_hex}"
+    );
+    if let Some(db) = crate::db::app_db::db() {
+        if let Err(e) = db.update_trade_counterparty(order_id, &peer_hex).await {
+            log::warn!("[orders] peer-reveal: failed to persist counterparty: {e}");
+        }
+    }
+    on_peer_pubkey_received(order_id, &peer_hex, trade_index, Some(role)).await;
```
grunch: **Minor — derives the trade keys a second time.**

`trade_keys` is already in hand from l. 2314; `on_peer_pubkey_received` immediately re-runs `get_active_trade_keys(trade_index)` (l. 2373) to produce the same value before delegating to `apply_peer_reveal`. Since this is now the only production caller, going straight to

```rust
apply_peer_reveal(order_id, &peer_hex, &trade_keys, trade_index, Some(role)).await;
```

drops a BIP-32 derivation and an identity-lock read from every reveal, and lets `role` stop being an `Option` (it is `None` only in `peer_pubkey_with_no_session_does_not_panic`, which could pass a real role or call `apply_peer_reveal` directly).

rust/src/api/orders.rs
```diff
@@ -2254,28 +2266,130 @@ pub(crate) async fn local_trade_status(order_id: &str) -> Option<OrderStatus> {
...
-async fn on_peer_pubkey_received(order_id: &str, peer_pubkey_hex: &str) {
-    // Resolve trade key index from order_id.
-    let trade_index = match get_trade_key_index(order_id).await {
-        Some(idx) => idx,
-        None => {
-            log::warn!("[orders] on_peer_pubkey_received: no trade key for order={order_id}");
+/// `trade_index` is the index of the key that decrypted the message — ours by
+/// construction, and generation-gated by the caller, so it is more reliable
+/// than a book lookup (a take's first reply arrives before the binding).
+async fn maybe_capture_peer_reveal(
```
grunch, line 2486: **Minor — worth an early return once the peer is already captured, and worth a test of its own.**

Widening capture from two actions to "any payload naming both trade pubkeys" means the full path — key derivation (twice, see the comment below), a DB `UPDATE`, ECDH, HKDF, a session update and a task spawn — now runs for essentially every daemon message of a trade, and again for the whole replayed history on each restart, since the global DM filter deliberately carries no `since`. A short-circuit when the session already holds this exact peer *and* a shared key keeps the self-heal property for rows that missed the reveal while making the common case free (and stops the IndexedDB stub's per-message warn on web).

Separately: this function is where the behaviour actually changed, and it is the one piece with no test. The identity dependency is what makes it awkward — the same split that produced `apply_peer_reveal` works here too: lift the payload → `(buyer_hex, seller_hex)` extraction into a pure helper and table-test it (single-sided payload, non-`Order` payload, unparseable hex). That covers the new decision logic without touching process-global identity state.

rust/src/api/orders.rs
```diff
@@ -901,6 +908,21 @@ pub async fn take_order(
             trade.order.clone(),
         )
         .await;
+    // In that same case the reveal ran before the trade row existed, so its
+    // durable write was a no-op — replay it from the session now that the
+    // row is persisted (#334).
+    if let Some(session) = crate::mostro::session::session_manager()
```
grunch, line 1073: **Minor — the returned `TradeInfo` now contradicts the row this block just fixed.**

The peer goes to the DB but not to `trade`, so `Ok(trade)` hands the bridge a `counterparty_pubkey: ""` for a trade whose persisted row has the real peer. Today nothing notices (`take_order_screen.dart` discards the return and re-reads via `rawTradesProvider`), but `TradeInfo.counterpartyPubkey` is precisely what `tradeInfoToChatRoom` gates the chat room on — a caller that trusts the return value silently gets no room. Setting it on the struct alongside the DB write keeps the two in agreement for one line.

rust/src/db/indexeddb.rs
```diff
@@ -314,6 +314,17 @@ impl Storage for IndexedDbStorage {
...
     }
+
+    async fn update_trade_counterparty(
+        &self,
+        _order_id: &str,
+        _counterparty_pubkey: &str,
+    ) -> Result<()> {
+        // No trades store on web yet (#233): chat identity survives only via
+        // the in-memory session created at reveal time.
+        log::warn!("update_trade_counterparty: IndexedDB backend not implemented — counterparty will not persist");
```
grunch: **Minor — `warn` is too loud for this one.** Unlike the `mark_trade_rated` stub (once per trade), `update_trade_counterparty` is now called for every qualifying daemon message and again for every replayed one, so a web session's log fills with a line that carries no information after the first. `log::debug!` — or warning once per order — keeps the signal without the spam. The short-circuit suggested on `maybe_capture_peer_reveal` would also largely solve this.

rust/src/api/orders.rs
```diff
@@ -1774,36 +1814,8 @@ async fn dispatch_mostro_message(
...
-                }
-            };
-            log::info!(
-                "[orders] daemon-msg {:?}: order={order_id} peer={peer_pubkey_hex}",
-                kind.action
-            );
-            // Derive the ECDH shared key and store in session so the chat API
-            // can encrypt/decrypt P2P messages and subscribe to the right p-tag.
-            on_peer_pubkey_received(&order_id, &peer_pubkey_hex).await;
+            // Peer capture happens in `maybe_capture_peer_reveal` before the
```
grunch, line 2017: **Nit — the comment above this one is now stale.** The terminal guard a few lines up still explains itself as stopping a replay from "re-derive the peer key, recreate session state, or respawn the chat subscription" — none of which this arm does any more. Behaviour is correct (`maybe_capture_peer_reveal` carries its own `status_sync_blocked_by_terminal`), but the guard is now evaluated twice per message for these two actions and the comment describes the pre-#334 shape.

Also: `let small_order = ... o.clone()` is now read only for `.status`, so the clone can go.

grunch (changes requested):
**Review — `fix: persist counterparty pubkey and create the maker's session at peer reveal`**

Both halves of the diagnosis check out against the code: `take_order` really did seed `counterparty_pubkey` with the 38383 event author (the Mostro node), and `create_session` really is called from `take_order` only, so a maker reached `on_peer_pubkey_received` with no session and the old code's `if let Some(session)` branch silently did nothing. The symmetric key-match resolution is a genuine simplification over the per-action role table, and moving the capture ahead of the take-waiter interception is the right call — that reply is consumed and never reaches the arms.

I verified the pieces this depends on: `create_session` errors on duplicates rather than overwriting (so the take-race ordering in `take_order` is safe), `save_trade` is `INSERT OR REPLACE` but has only two production call sites and both precede the reveal, `create_order` already writes a maker row so `update_trade_counterparty` has something to hit, the `ACTIVE_CHATS` guard makes repeat spawns real no-ops, and the global DM filter carries no `since`, which is what makes the "every replayed reveal self-heals" claim actually true. `cargo clippy --all-targets` is clean and `cargo test --lib` passes (296/296).

One moderate issue and a handful of smaller ones.

---

### 🟠 Moderate

**1. The new invariant guard is node-scoped, so it misses exactly the rows it exists to catch on every node but the active one.**

`chat_still_relevant` rejects `counterparty_pubkey == active_mostro_pubkey()`. But the `trades` table has no node column:

```sql
CREATE TABLE IF NOT EXISTS trades (
    id TEXT PRIMARY KEY, data TEXT NOT NULL, status TEXT NOT NULL,
    started_at INTEGER NOT NULL, completed_at INTEGER
);
```

and `resubscribe_active_chats` iterates `db.list_trades()` — every row, from every node the user has ever pointed at (`set_active_mostro_pubkey` / `refresh_subscriptions_for_active_node` clear the order book, not the trades). So a pre-fix row poisoned with node **B**'s pubkey sails through the guard while node **A** is active, derives garbage chat keys, and claims `ACTIVE_CHATS` — and `subscribe_incoming_chat` has *no idle timeout*, so it holds that claim for the life of the process and the self-healing reveal's spawn becomes a silent no-op. That is precisely the failure this guard was added to prevent, one node over.

A node-agnostic guard is available on the same row and is exact rather than heuristic:

```rust
&& trade.counterparty_pubkey != trade.order.creator_pubkey
```

The poison *is* `creator_pubkey` — that was `take_order`'s old seed — so this catches it regardless of which node published the 38383 event, needs no config read, and can never reject a legitimate peer (a peer trade pubkey is never the event author). Keeping both checks is fine too, but the `creator_pubkey` one is the load-bearing half.

---

### 🟡 Minor

**2. The trade keys are derived twice per qualifying message.** `maybe_capture_peer_reveal` loads `get_active_trade_keys(trade_index)` (l. 2314) to run `resolve_peer_side`, then calls `on_peer_pubkey_received`, which loads the *same* keys again (l. 2373) before delegating to `apply_peer_reveal`. Since `maybe_capture_peer_reveal` is now the only production caller of `on_peer_pubkey_received`, calling `apply_peer_reveal(order_id, &peer_hex, &trade_keys, trade_index, role)` directly drops a BIP-32 derivation plus an identity-lock read from every reveal — and makes `role: Option<TradeRole>` unnecessary (it is `None` only in `peer_pubkey_with_no_session_does_not_panic`).

**3. No short-circuit once the peer is already captured.** Widening capture from two actions to "any payload naming both trade pubkeys" means the whole path — key derivation ×2, DB `UPDATE`, ECDH, HKDF, session update, task spawn — now runs for essentially every daemon message in a trade's life, and again for the entire replayed history on each restart (the global DM filter deliberately has no `since`). An early return when the session already holds this exact peer and a shared key preserves the self-heal property for rows that missed it while making the overwhelmingly common case free. It also silences finding **4** on web.

**4. The IndexedDB stub warns on every message.** `update_trade_counterparty` logs at `warn` unconditionally. Unlike `mark_trade_rated` (once per trade), this now fires for every qualifying daemon message and every replayed one, so a web session's log fills with a message that carries no new information after the first. `log::debug!` — or a once-per-order warn — keeps the signal without the spam.

**5. `take_order` returns a `TradeInfo` it has just contradicted in the DB.** The replay block writes the peer to the row but leaves the returned struct's `counterparty_pubkey` empty, so the value handed across the bridge disagrees with what was persisted a line earlier. Harmless today — `take_order_screen.dart` discards the return and re-reads through `rawTradesProvider` — but `TradeInfo.counterpartyPubkey` is exactly what `tradeInfoToChatRoom` gates the chat room on, so a future caller that trusts the return value gets no chat room. One line (`trade.counterparty_pubkey = peer.clone();`) keeps them consistent.

**6. `maybe_capture_peer_reveal` itself has no test.** The tests cover `resolve_peer_side` (pure) and `apply_peer_reveal` (session creation) — both welcome — but not the layer whose behavior changed most: payload extraction (`Order` vs `PaymentRequest(Some(_), ..)`), the both-pubkeys-required rule, the terminal guard, the `BondSlashed` exemption, and the durable write. The identity dependency is what makes it awkward, and it is the same reason `apply_peer_reveal` was split out — the same trick works here: lift the payload → `(buyer_pk, seller_pk)` extraction into a pure helper and table-test it (single-sided payload, unparseable hex, non-`Order` payload). That covers the new decision logic without touching process-global identity state.

---

### 🔵 Nits

**7. Stale comment left behind in the `BuyerTookOrder | HoldInvoicePaymentAccepted` arm.** The terminal-guard comment still reads "a stale replay over a finished trade must not re-derive the peer key, recreate session state, or respawn the chat subscription either" — none of which this arm does any more. Behavior is fine (`maybe_capture_peer_reveal` carries its own `status_sync_blocked_by_terminal`), but the guard is now evaluated twice per message for these two actions and the comment describes the pre-#334 code.

**8. `let small_order = ... o.clone()`** in that same arm is now the only remaining use, and only `small_order.status` is read — the clone can go (pre-existing, but the arm shrank enough that it stands out).

---

### ✅ Verified

- `create_session` returns `SessionAlreadyExists` instead of overwriting, so the take-race path (`reveal creates session → take_order's create fails → replay persists from the session`) cannot lose the peer or the shared key. ✔
- `save_trade` (`INSERT OR REPLACE`, full-row) has exactly two production call sites, both before any reveal, so the `json_set` write cannot be clobbered. ✔
- `create_order` persists a maker row with an empty `counterparty_pubkey`, so the maker's durable write lands on a real row. ✔
- `resolve_peer_side`'s `TradeRole` matches the codebase's "our role" convention, and the maker's session role therefore agrees with the trade row's. ✔
- `session.order` is write-only in this crate, so the book-sourced `OrderInfo` used when the reveal creates the session cannot diverge into anything observable. ✔
- Placement is after the generation gate and after the local→daemon UUID reconciliation, so the capture always sees the daemon id. ✔

Requesting changes on **1**; the rest are cheap enough to fold into the same round.

Catrya (changes requested):
## Changes requested — the fix is right and I verified it against the real database; the contract did not come with it

I confirmed the premise on disk, measured the effect on the user's own trade rows, and drove the capture end to end. Everything in the description holds. What is missing is the spec update the repo's own rule requires.

## The bug is real, and it is in the database right now

`order_events.rs:77-78` is literal: `creator_pubkey = event.pubkey.to_hex()`, with the comment "creator_pubkey is the Mostro node's pubkey (the event author)". Seeding `counterparty_pubkey` from it stores the daemon.

Reading a real `mostro.db` (read-only copy): **27 trades, 5 with `counterparty_pubkey == creator_pubkey == the Mostro node`**. Running each branch's real `chat_still_relevant` over those rows:

| | `main` | this PR |
|---|---|---|
| Chat subscriptions armed at startup | **2** | **0** |
| …of which poisoned | **2** | 0 |

```
6a0f3c7b Success               -> false
2251082f Active                -> TRUE   <- arms a subscription with garbage keys
5066fa12 Canceled              -> false
3a24a01c FiatSent              -> TRUE   <- same
4ecb5857 CooperativelyCanceled -> false
```

So on that machine **every chat listener the app arms on restart is built from the daemon's pubkey** — not "somerows are bad", all of the ones that get that far.

One detail confirms the reasoning in the `chat_still_relevant` comment. In the probe the active node was thecompiled-in default, not the node those rows came from, so the `active_mostro_pubkey` comparison caught **none** ofthe five. The row-local `creator_pubkey` comparison caught all of them. The comment calling it "the load-bearingone" is exactly right, and the second check really is defence in depth.

## The durable write repairs a real poisoned row

Applying the new `update_trade_counterparty` to order `2251082f` in a writable copy of that database:

```
before:  counterparty=00000018…  chat_still_relevant=false
after:   counterparty=aaaaaaaa…  chat_still_relevant=true
```

Survives closing and reopening the database, and the empty-value write is refused. The full cycle is demonstrated on the real row shape: poisoned → excluded by the guard → repaired by the write → relevant again.

## The capture, driven end to end

Seeding an identity, creating the row as a take leaves it (counterparty empty), then feeding **one** daemon messagenaming both trade pubkeys through `dispatch_mostro_message`. Same test, both branches:

| | `main` | this PR |
|---|---|---|
| `counterparty_pubkey` on the row | **`<empty>`** | the peer's real pubkey |
| Session | **none** | created, with peer and `shared_key` |

That is #334 in one line: in the maker's shape `main` loses the counterparty entirely — no row value, no session, so no chat is possible ever.

## Blocking: the contract does not accompany the change

The repo rule is explicit — *"Specs are a living artifact: a behaviour change updates its contract in the same PR."* This PR changes two things `specs/004` describes:

- `contracts/orders.md:328` gives the `BuyerTookOrder` / `HoldInvoicePaymentAccepted` row as *"`status → Active`"*.When that was written, the arm also did the peer reveal. The reveal is now a cross-cutting step **before** dispatchthat applies to any action whose payload names both trade pubkeys, and no row in the table mentions it. Someoneimplementing from the contract would not implement the capture.
- `contracts/orders.md:105-122` says `take_order` is where "the trade session/subscriptions start". With this PR the session may already exist (created by the reveal), the duplicate `create_session` is an expected error, and`counterparty_pubkey` is no longer seeded there.

Two paragraphs: one line in `orders.md` for the pre-dispatch capture (with the `BondSlashed` exemption), and acorrection to the `take_order` section. While there, `types.md:153` lists `counterparty_pubkey` with no semantics —worth a sentence saying when it is populated and that the row is the durable record while the session is a cache.

## Minor

- **The seam this PR creates is undefended: deleting the wiring leaves all 341 tests green.** Mutating each piece:

  | Mutation | Caught by a test? |
  |---|---|
  | Drop the two new `chat_still_relevant` conditions | **yes** |
  | Swap the sides in `resolve_peer_side` | **yes** |
  | Allow an empty pubkey in `update_trade_counterparty` | **yes** |
  | Restore `counterparty_pubkey: order.creator_pubkey` in `take_order` | no |
  | **Delete the `maybe_capture_peer_reveal` call from the dispatcher** | no |

  The last one is the whole feature. I wrote the missing test to check the objection in the doc comment — thatcovering it means touching the process-global identity — and it is **writable in about ninety lines**, using onlywhat the file already has plus `import_from_mnemonic`. It does have to be `#[ignore]`d, which is the pattern thisrepo already uses for anything that cannot share process state (`restore_e2e_tests`, the `app_db` `OnceCell`). Thatis the C table above; happy to hand over the probe.

  The `take_order` seed mutation matters less than it looks, and it is worth saying why: its consequence is nowdefended by the `chat_still_relevant` guard, which **is** covered. The poison can come back silently, but it nolonger reaches key derivation.

- **`peer_reveal_pubkeys` accepts empty strings where mostrix rejects them.** The reference does `if buyer_s.is_empty() || seller_s.is_empty() { return None; }` (`chat_utils.rs:256-258`); here a `Some("")` passes thefilter and dies later in `PublicKey::from_hex`, on the branch that logs `warn!("unparseable trade pubkeys inpayload")`. Safe, but this function runs for every daemon message and for the whole replayed history on eachrestart, so a daemon emitting `Some("")` instead of `None` would fill the log. Two lines to match the reference.

## Also verified

- **Ordering in the dispatcher is correct and load-bearing**: generation gate → local→daemon UUID reconciliation →**capture** → take-waiter. The capture must follow the reconciliation, because it writes by `order.id` and by thenthe row has been rebound to the daemon id.
- **`trade_index` really is ours**: on the global path `resolve_dm_recipient` has already pinned the addressed trade key and `unwrap_mostro_message` decrypted with those keys, so the "ours by construction" claim holds.
- **"Mirrors mostrix" is accurate**: `~/Github/mostrix/src/util/chat_utils.rs:251-277` does the same symmetric match and returns `None` when we are neither party.
- **The UI does learn about it.** The capture emits no `TradeUpdate`, but in the normal flow the same messagecontinues into the status arm, which does. The one case where it does not — a take's first reply, consumed by thewaiter — is covered deliberately: the new block at the end of `take_order` replays the write and mirrors the peeronto the returned struct, with the comment explaining that `tradeInfoToChatRoom` gates on that field.
- **One row per `order.id`** in the real database (27 rows, 27 distinct ids), so the `WHEREjson_extract(data,'$.order.id') = ?` update touches exactly one.
- **CI on the tree merged with current `main`**: merges clean despite being **64 commits behind**; `cargo test--locked` → **341 passed, 0 failed**; `cargo clippy --locked -- -D warnings` → clean; `cargo check --locked --target wasm32-unknown-unknown` → clean.

Catrya (approved):
tACK

PR #347, opened 2026-08-31
fix: persist counterparty pubkey so the maker's chat resolves the peer

### LINKED ISSUE #346: Maker's trade chat shows Unknown instead of the counterparty's nym
As a maker, the trade chat header (and the chat list entry) always shows the localized **\"Unknown\" / \"Desconocido\"** instead of the counterparty's deterministic nym, with the default teal avatar. The v1 client (MostroP2P/mobile) correctly shows a handle like \`trustless-phoenix\` in the same situation.

### 1. Rust: the counterparty pubkey is never persisted for makers

The maker's trade row is created with an empty \`counterparty_pubkey\` (\`create_order\`, \`rust/src/api/orders.rs\`) and nothing ever writes it back. When the daemon's \`BuyerTookOrder\` / \`HoldInvoicePaymentAccepted\` message arrives, \`on_peer_pubkey_received\` derives the ECDH shared key and chat keys and stores the peer pubkey **only in the in-memory session**  the trade row in the DB keeps its empty counterparty forever.

Consequences beyond the cosmetic nym:

- \`tradeInfoToChatRoom\` (Dart) filters out trades with an empty \`counterpartyPubkey\`, so the maker's chat room can never resolve the peer identity.
- \`resubscribe_active_chats\` (\`rust/src/api/messages.rs\`) filters on a non-empty \`counterparty_pubkey\`, so **a maker who restarts the app silently loses the incoming-chat subscription** — the buyer's messages stop arriving until the daemon happens to resend a peer notification.

Takers are unaffected: \`take_order\` stores the maker's pubkey at take time.

### 2. Dart: the chat room screen depends on the chat-list tab having run first

\`ChatRoomScreen\` resolves its room from \`chatRoomsNotifierProvider\`, but the only code that hydrates that notifier from the trade DB is \`ChatRoomsScreen.initState\` (the Chat tab). Opening a chat directly  trade-detail chat chip, deep link  without having visited the Chat tab in that session falls back to a placeholder with an empty handle, which renders as \"Unknown\" even when the pubkey is available in the DB. Sending/receiving a message then upserts that empty placeholder into the room list, cementing the \"Unknown\" entry there too.

Forte11Cuba:
Close #346 

## Problem

`ChatRoomScreen` renders its header from `chatRoomsNotifierProvider`, but only the chat-list screen populates it. Opening a chat any other way — the trade-detail chat chip, a deep link — lands on an empty placeholder and shows "Unknown", even when the trade row has the counterparty.

## Fix

`lib/features/chat/screens/chat_room_screen.dart` only:

- The screen listens to `tradeInfoProvider(orderId)` and upserts its room whenever the provider emits the trade; an `initState` read covers the case where the provider is already alive with data. Since `rawTradesProvider` refetches itself on every trade update, the daemon reveal that fills the counterparty reaches an open screen as a fresh emission — the header flips to the nym live, with no manual refresh or retries.
- Before upserting, the hydrated room is merged with the live list entry: if the live entry is at least as recent, its `lastMessage` / `unreadCount` are kept, so a message racing the hydration isn't clobbered and a just-zeroed unread badge isn't resurrected.

## Tests

`test/features/chat/chat_room_hydration_test.dart` (bridge-free, like its siblings):

1. Direct entry resolves the peer identity from the trade row.
2. Hydration keeps the newer live preview and doesn't resurrect the unread count.
3. A reveal landing while the screen is open flips the header.


Catrya (changes requested):
Reviewed at `48659ae`, merged locally against current `main` (`7bbfa1d` — the branch is 16 commits behind and merges clean). On that merge: **315 Rust tests** pass, `cargo clippy --locked -- -D warnings` and `cargo check --locked --target wasm32-unknown-unknown` are clean, `flutter analyze` reports nothing, `flutter test` → **292 pass**. `frb-generate.sh` produces no drift.

The maker fix is real and the session-lifecycle gap from the previous round is genuinely closed. But rebuilding sessions inside `resubscribe_active_chats` turns a pre-existing bad row into an outgoing channel, so I can't approve this as it stands.

## Blocking: the session rebuild makes the trade chat publish to the Mostro node

`take_order` seeds the trade row with `counterparty_pubkey: order.creator_pubkey` (`orders.rs:833`), and on a Kind 38383 book order `creator_pubkey` is **the Mostro node's own pubkey** — `nostr/order_events.rs:77` says so in as many words ("creator_pubkey is the Mostro node's pubkey (the event author)"). On `main` nothing ever writes that field back, so every taken trade keeps it.

This is not hypothetical. In my local `mostro.db`, 6 of 27 trade rows carry my mostrod's pubkey as `counterparty_pubkey`, and two of them are still live with no `outcome` (one `FiatSent`, one `Active`). `chat_still_relevant` accepts both — non-empty counterparty, live status.

With this PR, `resubscribe_active_chats` now runs

```rust
upsert_peer_session(&trade, &trade.counterparty_pubkey, shared_key)
```

for exactly those rows, so the session's `peer_pubkey` becomes the daemon's key. `send_message` then takes its normal path — `session.peer_pubkey` → `chat_context` → `derive_chat_keys` → `publish_chat_payload` — and puts the message on the relay.

I verified this by running the real code rather than reading it. A scratch test that saves the row `take_order` produces today and calls `resubscribe_active_chats`:

```
PROBE: session peer_pubkey = Some("82fa8cb978b43c79b2156585bac2c011176a21d2aead6d9f7c575c005be88390")
test result: ok. 1 passed
```

The same probe on clean `main` fails with `PROBE: no session was built` — today there is no session, `send_message` stays on its local-only path, and **nothing leaves the device**. With this PR it does.

And the recipient can read it. The chat ECDH is symmetric, so whoever holds the secret behind that pubkey derives the same `K_conv`. Also executed:

```
PROBE: the node derives the identical K_conv/K_sign — it can decrypt the chat
```

So the net effect is: trade chat content encrypted to the Mostro node instead of the counterparty, decryptable by the node operator, never delivered to the peer, while the UI shows it as sent. That is strictly worse than the current silent local-only behavior.

**Fix, either way:**

- **Preferred — land #345 first and rebase this on top.** #345 fixes the root cause (stops seeding `creator_pubkey` in `take_order`) and adds the `chat_still_relevant` guard rejecting a counterparty equal to the active Mostro pubkey, which covers rows that are already poisoned. I confirmed it: with `main + #345` my probe fails at `assert!(chat_still_relevant(&trade))`, i.e. the bad row is correctly excluded from chat recovery.
- **Minimum inside this PR:** the same guard in `chat_still_relevant`, plus dropping the `creator_pubkey` seed. Note the guard alone only compares against the *active* node, so it would not cover a row seeded by a different node — which is why not writing that pubkey in the first place is the real fix.

**Merge order matters: #345 and #347 conflict.** Merging one after the other produces conflicts in `rust/src/api/orders.rs` and `rust/src/db/sqlite.rs` — both add `update_trade_counterparty` and both touch `on_peer_pubkey_received`. One of them has to be the base.

## Minor

- **The PR body describes the dangerous case incorrectly.** "Trades taken before this fix keep their empty counterparty" is true for makers, but a taker's row is *not* empty — it holds the daemon pubkey, which is precisely the row that triggers the blocker above. Worth rewording so the case isn't hidden.
- **The seam this PR creates has no test.** The three new tests cover `upsert_peer_session` and the SQLite round-trip, but nothing exercises `resubscribe_active_chats` → session, which is the new seam and where the defect is. A test there would have caught it — my probe is literally that test.
- **The maker fix does not apply on web.** `IndexedDbStorage::get_trade_by_order_id` returns `Ok(None)` (#233), so the new branch in `on_peer_pubkey_received` falls through to the warn and a web maker stays local-only. Not a regression, but the PR reads as if the gap were closed everywhere. #345 covers this by building the session without depending on the row.
- **`update_trade_counterparty` does not refuse an empty pubkey.** The only caller already guarantees non-empty, so there is no live bug, but it is a new trait seam that can blank a good value. The equivalent in #345 rejects it.
- **Dart: `_hydrateRoom()` is invoked from `build()`.** While `room.peerPubkey` is empty, every rebuild fires `ref.read(tradeInfoProvider(...).future)` — one DB round-trip per rebuild. No infinite loop (`tradeInfoToChatRoom` returns `null` for an empty pubkey, so nothing is upserted to feed it back), but a simple `_hydrating` flag would bound it.
- **`refreshTrades(ref)` invalidates `rawTradesProvider` wholesale** on every `tradeUpdate` for this order. Works, but broader than needed.

## What I verified

- Merge against today's `main`, full CI as above.
- The blocker chain, executed: with this PR the session is built with the daemon's pubkey; on clean `main` no session is built; with `main + #345` the row is excluded by `chat_still_relevant`. ECDH symmetry proven with the real `derive_chat_keys`.
- No identity-key exposure anywhere in this change: what is persisted is `SmallOrder.{buyer,seller}_trade_pubkey` (`orders.rs:1825`), the `master_*_pubkey` fields are never referenced in this repo, `resubscribe_active_chats` derives from `get_active_trade_keys(trade.trade_key_index)`, and the chat envelope is still kind 14 signed with `K_sign` plus an inner kind 1 signed by the trade key.
- Cross-PR conflict with #345.

Catrya (approved):
tACK

PR #350, opened 2026-08-31
fix(db): index the trade lookups that run once per order event

grunch:
Phase 1, PR 1.2 of the optimization plan (#348).

## Problem

Because `trades.id` is a fresh UUID for takers, every lookup by order id has to reach inside the JSON blob:

```sql
WHERE json_extract(data, '$.order.id') = ?
```

Six queries in `sqlite.rs` use that predicate — `get_trade_by_order_id`, `delete_trade_by_order_id`, `update_trade_order_id`, `update_trade_fields`, `update_trade_peer_reputation`, `mark_trade_rated` — and **none of them had an index**, so each was a full table scan that re-parsed every row's JSON.

This is not a cold path. `local_trade_status` (`orders.rs:2246`) runs one of these for **every non-pending Kind 38383 event**, and the reputation path adds another per rated trade, so the cost scales with both order-book size and trade history.

## Change

One `CREATE INDEX IF NOT EXISTS` added to `SQLITE_INIT_SQL`: `idx_trades_order_id`, an expression index on `json_extract(data, '$.order.id')`.

SQLite only uses an expression index when the query's expression matches the indexed one exactly, so there's a comment in `schema.rs` tying the two together; if someone reformats the predicate in `sqlite.rs`, the index silently stops applying. That is what the test is there to catch.

`SQLITE_INIT_SQL` already runs unconditionally on every `SqliteStorage::open()` and every statement in it is idempotent, so existing databases pick the index up on next launch — no migration step, no `SCHEMA_VERSION` bump needed (nothing keys off it at runtime).

## Dropped after review (fb6900f)

An earlier revision also added `idx_trades_status`. Review on #348 caught that it was unreachable, and that is confirmed: the 30-minute stale sweep calls `db.list_trades()`, which is `SELECT id, data FROM trades ORDER BY started_at DESC` with no `WHERE`, and filters status in Rust afterwards (`orders.rs:2637-2648`). No query in the crate filters on `trades.status` at all, so that index bought nothing while costing an extra B-tree write on every trade insert and update.

Pushing the sweep's filter into SQL would make a status index worth having, but it changes what the sweep does and belongs in its own PR.

## Test plan

- [x] New test asserts via `EXPLAIN QUERY PLAN` that the order-id predicate hits its index. Confirmed it fails first (`expected the order-id expression index, got: SCAN trades`) and passes after.

### DIFF
--- a/rust/src/db/schema.rs
+++ b/rust/src/db/schema.rs
@@ -60,6 +60,15 @@ CREATE TABLE IF NOT EXISTS trades (
     completed_at    INTEGER
 );
 
+-- `trades.id` is a fresh UUID for takers, so every lookup by order id has to
+-- reach inside the JSON blob. The expression here must stay byte-identical to
+-- the one in the six `WHERE json_extract(data, '$.order.id') = ?` queries in
+-- sqlite.rs, or SQLite silently falls back to a full scan that re-parses every
+-- row — on the ingest path that runs once per non-pending order event.
+CREATE INDEX IF NOT EXISTS idx_trades_order_id
+    ON trades(json_extract(data, '$.order.id'));
+CREATE INDEX IF NOT EXISTS idx_trades_status ON trades(status);
+
 -- Chat history + durable replay dedup (issue #246). `trade_id` here is the
 -- **order id** — the identity chat keys are derived from — which for taken
 -- orders differs from the `trades.id` UUID, so deliberately NO foreign key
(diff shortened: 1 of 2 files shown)


grunch:
Good catch from the review on #348 — verified and fixed in fb6900f.

`idx_trades_status` was unreachable. The 30-minute stale sweep calls `db.list_trades()`, which is:

```sql
SELECT id, data FROM trades ORDER BY started_at DESC
```

No `WHERE` clause — it retrieves every row and filters status in Rust afterwards (`orders.rs:2637-2648`). I grepped the crate: **no query filters on `trades.status` at all.** So that index was never used by anything, and an unused index is not neutral — it costs an extra B-tree write on every trade insert and update.

Dropped it, along with its test. The order-id expression index stays; six queries genuinely use it, and that one is the whole point of the PR.

Pushing the sweep's filter into SQL is the right follow-up and would make a status index worth having, but it changes what the sweep does and deserves its own PR rather than riding along with an index change.

`cargo test --lib db::sqlite` — 13 passed.

PR #352, opened 2026-08-31
fix(perf): make daemon-message dedup a hash lookup

grunch:
Phase 1, PR 1.4 of the optimization plan (#348).

## Problem

`is_duplicate_daemon_message` (`orders.rs:319`) held a `VecDeque<String>` of up to 512 hex event ids and answered "have I seen this?" with:

```rust
if guard.iter().any(|id| id == event_id) {
```

That is a linear scan of up to 512 64-character string comparisons, taken **under the mutex**, once per kind 14 event — on both call sites, the per-trade subscription (`orders.rs:1249`) and the global daemon feed (`orders.rs:3043`). The window is at its fullest exactly when it matters most: during a history replay, which is the case the window exists for.

## Change

Membership moves to a `HashSet<String>`. The `VecDeque` stays, but its only job now is recording insertion order so eviction drops the right id:

```rust
struct DedupWindow {
    seen: HashSet<String>,
    order: VecDeque<String>,
}
```

Same 512-entry window, same semantics, O(1) instead of O(n). The `MAX_ENTRIES` constant moves out to `DEDUP_MAX_ENTRIES` at module scope so the test can assert against the real capacity rather than a hardcoded copy.

The two structures have to be mutated together or `seen` grows without bound, which is the one thing that could go wrong here — hence the comment on the struct and the eviction half of the test.

### Not done here

The plan also suggested keying on `EventId` bytes instead of hex strings. Left out: both call sites already need the hex string for their log lines, so the allocation does not actually disappear, and changing the signature would touch code beyond this fix. Worth revisiting when the logging around it changes.

## Test plan

- [x] New test asserts both halves: a repeated id is reported as duplicate, and an id is forgotten once `DEDUP_MAX_ENTRIES` newer ones have arrived. Confirmed it fails to build against the old code (no such constant) and passes after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon message deduplication to reliably detect repeated messages.
  * Ensured older deduplication entries are evicted when the tracking limit is reached.

* **Tests**
  * Added coverage for duplicate detection and removal of the oldest tracked entry.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -319,12 +319,38 @@ const DEDUP_MAX_ENTRIES: usize = 512;
 /// the per-trade and the global subscription is only handled once.
 ///
 /// `seen` answers the membership question; `order` exists only to know which
-/// id to drop when the window is full. The two must be mutated together or the
-/// set grows without bound.
+/// id to drop when the window is full. Both hold the same `Arc<str>`, so a new
+/// id is allocated once, and `record` is the only thing that writes them —
+/// split those two writes across call sites and `seen` grows without bound.
+///
+/// `frb(ignore)` because this module is part of `crate::api`, which
+/// flutter_rust_bridge scans: without it the codegen emits bindings for this
+/// private, non-bridgeable type and the wasm build stops compiling.
 #[derive(Default)]
+#[flutter_rust_bridge::frb(ignore)]
 struct DedupWindow {
-    seen: std::collections::HashSet<String>,
-    order: std::collections::VecDeque<String>,
+    seen: std::collections::HashSet<Arc<str>>,
+    order: std::collections::VecDeque<Arc<str>>,
+}
+
+impl DedupWindow {
+    /// Returns `true` if `event_id` is already in the window. Otherwise records
+    /// it — evicting the oldest id when the window is full — and returns
+    /// `false`.
+    fn record(&mut self, event_id: &str) -> bool {
+        if self.seen.contains(event_id) {
+            return true;
+        }
+        let id: Arc<str> = Arc::from(event_id);
+        self.seen.insert(Arc::clone(&id));
+        self.order.push_back(id);
+        if self.order.len() > DEDUP_MAX_ENTRIES {
+            if let Some(evicted) = self.order.pop_front() {
+                self.seen.remove(&evicted);
+            }
+        }
+        false
+    }
 }
 
 static PROCESSED_GW: OnceLock<std::sync::Mutex<DedupWindow>> = OnceLock::new();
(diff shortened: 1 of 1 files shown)


grunch:
@coderabbitai review

grunch:
Strict self-review of this PR — three findings, all fixed in 9e952a6.

---

**1. CRITICAL — the new struct lands on the bridge surface (this is the red CI check)**

`DedupWindow` is declared in `crate::api::orders`, and `crate::api` is exactly what flutter_rust_bridge scans. The codegen emitted bindings for a private, non-bridgeable type, so the web job's `Build Rust core to WASM` step failed with 16 errors:

```
error[E0425]: cannot find type `DedupWindow` in this scope   (x14)
error[E0603]: struct `DedupWindow` is private
error[E0277]: the trait bound `DedupWindow: SseEncode` is not satisfied
```

Only the web job caught it because that job regenerates the bindings; the native jobs built against the committed `frb_generated.rs`, which predates this struct. `ResetGuard` in the same file survives only by accident — it is a unit struct, which the parser skips outright.

Fixed with `#[flutter_rust_bridge::frb(ignore)]`, matching `restore_session` in this file. Verified by running `./scripts/frb-generate.sh`: `DedupWindow` no longer appears in the generated output, and the regenerated file is byte-identical to the committed one once `rustfmt` is applied — so no generated file needed to change.

**2. MEDIUM — the allocation count doubled, in a perf PR**

```rust
if !guard.seen.insert(event_id.to_string()) { ... }
guard.order.push_back(event_id.to_string());   // second allocation
```

The `VecDeque` code this replaces allocated once per new id. `seen` and `order` now share one `Arc<str>`, so it is one allocation again and eviction removes by the same pointer.

**3. MEDIUM — the test drives the process-global window**

It called `is_duplicate_daemon_message` directly, so it assumed `PROCESSED_GW` was pristine on entry and left 513 entries in it for every later test in the binary — flaky in both directions the moment anything else touches that window. The logic moved to `DedupWindow::record`, `is_duplicate_daemon_message` is now just the lock wrapper, and the test drives a local instance.

While there, the test now also asserts the invariant the struct's own comment names as the one thing that can go wrong: `seen.len() == order.len()`, and neither exceeds `DEDUP_MAX_ENTRIES`.

---

Not changed: the decision to keep hex `String` ids rather than `EventId` bytes still reads correctly — both call sites need the hex for their log lines, so nothing would be saved.

**Verification**
- `cargo test` — 294 passed, 0 failed
- `cargo clippy --all-targets` — no errors, no new warnings
- `./scripts/frb-generate.sh` — clean, no binding changes
- `./scripts/build-web.sh` — the step that was failing now completes: `✓ WASM build complete`

PR #353, opened 2026-08-31
fix(perf): stop cloning the whole trade-key map per daemon message

grunch:
Phase 1, PR 1.5 of the optimization plan (#348).

## Problem

The global daemon-message path did this for **every** kind 14 event (`orders.rs:3295`):

```rust
let keys = global_dm_keys().read().await.clone();
handle_global_daemon_message(&event, &keys).await;
```

That clones a `HashMap<String, (Keys, u32)>` holding **one entry per lifetime trade** — hundreds for an active user — and the handler then uses exactly one of them: the entry matching the event's `p` tag.

## Change

The lookup moves into `resolve_dm_recipient`, which takes the read guard, finds the matching key, clones just that one, and returns. `handle_global_daemon_message` now receives the resolved `(hex, keys, index)` directly instead of the map.

The read guard is deliberately *not* held across the handler: handling a message can reach `ensure_global_dm_coverage`, which takes the same lock for writing (`orders.rs:2915`). Holding it would deadlock. That constraint is now written down at the resolver, since it is the reason the obvious "just pass the guard" version is wrong.

The `no-matching-p-tag` warning moves with the lookup and is otherwise unchanged.

## Dropped from this PR

The plan also listed the `map.retain(...)` sweep in `lock_order` (`orders.rs:155`) as an O(n)-per-acquisition cost. On inspection that one is **not worth changing**: the eager sweep is what keeps the registry at roughly the number of concurrently-dispatched orders (a handful), so the scan it performs is over a map that its own eagerness keeps tiny. Making it periodic would trade a real, tested invariant — `the_registry_drops_locks_no_handler_holds` asserts finished locks are gone immediately — for no measurable saving, and it fails that test. Left alone deliberately; the plan entry should be considered closed rather than pending.

## Test plan

- [x] New test asserts `resolve_dm_recipient` returns the trade key whose pubkey matches the event's `p` tag (with its index), and `None` for an event addressed to someone else. Confirmed it fails to build against the old code.


grunch:
## Review — strict pass

The perf claim holds and the core refactor is correct: the tag scan, the first-match-wins semantics, the `no-matching-p-tag` warn, and the ordering of the duplicate check relative to dispatch are all preserved, and the `None` case skips the handler exactly where the old code returned early. The read guard is held only across a synchronous loop and one log call — no `.await` under it — so the deadlock against `ensure_global_dm_coverage`'s write lock that the new doc warns about cannot happen. Single caller, correctly guarded.

Two defects in the diff itself, both fixed in df44064.

### 1. The doc block ended up on the wrong function (`orders.rs`)

The new function was inserted directly beneath the existing doc comment for `handle_global_daemon_message` without moving it, so the block

> Handle a kind-14 Mostro reply received on the global subscription. […] decrypts via `mostro_core::transport::unwrap_incoming`, and dispatches the recovered `Message` through `dispatch_mostro_message`.

attached to `resolve_dm_recipient`, which does none of those things — while `handle_global_daemon_message`, the function that does all of them, was left with no doc at all. Split back apart, with the "finds which trade key" sentence rewritten since that responsibility genuinely moved out.

### 2. The new test leaks into a process-global map, making an existing test flaky (`orders.rs`)

`dm_recipient_is_resolved_from_the_p_tag` inserted into `global_dm_keys()` and never removed the entry. That map is a `'static` shared by every test in this binary, and the suite runs in parallel — `a_late_derived_key_joins_the_global_dm_coverage` asserts that

```rust
let before = global_dm_keys().read().await.len();
ensure_global_dm_coverage(&keys, 91).await;
assert_eq!(global_dm_keys().read().await.len(), before);
```

holds across the call. An insert landing between those two reads fails it, by scheduling alone. The other tests touching this map are written to tolerate a shared one (`contains_key`, relative counts); this was the first to add a permanent entry.

Both events are now built up front, both lookups run back to back, and the entry is removed *before* any assertion — so a panicking assertion cannot leave the map grown for whoever runs next either. The negative case never depended on the map's contents (a freshly generated stranger key is never in it), so it is unaffected.

### Branch was stale

The branch was based on `9465974`, missing #348–#352, and conflicted with `main`: both sides had added a test at the top of the `orders` test module. Purely positional — merged in `9f94543`, keeping both.

### Not findings

- `cargo fmt --check` reports drift across ~32 files including this one, but it is repo-wide and pre-existing; CI does not run it.
- The 23 `MutexGuard held across await` clippy warnings are all in `escrow.rs` / `reputation.rs` / `settings.rs`, untouched here.

### Verification

`cargo test` — 330 passed, 0 failed. `cargo clippy --all-targets` — no new warnings. Both after the merge with `main`.

### On the dropped `map.retain` item

Agreed, and the reasoning in the description is the right one: the sweep's eagerness is what bounds the map it scans, and `the_registry_drops_locks_no_handler_holds` encodes that as an invariant. Closing rather than deferring that plan entry is correct.

Catrya (approved):
Reviewed at `df440643`, merged locally against current `main` (`5c72e41` — the branch is 3 commits behind and merges clean). On that merge: **330 Rust tests** pass, `cargo clippy --locked -- -D warnings` and `cargo check --locked --target wasm32-unknown-unknown` are clean.

**Approving.** No regressions found, and I measured the change rather than assuming it.

## Behaviour is exactly equivalent

- **Ordering is unchanged.** Before: clone the map → handler → p-tag lookup → warn + return if none; the dedup check came after. Now: lookup + warn in the resolver → handler only on a hit → dedup. In both, a kind-14 with no matching key never enters the dedup window.
- **Multiple `p` tags**: first match wins in both (`break` → `return`).
- **The `no-matching-p-tag` warn** keeps its format and its `map.len()`.
- **The control flow in the kind-14 branch is right**, which was the one place a slip would have been serious: the `continue` sits *outside* the `if let`, so an unresolvable kind-14 still never reaches `ingest_order_event`.
- Single caller of `handle_global_daemon_message`, so the signature change leaves nothing dangling. The per-trade path (`subscribe_daemon_messages`) has its own key and does not go through here.

**The deadlock the plan warned about is genuinely avoided.** Plan item 1.5 flags that the guard cannot outlive the lookup because handling a message can reach `ensure_global_dm_coverage`, which takes the same lock for writing (`orders.rs:2988`). The guard is a local of `resolve_dm_recipient` and is dropped on return; the handler's `.await` runs with no guard held. It also *reduces* contention: the read lock used to be held across a ~50-entry clone, now it is held across a bounded tag scan with no await inside.

**The "Dropped from this PR" claim checks out.** `map.retain(...)` is untouched at `orders.rs:155`, and `the_registry_drops_locks_no_handler_holds` — the test cited as the reason to leave it alone — exists at `orders.rs:5072`.

## Measurements

Micro-benchmark against the real `global_dm_keys()` global and the real `resolve_dm_recipient`, with the deleted two lines recreated as the "old" side. 2000 iterations per point, `--release`, after warm-up; three consecutive runs agree within ±3%:

```
  entries |   old (clone+lookup) |   new (resolve only) | speedup
  --------+----------------------+----------------------+--------
        1 |             262 ns/ev |             327 ns/ev |   0.8x
       10 |            1433 ns/ev |             206 ns/ev |   7.0x
       50 |            4405 ns/ev |             165 ns/ev |  26.7x
      101 |            8086 ns/ev |             153 ns/ev |  52.9x
      250 |           20235 ns/ev |             167 ns/ev | 121.3x
```

The multiplier is not the point — the shape is. The old column is linear (~81 ns per map entry); the new one is **flat** at ~150-170 ns regardless of size. At one entry the new path is *slower* by ~65 ns, i.e. noise: below ~10 keys this change buys nothing.

Then a real session against a live daemon (`flutter run -d linux`, ~10 minutes):

| | |
|---|---|
| `trade key map: 50 keys derived for daemon-message decryption` | **50 entries** |
| kind-14 events reaching this path (one clone each on `main`) | **186** |
| — `reason=duplicate` / `no-matching-p-tag` | 0 / 0 |
| raw kind=14 frames across both relays | 297 (the 111 extra are the second relay's redelivery, deduped by the SDK before the `Event` notification) |
| window they arrived in | **2 seconds** (163 + 23) |
| relay traffic in the following 9m36s | none |

Crossing the two, at the session's real map size of 50:

| | time spent in this path |
|---|---|
| `main` today | **818 µs** |
| with this PR | **32 µs** |
| saved | **≈0.79 ms**, all inside a 2-second startup burst |

Scaling with the map, same 186-event burst: 1.50 ms at 101 entries, 3.74 ms at 250.

So: the absolute saving today is sub-millisecond, and I would not claim more than that. What makes it worth merging is that it lands inside the startup burst — the moment the app is busiest — and that it removes an O(n) term over a map that grows one entry per lifetime trade key and is never pruned. In steady state the saving is zero, because no events arrive at all.

## Non-blocking notes

1. **The new test writes to a process-global that another test asserts the size of.** `dm_recipient_is_resolved_from_the_p_tag` inserts into and removes from `global_dm_keys()`, while `a_late_derived_key_joins_the_global_dm_coverage` does `let before = len(); …; assert_eq!(len(), before)`. Any insert or remove landing between those two reads fails it. The test's own comment sees the hazard, but removing the entry before *its own* assertions does not close the window for the other test, which reads at an arbitrary moment.

   To be fair, the race pre-exists: `seeding_coverage_never_evicts_existing_keys` already writes to that map via `ensure_global_dm_coverage`. This PR adds a third writer and widens the window rather than creating it. Empirically I saw one failure in 80 runs of those two tests on this branch, then zero in 500 more on the branch and 250 on `main` — too few samples to attribute, so I am reporting the mechanism, not a confirmed diagnosis.

   A cleaner shape: split out a pure `fn resolve_from_map(event: &Event, map: &HashMap<…>) -> Option<…>` and let `resolve_dm_recipient` be the wrapper that takes the guard. The test then uses a local `HashMap`, touches no global, and still exercises the real matching logic.

2. **The seam this PR creates — the call site in `_run_order_subscription` — is not covered.** The test exercises `resolve_dm_recipient` in isolation; reverting the call site to `global_dm_keys().read().await.clone()` would leave it green. Hard to test directly ("no clone happens"), so not something I would ask for, but worth knowing the headline claim is not pinned.

3. **Nit**: the `blog_warn` is emitted with the read guard still held, to read `map.len()`. It is a sync call so there is no deadlock, but it holds the RwLock across formatting and the log-buffer write. Copying `map.len()` into a local and dropping the guard first is one line.

PR #355, opened 2026-08-31
fix(observability): log order-book stream lag

grunch:
Phase 1, PR 1.7 of the optimization plan (#348). **Prerequisite for the Phase 3 delta work.**

## Problem

```rust
Err(broadcast::error::RecvError::Lagged(_)) => continue,
```

`OrdersStream::next` (`orders.rs:3506`) swallowed lag silently. The sibling `TradeUpdatesStream::next` logs the identical condition with a comment explaining why dropping is tolerable — the order-book stream just discarded it.

The channel held **16** snapshots (`orders.rs:186`). A cold start or a refetch publishes one snapshot per ingested event, far more than 16 before the UI reads any of them, so overflowing was the normal case rather than an exceptional one — and there was no way to see it happening.

## Change

- Lag is logged at warn, with the count.
- Capacity becomes a named `ORDER_STREAM_CAPACITY = 64`, sized for the burst rather than steady state.

Both the constant and the log record *why* dropping is currently survivable — every message is a full snapshot, so the newest supersedes anything lost — and that this stops being true the moment the channel carries deltas. That is the condition Phase 3 changes, which is why this PR comes first: the delta work needs lag to mean "resync", and it needs to be visible before it can mean anything.

Widening the window reduces how often lag happens; it does not fix the reason it happens, which is PR 2.1/2.2 (one broadcast per ingested event).

## Test plan

- [x] New test publishes 100 snapshots without reading, then asserts the subscriber resumes at the oldest **retained** snapshot rather than closing — which pins the retained-window depth behaviourally, not just the constant's value. Confirmed it fails to build against the old code.

Note: the test covers the recovery contract and the window depth. The warn line itself is not asserted (no log capture in this suite) — it was verified by reading, and mirrors the already-reviewed trade-updates handler.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved order-book stream recovery when a consumer falls behind, allowing updates to resume from the oldest retained snapshot.
  - Added warning information indicating how many snapshots were dropped during stream lag, improving visibility into missed updates.
  - Improved reliability when processing high-volume sequences of order-book snapshots.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -175,6 +175,15 @@ pub struct OrderBook {
     tx: broadcast::Sender<Vec<OrderInfo>>,
 }
 
+/// Snapshots retained for a subscriber that has fallen behind.
+///
+/// A cold-start or refetch burst publishes far more updates than the UI reads
+/// in the same instant, so this is sized for that burst rather than for steady
+/// state. While each message is a full snapshot, overflowing is survivable —
+/// the newest snapshot supersedes the dropped ones. That stops being true if
+/// this channel ever carries deltas.
+const ORDER_STREAM_CAPACITY: usize = 64;
+
 impl Default for OrderBook {
     fn default() -> Self {
         Self::new()
@@ -183,7 +192,7 @@ impl Default for OrderBook {
 
 impl OrderBook {
     pub fn new() -> Self {
-        let (tx, _) = broadcast::channel(16);
+        let (tx, _) = broadcast::channel(ORDER_STREAM_CAPACITY);
         Self {
             orders: Arc::new(RwLock::new(Vec::new())),
             tx,
@@ -3638,7 +3647,15 @@ impl OrdersStream {
         loop {
             match self.rx.recv().await {
                 Ok(orders) => return Some(orders),
-                Err(broadcast::error::RecvError::Lagged(_)) => continue,
+                // Each message is a full snapshot, so dropping some is
+                // survivable: the next one carries the whole book. Log it
+                // anyway — this is the only backpressure signal there is, and
+                // it stops being harmless the moment this channel carries
+                // deltas instead of snapshots.
+                Err(broadcast::error::RecvError::Lagged(n)) => {
+                    log::warn!("[orders] order-book stream lagged, dropped {n} snapshots");
+                    continue;
+                }
                 Err(broadcast::error::RecvError::Closed) => return None,
             }
         }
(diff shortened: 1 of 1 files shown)


Catrya (changes requested):
Changes requested — the logging half is right and I'd take it as-is. The capacity bump is the problem.

The warn! is good

OrdersStream::next was the only Lagged arm in this file that discarded the count silently; the three siblings (:1314, :3477, :3569) all log it, and this now matches them. The named constant with a doc comment is an improvement over the bare 16 regardless of its value. The test earns its place too: pinning the retained-window depth behaviourally rather than asserting the constant against itself is the right shape.

Blocker: ORDER_STREAM_CAPACITY = 64 is an anti-optimization while this channel carries snapshots

Every message here is the whole book — set_orders / upsert_order / clear clone and send a full Vec<OrderInfo> (orders.rs:203-227). Under "latest wins" semantics, the small buffer is the coalescing: lag discards work that is already superseded, for free. Making the buffer bigger doesn't stop the consumer from falling behind; it just makes it drag more dead weight when it catches up.

Measured rather than reasoned — a tokio::broadcast probe, 2000-send burst, slow consumer:

capacity=16: snapshots delivered to the consumer after the burst = 16, lag events = 1
capacity=64: snapshots delivered to the consumer after the burst = 64, lag events = 1

When the burst ends, the receiver has to drain capacity snapshots sequentially, and the first capacity - 1 of them are already superseded by the last. At 64 that is 48 extra redundant emissions at the tail of every cold start, node switch and pull-to-refresh, and each one costs:

- a full FRB bridge crossing with the entire book (2000 OrderInfo serialized), plus
- a complete map + filter + sort on the Dart side — home_order_providers.dart:158-161 re-materializes every OrderItem per emission, and filteredOrdersProvider (:169) re-filters and re-sorts the whole list.

That is a 4× multiplier on exactly the cost PRs 2.1/2.2 exist to remove, landing in exactly the scenario the plan is written against.

Worst-case retained memory scales the same way. I confirmed tokio frees a slot's value once the last receiver has read it (broadcast.rs:1753), so this is only paid while a subscriber is behind — but "while a subscriber is behind" is precisely the startup burst. At 2000 orders, ~600 KB per snapshot means the retained window goes from ~10 MB to ~38 MB on mobile.

The PR body's justification — "widening the window reduces how often lag happens" — is true, and that is the issue: it reduces lag by doing more work. Lag on a snapshot channel isn't a failure mode, it's the mechanism that protects the UI.

Buffer depth will genuinely matter in Phase 3, when the channel carries deltas and lag has to mean "resync". But it should be chosen then, together with the resync mechanism it exists to serve — sizing it now, ahead of the semantics that give it meaning, buys nothing and costs the drain tail above.

Suggested change: keep the named constant and the doc comment, set it to 16, and reword the comment to say why a small window is correct for snapshots and what has to change when deltas land. The new test passes unmodified — with SENT = 100 it just resumes at order-84 instead of order-36.

Minor

- The test leans on an undocumented tokio detail: broadcast::channel rounds capacity up to the next power of two (tokio/src/sync/broadcast.rs:583). Harmless at 16 or 64, but SENT - ORDER_STREAM_CAPACITY would fail confusingly if anyone ever sets it to, say, 48 or 100. Worth a one-line comment on the constant.
- The plan item asked for a test that observes the log; this one doesn't. Called out honestly in the PR body, and the warn! line is a literal mirror of the already-reviewed trade-updates handler, so I'm fine with it.

grunch:
Review round 1 addressed in ae5b14e (and `main` merged in 8563374 — the conflict was both sides appending tests at the same spot in `mod tests`; kept both).

- `ORDER_STREAM_CAPACITY` is back to **16**. Agreed on the reasoning: on a snapshot channel lag is the coalescing, so a wider buffer only lengthens the stale drain tail. The doc comment now says exactly that and that the depth has to be re-chosen with the resync mechanism once deltas land.
- Added the note that tokio rounds the capacity up to a power of two, since the test counts on the exact value.
- Test unchanged apart from the overflow guard becoming a const assertion (clippy `assertions_on_constants`); with 16 it now resumes at `order-84`.
- PR title updated — nothing is widened any more.

`cargo clippy --locked -- -D warnings` clean, 331 tests pass.

grunch:
Rebased onto current `main` (8dd5f5a). The only conflict was additive: `main` had grown new tests at the same spot in `orders.rs`'s `mod tests` (the bulk-ingest and dedup-window ones from #352/#351); both sides are kept. One adjustment while here: the `assert!(SENT > ORDER_STREAM_CAPACITY, ...)` guard is now a `const _: () = assert!(...)` — clippy on the current toolchain flags a runtime assert with a constant value, and a compile-time check is what it meant anyway. `cargo test`: 333 passed; `cargo clippy --all-targets`: no new warnings.

Issue #305, opened 2026-08-18
Taker reputation is received but never shown


Catrya:
The daemon sends the maker an extra DM with the taker's
reputation: a `Peer` payload with an empty `pubkey` and
`reputation: Some(UserInfo { rating, reviews, operating_days })`, riding the
same `PayInvoice` / `AddInvoice` action as the flow message
(`notify_taker_reputation()` in mostrod's `src/util.rs`).

The app parses it — it shows up in the logs — but nothing surfaces it in the
UI, so the maker never sees who took their order:

gift-wrap: action=PayInvoice order_id=Some(9e60e970-…) trade_pubkey=bfff4ce5
payload=Peer(Peer { pubkey: "", reputation: Some(UserInfo { rating: 4.375, reviews: 4, operating_days: 64 }) })

The order book already renders the maker's reputation from the kind 38383
`rating` tag, so this is the missing counterpart side.

**Expected:** show it where the maker decides whether to continue — the
pay/add invoice screen and the trade detail — deriving the counterpart's role
from the user's own (paying the hold invoice ⇒ the taker is the buyer; adding
an invoice ⇒ the taker is the seller).

**Note:** a brand-new user and a full-privacy taker both arrive as zeros, so
they are indistinguishable on the wire.

Issue #273, opened 2026-08-05
fix(account): "Generate new user" keeps the previous identity's trades and chats


Catrya:
Generating a new user rotates the identity but never deletes the data derived
from the old one. `delete_identity()` (`rust/src/api/identity.rs:287`) clears
the in-memory identity, the persisted identity row, the trade-key mappings and
the logs — but the **`trades` table (full My Trades history) and the
`messages` table (chat history) survive**, and the in-memory sessions are not
emptied either. The new identity starts with fresh keys yet inherits the
previous user's entire trade list and conversations.

This is a privacy issue, not just cosmetic: the point of "Generate new user"
is a clean start. The code's own comment shows the intent ("a new mnemonic
must start counting from zero instead of inheriting") but it was only applied
to keys, not data. The leftover rows are also functionally dead — their trade
keys were cleared, so nothing can operate on those orders anymore.

## Reproduce

Account → Generate new user → My Trades still lists every order (and chat)
from the previous identity.

## Proposed fix

- Add `clear_trades` / `clear_messages` to the DB trait (+ SQLite and
  IndexedDB backends), mirroring the existing `clear_trade_keys`.
- Call them from `delete_identity()`, and empty the in-memory
  `session_manager` there too.
- Dart: invalidate `rawTradesProvider` after `IdentityService.regenerate()`
  so the UI reflects the clean slate immediately.

Related: the configurable retention setting (#131) will need the same bulk
DB operations.

codaMW:
Taking this. Plan: add clear_trades / clear_messages to the Storage trait (mirroring clear_trade_keys) with SQLite + IndexedDB impls, call them from delete_identity() alongside the existing trade-key cleanup, empty the in-memory session_manager there too, and invalidate rawTradesProvider after regenerate() on the Dart side so My Trades reflects the clean slate immediately.

grunch:
Superseded by #533, which covers the same bug with the full scope: besides the `trades` and `messages` tables, the rotation also leaves behind the in-memory `DISPUTE_STORE` and `RATING_STORE` (the user sees the previous identity's disputes), the Dart notifications store and its `processed_events` ledger, `bond_claims`, `queued_messages`, the `is_mine` marking on `orders`, and the `mostro-dm` subscription filter still listing the old trade pubkeys. #533 also covers the import path, which goes through the same `delete_identity()`.

[...]

PR #358, opened 2026-08-31
fix(ui): build each theme once and make push init idempotent

grunch:
Phase 1, PR 1.11 of the optimization plan (#348). Two unrelated lifecycle papercuts.

## 1. Themes were rebuilt on every frame that rebuilt the app

`MostroApp.build` called `buildLightTheme()` and `buildDarkTheme()` inline (`app.dart:54-55`), and each call constructed a full `ThemeData` — colour scheme, text theme, every component theme. `MaterialApp.router` rebuilds on every locale change, theme-mode change and route change, and asks for **both** themes each time, including the one not currently in use.

`ThemeData` is immutable, so both are now built once and shared. The memoization lives in `app_theme.dart` rather than in the widget, so the ~15 widget tests and the golden harness that call these directly get the same benefit.

## 2. `PushNotificationService.initialize` could attach duplicate listeners

Steps 4, 5 and 7 attach `onMessage`, `onMessageOpenedApp` and `onTokenRefresh` listeners, and none of the subscriptions is stored or cancelled. Nothing guarded against a second run: `_initialized` was only *written* at the end and read somewhere else entirely. A second call would double every foreground notification and every token re-registration.

`initialize` now claims a guard flag before its first `await` — at the top, because the awaits in between are exactly where a concurrent second caller would slip through.

The guard is a **separate flag** from `_initialized`, which was tempting to reuse but wrong: `_initialized` gates `_fcm.deleteToken()` in `unregisterAllTokens`, so it has to stay false on the paths that bail out early (Firebase unavailable, permission denied). Setting it at the top would make `unregisterAllTokens` call `deleteToken` against a Firebase that was never available.

Today `initialize` is called once, from a post-frame callback in `app.dart:32`, so this is a latent bug rather than an active one — but the guard is one line and the failure mode is user-visible duplicate notifications.

## Dropped from this PR

The plan also listed `escrowModeProvider`'s `while (true) yield await stream.next();` as missing a null/close guard. It isn't: unlike the orders stream, `EscrowModeStream::next` returns `Result` and **bails** on a closed channel (`rust/src/api/escrow.rs:172-182`), so the Dart side sees a thrown error, the loop unwinds, and the provider enters its error state. There is no null to check and no hot loop. Its non-`autoDispose` lifetime is likewise deliberate — it is an app-wide capability gate fed by a rare wake-up event (node switch), not a per-screen subscription. No change made.

## Test plan

- [x] New tests: each builder returns the same instance across calls, and the two themes stay distinct (different `brightness`). Confirmed they fail against the old code.
- [x] `flutter test` — **260 passed, 0 failed** (`main`: 257 passed, 0 failed; this PR adds 3 tests)
- [x] `flutter analyze` — **7 issues, all `info`-level deprecation notices, 0 errors** — identical to `main`

Both figures are measured after running the two generation steps CI runs first (`./scripts/frb-generate.sh`, `flutter gen-l10n`), since `lib/src/rust/` and `lib/l10n/app_localizations*.dart` are gitignored. An earlier revision of this description reported pre-existing failures and a broken `test/support/fake_trades.dart`; that was stale generated code in my own checkout, not a repo problem. See the correction comment below.

### DIFF
--- a/lib/core/app_theme.dart
+++ b/lib/core/app_theme.dart
@@ -402,13 +402,19 @@ const _light = AppColors(
 
 // ── ThemeData factories ────────────────────────────────────────────────────────
 
-ThemeData buildDarkTheme() => _buildTheme(
+// Built once and shared. ThemeData is immutable, and `MaterialApp.router` asks
+// for both themes on every rebuild — which includes every locale change,
+// theme-mode change and route change.
+ThemeData? _darkTheme;
+ThemeData? _lightTheme;
+
+ThemeData buildDarkTheme() => _darkTheme ??= _buildTheme(
   brightness: Brightness.dark,
   colors: _dark,
   scaffold: const Color(0xFF1B1E28),
 );
 
-ThemeData buildLightTheme() => _buildTheme(
+ThemeData buildLightTheme() => _lightTheme ??= _buildTheme(
   brightness: Brightness.light,
   colors: _light,
   scaffold: const Color(0xFFFFFFFF),
--- a/lib/features/notifications/services/push_notification_service.dart
+++ b/lib/features/notifications/services/push_notification_service.dart
@@ -28,6 +28,9 @@ class PushNotificationService {
   FirebaseMessaging get _fcm => _fcmInstance ??= FirebaseMessaging.instance;
   String? _token;
   bool _initialized = false;
+
+  /// Guards [initialize] against a second run attaching duplicate listeners.
+  bool _initStarted = false;
   SharedPreferences? _cachedPrefs;
   final Set<String> _registeredTradePubkeys = {};
 
@@ -36,6 +39,13 @@ class PushNotificationService {
 
   Future<void> initialize({ProviderContainer? container}) async {
     if (!_isSupported) return;
+    // Steps 4, 5 and 7 below attach stream listeners that are never
+    // cancelled, so a second run would double every foreground notification
+    // and every token re-registration. This is a separate flag from
+    // `_initialized`, which means "there is a token to delete" and must stay
+    // false on the paths that bail out below.
+    if (_initStarted) return;
+    _initStarted = true;
 
     // Bail out if Firebase hasn't been initialized (placeholder firebase_options).
     try {
--- a/test/core/app_theme_test.dart
+++ b/test/core/app_theme_test.dart
@@ -1,0 +1,22 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mostro/core/app_theme.dart';
+
+void main() {
+  group('theme construction', () {
+    // MaterialApp.router rebuilds on every locale, theme-mode and route
+    // change, and it asks for both themes each time. ThemeData is immutable,
+    // so there is nothing to gain from rebuilding it.
+    test('returns the same light theme instance every time', () {
+      expect(identical(buildLightTheme(), buildLightTheme()), isTrue);
+    });
+
+    test('returns the same dark theme instance every time', () {
+      expect(identical(buildDarkTheme(), buildDarkTheme()), isTrue);
+    });
+
+    test('keeps the two themes distinct', () {
+      expect(identical(buildLightTheme(), buildDarkTheme()), isFalse);
+      expect(buildLightTheme().brightness, isNot(buildDarkTheme().brightness));
+    });
+  });
+}


grunch:
## Correction: my earlier test numbers were measured against a stale local build

Retracting what I wrote in the test plan above. There are **no pre-existing test failures and no broken test helper**. The failures I reported came from my own working copy, not from the repo.

`lib/src/rust/` and `lib/l10n/app_localizations*.dart` are **gitignored** (`.gitignore:9,13`) — generated, never committed. CI regenerates both before it analyzes or tests (`ci.yml:109` `./scripts/frb-generate.sh`, `:116` `flutter gen-l10n`, then `:130`/`:133`). My checkout had stale copies predating the `peer_rating` / `peer_reviews` / `peer_days` fields (`rust/src/api/types.rs:271-275`, issue #305) and the newer l10n keys, so 27 analyzer errors and 10 unloadable test files were purely local staleness.

After running the same two generation steps CI runs:

| | `main` | this branch |
|---|---|---|
| `flutter test` | **257 passed, 0 failed** | see below |
| `flutter analyze` | **7 issues, all `info`-level deprecations, 0 errors** | 7, identical |

Every Dart branch in this Phase 1 series was re-verified against that corrected baseline, and all are fully green:

- #356 — 258 passed (257 + 1 new test)
- #357 — 260 passed (257 + 3 new tests), goldens unchanged
- #358 — 260 passed (257 + 3 new tests)

So: `test/support/fake_trades.dart` is correct as written, those 10 test files **do** run in CI, and the code changes in this PR are unaffected — only my reported numbers were wrong. Apologies for the noise.

Catrya (approved):
**Approving.** Both halves are correct. I measured the theme half rather than taking the reasoning on trust, and the numbers ask for one correction to the description plus one comment in the code.

## Correctness

**The memoization is safe because `_buildTheme` is pure.** It takes only `brightness`, `colors` and `scaffold`, and at both call sites those are compile-time constants (`_dark`/`_light` are `const AppColors`, the scaffold colours are literals). It reads no `BuildContext`, `MediaQuery`, locale or platform — the only `BuildContext` in the file belongs to `_NoTransitionBuilder.buildTransitions`, which is not part of theme construction. `ThemeData` is immutable, so sharing one instance between the app, the ~15 widget tests and `golden_harness.dart` cannot diverge.

**The push guard is placed and named correctly.** `if (!_isSupported) return;` stays *above* it, so unsupported platforms never set it. And keeping it separate from `_initialized` is right: that flag gates `_fcm.deleteToken()` in `unregisterAllTokens` (`:214`), so settieToken` against a Firebase that was neveravailable. The "called once today" premise holds — `initialize` has exactly one caller in `lib/`, the post-frame
callback in `_MostroAppState.initState` (`a

**The three new tests pin what the PR changheme(), buildLightTheme())` cannot passagainst the old code — and the third one rules out the obvious mistake of memoizing both into the same field.

## Measurements

**Cost per rebuild.** A test calling `buildme()` 2000 times, run in two worktrees:

| | per `MostroApp.build` |
|---|---|
| `main` | **94.13 µs** |
| this PR | **0.09 µs** |

**How often that actually happens.** I instrumented `MostroApp.build` with a counter and `_buildTheme` with a print, ran the app against a live daemon and drove the UI:

| action | `MostroApp.build` | `_buildTheme
|---|---|---|
| startup | 1 | 2 (light + dark) |
| back out of the trade detail | 1 | 2 |
| order book → My Trades → Chat → My Trades | 1 | 2 |
| open Settings | 1 | 2 |
| **change Appearance to Light** | **2** | **2** |

Two things follow.

The change **works**: at the second root re. On `main` that point would show 4
constructions; here it stays at 2.

But **the description overstates the trigger**. It says `MaterialApp.router` rebuilds "on every locale change,
theme-mode change and route change". Route ostroApp` — I did five and the counterstayed at 1. That is expected once you look: `MostroApp.build` only watches `localeProvider` and
`settingsProvider.select(themeMode)`; routilApp.router` without rebuilding its parent.
So the real triggers are language and themer actions. The saving in a normal session is a few tenths of a millisecond, not a steady drip. Worth fixing the sentence so the next reader does not carry the
wrong model.

## The one thing I would like added: a comment about hot reload

Memoizing into top-level fields means theme edits no longer show up on hot reload. Flutter documents this — *"Global
variables and static fields are treated as reinitialized during hot reload"* — and Iconfirmed it on the running app. With the app in light mode I changed the light scaffold to magenta (`0xFFFFFFFF` →
`0xFFFF00AA`) and hot-reloaded:

```
Performing hot reload...
[probe] MostroApp.build #3
Reloaded 84 of 2113 libraries in 3.201ms
[probe] _buildTheme RAN    ← never printed
```

The screen stayed white. The tree rebuilt, the cached theme survived, the new colour never applied.

And the documented workaround does not exist here: **hot restart is broken in this app**, independently of this PR —

```
Performing hot restart...
Restarted application in 982ms.                                                                                   Unhandled Exception: AnyhowException(Alread
#3  bootstrapAndRun (package:mostro/core/app_bootstrap.dart:153)                                                  ```
                                                                                                                  — the Rust core cannot be re-initialized ine-existing and nothing to do with this PR(which touches only `app_theme.dart` and th that with this change in, the only way tosee a colour edit is to stop the app and rent next to `_darkTheme` / `_lightTheme`saying so would save the next person a conf
One trap to note if anyone tries to "fix" it: wrapping the cache in `kReleaseMode` would break the three new testwhich run in debug.
                                                                                                                 ## Non-blocking notes
                                                                                                                 1. **The guard also blocks legitimate retris claimed above the two early returns —Firebase unavailable (`:41-46`) and permission denied (`:53-57`) — so an `initialize` that failed for either reascan never be retried for the life of the pra single caller, but it constrains whatcomes next: an "enable notifications" buttodo nothing instead of re-prompting. Twolines fix it — set `_initStarted = false` bprotection against a concurrent secondcaller is unaffected, since the flag stays held across every intermediate `await`.                               
2. **Context for the `_initialized` reasoning**: `unregisterAllTokens` currently has **no callers outside the    service**. The argument for not reusing thait is defensive rather than an activeprotection — presumably it becomes live when identity deletion is wired up (#273 / #298). Worth saying so nobody goes hunting for the path.
3. **The numbers in the description are stale**: it reports `flutter test — 260 passed` and `flutter analyze — 7 issues, all info-level`. Against today's `m tests** and **0 analyze issues**. Not adefect, but worth refreshing so the descrip## What I could not testThe push half. `firebase_options` is a plac`initialize` exits at `[main] Firebase notconfigured … push notifications disabled` bteners. The guard is verified by reading,not exercised. Testing it for real needs eioject or making `FirebaseMessaging`injectable — a separate PR.

PR #360, opened 2026-08-31
fix(perf): publish the order book once per refetch, not once per event

grunch:
Phase 2, PR 2.1 of the optimization plan (#348). **The single biggest defect in the plan.**

## Problem

`refetch_active_node_orders` (`orders.rs:2755`) replays the node's entire book through `ingest_order_event`, one event at a time:

```rust
for event in events.into_iter() {
    ingest_order_event(&event).await;
}
```

and every `upsert_order` clones the whole `Vec<OrderInfo>` and broadcasts it (`orders.rs:209-219`). So ingesting N orders costs **N clones of an N-element vector** and **N full-book payloads across the FRB bridge** — O(N²) in both.

That path runs on exactly two triggers, and both replay the whole book: the node switch (refresh_subscriptions_for_active_node, orders.rs:2972) and the Account screen's refresh (account_screen.dart:520 → restart_orders_subscription, orders.rs:2816). At the few thousand orders Mostro now carries, one of those is ~2M struct clones and ~2000 whole-book bridge crossings, each of which the Dart side then re-maps, re-filters and re-sorts.

Cold start is not one of them, and this PR does not fix it. Coming online goes through subscribe_orders() → _run_order_subscription(), where the relay's stored events arrive one notification at a time and land in ingest_order_event — so the first fill of the book keeps the per-event emission and its quadratic cost. There is also no pull-to-refresh on the order book: the only RefreshIndicators in the app are on Trades and Notifications.

## Change

Ingest takes a publish mode:

```rust
enum Publish {
    Immediately,     // live subscription: one event, one emission
    WhenBatchEnds,   // bulk ingest: caller publishes once
}
```

The live subscription path is unchanged — `ingest_order_event` still means "publish immediately". The refetch defers and calls `order_book().publish()` once when the batch is done. N events now cost N upserts and **one** snapshot.

`upsert_order` keeps its exact semantics, including taking the snapshot while still holding the write lock; the shared insert logic is factored into `apply_upsert` so the deferred variant cannot drift from it.

This does not change *what* the UI ends up displaying — the published snapshot is the same book the last of the N emissions would have carried. It removes the N-1 intermediate emissions nobody could see.

### Not the whole fix

The live path still emits a full snapshot per relay event; coalescing that is PR 2.2, and replacing snapshots with deltas is Phase 3. This PR is the one that removes the quadratic term.

## Bridge surface

Both new methods are `pub(crate)`, so nothing is added to the FRB API. `./scripts/frb-generate.sh --check` passes with no regeneration needed and no diff in `lib/src/rust/`.

## Test plan

- [x] New test: 50 deferred upserts produce **zero** emissions, then `publish()` produces **exactly one** snapshot carrying all 50. Confirmed it fails to build against the old code (no such methods).
- [x] New regression test: a live `upsert_order` still publishes immediately, so deferring stays opt-in.
- [x] `cargo test` — 295 passed, 0 failed
- [x] `./scripts/frb-generate.sh --check` — clean
- [ ] Manual: switch nodes (or hit the Account screen's refresh) against a node with a large book, confirm the stall is gone and the list still fills


Catrya (changes requested):
**Changes requested** — the code is right and the saving is real (I measured two orders of magnitude). What needs fixing is the description: it credits this change with removing the cold-start stall, and cold start never touches this path.

## Blocker: `refetch_active_node_orders` does not run on cold start, and there is no pull-to-refresh

The Problem section says it *"runs on cold start, on every node switch … and on every pull-to-refresh"*. I followed all three:

- **Cold start:** the pool goes Online → `nostr.rs:58` calls `subscribe_orders()` → `_run_order_subscription()`. The relay's stored events arrive **one at a time** as notifications and land in `ingest_order_event(&event)` (`orders.rs:3350`) — i.e. `upsert_order`, one emission per event. `subscribe_orders` never calls the refetch, and `rehydrate_active_mostro_node` (`settings.rs:217`) only writes config; it starts no subscriptions. **Cold start keeps the full O(N²).**
- **Pull-to-refresh:** the order book has none. The only `RefreshIndicator`s in the app are on Trades and Notifications; Home's retry button is `ref.)` (`home_screen.dart:270`), which isDart-side only.
- **What this does fix, and it is not small:** exactly two paths — node switch (`settings.rs:207`               `set_active_mostro_node` → `refresh_subscri the Account screen's refresh(`account_screen.dart:520` → `restart_order
                                                                                                                The magnitude claim holds. Measured in rele a live subscriber attached:
                                                                                                                ```
n=500 :  per-event= 124.9 ms   batched=  1.5 ms    83x                                                          n=2000:  per-event=2193.5 ms   batched= 40.```
                                                                                                                Those 2.19 s at 2000 orders are literally tdescription names — **and they are stillthere on cold start**, because that path isg this (or the plan's M2 row, *"Cold start /refresh / node-switch stalls eliminated"*) will believe startup is fixed when it isn't; the "Not the whole fix"   section already says the opposite of the "Pthe description costs nothing and keeps therecord honest, the way this plan already does for 1.6 and 1.10.                                                   On who owns cold start: PR 2.2's debounce cyou want something sharper, **thesubscription loop already receives `RelayMed currently only logs it** — that is thenatural batch boundary for the startup bursrimitive this PR introduces. Not a one-liner(EOSE arrives per relay, and its ordering ad event notifications needs checking), butworth evaluating before declaring startup c

## Minor

- **Doc rot introduced by the patch.** `enutween** `ingest_order_event`'s doc commentand the function, so that block — *"Parse a it into the order book … Shared by the livesubscription loop and the node-switch refete book identically"* — now documents theenum, and both ingest functions end up undo is precisely what belongs on`ingest_order_event_with`, which is what thg the enum above the block fixes it. - **An empty batch costs one extra emissions `Ok` with zero events, `publish()` re-emits an unchanged book, and on the Dart side thaer + re-sort for nothing. It used to be zero emissions. An `if !events.is_empty()` cover- **The tests cover the primitive, not the  operate on `OrderBook` directly; neitherexercises the caller publishing. If someonesh().await` from`refetch_active_node_orders`, all 331 tests stay green and a node switch would leave the UI **empty** until the next live event — because `refresh_subscriptionser_book().clear()` immediately before. Thatfailure was impossible before this PR. A caller-level test drags in the pool and process-global state (`app_db` is a `OnceCell`), so the cheap version is to extract the loop into an `ingest_batch(events)` and assert "N events →exactly one emission" there.

## What I verified

- **The deferral is complete:** inside `ingest_order_event_with` there is no other `order_book()` call besides the final upsert, so nothing in a batch escapes through a second route.
- **CI on the tree merged with current `main`:** `cargo test --locked` → **331 passed, 0 failed**; `cargo clippy --locked -- -D warnings` (CI's exact command) → clean. The 23 `--all-targets` errors (`MutexGuard` held across anawait) are **identical on `origin/main`** — pre-existing, not this PR.
- **#349's demotion to `debug!` survives the merge** — checked in the merged result rather than assumed, since thisbranch is based on a commit predating #349 and edits the same function.
- The benchmark above, and that the two callers of the refetch are the ones named (searched every caller on both the Rust and Dart sides).

Catrya (dismissed):
ACK

PR #361, opened 2026-08-31
perf(bridge): coalesce relay-driven book updates into one emission

grunch:
Phase 2, PR 2.2 of the optimization plan (#348).

> **Stacked on #360** — it introduces the deferred-upsert primitive this builds on. Base is `perf/batch-order-ingest`; it will retarget to `main` once #360 merges. Review #360 first.

## Problem

PR 2.1 removed the quadratic cost of a *refetch*. The steady-state firehose is still untouched: every Kind 38383 event that arrives calls `upsert_order`, which clones the whole `Vec<OrderInfo>` and broadcasts it (`orders.rs:209-219`).

So on a book of a few thousand orders, each individual order update — a status change, a new order, an expiry — costs a full-list clone, a full-list payload across the FRB bridge, and a full re-map + re-filter + re-sort on the Dart side (`home_order_providers.dart:144-219`). Orders arrive in bursts, and the UI cannot consume snapshots anywhere near that rate. There is no throttling anywhere in the bridge path today.

## Change

Relay ingest publishes through a **200 ms trailing window**. A burst inside one window collapses to a single emission carrying the settled book.

- The first update arms the window; later ones inside it just mutate the book.
- The flag is cleared **before** the snapshot is read, so an update landing mid-publish opens a new window instead of being swallowed. That ordering is the one thing that could silently drop an update, so it is commented in place.
- The window re-arms on the next update — covered by its own test, since a "publish once then go quiet forever" bug would be invisible until someone left the app open.

Trailing rather than leading was chosen deliberately: the subscriber sees the settled book rather than the first state of a burst plus a correction 200 ms later.

### Scope

Only the firehose is coalesced. `Publish::Coalesced` applies to `ingest_order_event`; **daemon-message handlers and user actions keep publishing immediately** (`orders.rs:903`, `:1506`, `:1905`, `:2463`, `:3566`) — those fire once per user action and are latency-sensitive, and there is nothing to coalesce.

`set_orders`, `clear`, `remove_order` and `update_order_status` are likewise unchanged.

## Bridge surface

`upsert_order_coalesced` is `pub(crate)`. `./scripts/frb-generate.sh --check` passes — no regeneration needed.

## Test plan

- [x] New test: a 20-event burst publishes **nothing** before the window closes, then **exactly one** snapshot carrying all 20. Confirmed it fails to build against the old code.
- [x] New test: a later update opens a new window (the flag re-arms).
- [x] Existing regression test renamed and kept — a direct `upsert_order` still publishes immediately, so coalescing stays opt-in.
- [x] `cargo test` — 297 passed, 0 failed
- [ ] Manual: against a live node, confirm new orders still appear promptly (the 200 ms should not be perceptible) and status changes still land


Catrya (approved):
**Approved.** Measured end to end against a local relay: this is the PR that removes the cold-start stall, and the part of the design that could have silently dropped an update is correct — I verified it by counterexample.

## End-to-end measurement (local `nostr-rs-relay`, 2000 synthetic Kind 38383 orders)

Published 2000 orders under a throwaway "mostro" key, then started the relay pool and `subscribe_orders()` exactly as the app does, counting whole-book snapshots until the book settled. Same relay, same book, both branches, release build:

| | `main` | this PR (on #360) |
|---|---|---|
| Final book | 2000 | 2000 |
| **Snapshots emitted** | **2000** | **1** |
| Read by the subscriber | 1520 | 1 |
| **Dropped by channel lag** | **480** | 0 |
| First emission | 525 ms | 712 ms |
| **Whole book on screen** | **2.21 s** | **0.71 s** |

Three things that only show up on the real path:

1. **On `main` the channel had to drop 480 whole-book snapshots**, silently — and that is with a consumer that does nothing but count. The real consumer is Dart re-mapping, re-filtering and re-sorting 2000 orders per emission, so in the app the drop count would be far worse.
2. **Time to a complete book goes 2.21 s → 0.71 s**, and those 712 ms include the relay delivering 2000 events. The cost stops being the client and becomes the network.
3. **The trailing window's price is visible and small:** first paint arrives ~190 ms later (525 → 712 ms). In exchange it carries the whole book instead of 1/2000th of it, and it lands 1.5 s sooner. Worth stating in the description as a measured trade-off rather than an argued one.

## Synthetic harness through `ingest_order_event` (no relay, both branches)

Feeding N events through the real live-ingest path isolates the client-side cost:

| N | `main` | this PR |
|---|---|---|
| 50 | 0.94 ms · 50 emissions · 0.019 ms/order | 0.24 ms · **1** · 0.005 ms/order |
| 200 | 15.8 ms · 200 · 0.079 ms/order | 0.96 ms · **1** · 0.005 ms/order |
| 1000 | 414 ms · 1000 · 0.414 ms/order | 4.5 ms · **1** · 0.004 ms/order |
| 3000 | **4.18 s** · 3000 · 1.393 ms/order | **78 ms** · **1** · 0.026 ms/order |

The number that matters is the per-order column on `main`: 0.019 → 0.079 → 0.414 → 1.393, scaling with N. That is the O(N²) signature, measured rather than argued. Under this PR it stays flat at ~0.005 ms — linear.

## The flag ordering is load-bearing, and it is correct

The PR calls out that clearing `publish_scheduled` **before** reading the snapshot is the one thing standing between this design and a swallowed update. I checked it by counterexample, deterministically:

```
PROBE wrong order   -> order-2 published? false          (flag cleared AFTER the send)
PROBE shipped order -> order-2 published? true (carrying 2)
```

With the flag cleared after the send, an update landing in that gap is never published **and arms no new window** — on a quiet book it would stay invisible indefinitely. With the shipped ordering it goes out. I also walked the other interleavings (update between the clear and the read; update blocked on the write lock): both publish, at worst with one redundant emission. The comment in the code says exactly what the code does.

## Other checks

- **The four remaining immediate publishers are the right ones**: `:946` (first daemon reply, calculated sats), `:1549` (local→daemon id reconciliation), `:1948` (kind14/AddInvoice, where the screen reads sats from the book) and `:2506` (inside `subscribe_single_order`, i.e. the user's own orders). All are latency-sensitive and fire once per action — nothing to coalesce.
- **CI on the tree merged with current `main`**: `cargo test --locked` → **333 passed, 0 failed**; `cargo clippy --locked -- -D warnings` → clean; **`cargo check --locked --target wasm32-unknown-unknown` → clean**, which was the real risk here given the added `spawn` + timer. Both go through `crate::rt::spawn` and `crate::rt::time::sleep`, i.e. the correct shims (`wasm_bindgen_futures` / `wasmtimer` on web).
- Confirmed this is genuinely the cold-start path: pool Online → `nostr.rs:58` → `subscribe_orders()` → the live loop → `ingest_order_event`. #360's refetch is not involved in startup.

## Minor (none blocking)

- **The doc-comment placement from #360 survives, and this PR edits that exact line.** `enum Publish` still sits between `ingest_order_event`'s doc block — *"Parse a Kind 38383 event and upsert it into the order book … Shared by the live subscription loop and the node-switch refetch"* — and the function, so the ingest functions are undocumented and that text describes the enum. Natural place to fix it.
- **One of the five "still immediate" citations is dead code.** `orders.rs:3566` is `process_order_event`, marked `#[allow(dead_code)]` and documented as *"Currently unused"*. The other four are real (listed above).
- **The new seam is untested at the call site** (same shape as #360): all three tests drive `OrderBook` directly, none asserts which mode `ingest_order_event` picks. The enum now has two variants and choosing `WhenBatchEnds` on the live path would mean mutations **nobody ever publishes** — a frozen book, with every test still green.
- **The two new tests sleep 800 ms each** (`PUBLISH_COALESCE_MS * 4`), adding ~2.4 s of real time to the suite. `tokio::time::pause()` would make them instant, but `crate::rt::time::sleep` is `wasmtimer` on web, so I understand the choice — just noting it.
- **Behaviour note worth recording:** first paint on cold start moves to +200 ms but arrives settled. Better, and measured above — but someone tracking "time to first order" should know it changed.

Issue #326, opened 2026-08-27
Trade-key index for a taken range order is corrupted by the content-fingerprint restore -> release/cancel/rate signed with the wrong key (InvalidPeer)


Forte11Cuba:
### Summary
Post-take actions (`release`, `cancel`, `rate`) can be signed with the **wrong trade key** and rejected by the daemon with `InvalidPeer` / `InvalidOrderStatus`. The trade gets stuck (e.g. stays at *fiat-sent*, never releases; the rating never lands).

Root cause: the daemon-UUID → `trade_index` mapping these actions rely on is derived from a **content fingerprint** that is neither unique per order nor stable across an order's lifecycle, so it gets overwritten with another order's index.

### Root cause
Two persistence stores disagree for the same order:
- `trades.trade_key_index` — set at creation — **correct**.
- `trade_keys.key_index` — the table `get_trade_key_index()` reads for `release`/`cancel`/`rate` — **wrong**.

The `trade_keys` mapping for a daemon UUID is (re)established only by the Kind 38383 fingerprint-restore path in `rust/src/api/orders.rs` (`store_trade_key_index(&info.id, get_trade_key_index(&content_key))`). `order_content_key` (`rust/src/mostro/pending.rs:378`) gives **absolute precedence to `fiat_amount`** (`Some(a) → f{a}`), so:

1. **Range orders lose their range on the wire once taken.** A range order (`min..max`) is published as a plain fixed-amount order (`fa=<taken amount>`, no min/max tags) once in-progress. The restore path re-parses it as a *fixed* order and computes a *different* fingerprint than at creation — which can match an **unrelated fixed order** and copy its index. **(This is the mechanism observed in the logs below.)**
2. **Two identical open orders** share one fingerprint slot; the last write wins. (Secondary path — same fix closes it.)

`create_order` already correlates the daemon's `new-order` reply via `request_id` and obtains the real `daemon_id`, but it **never persists** `daemon_id → trade_index` into `trade_keys`, leaving that mapping entirely to the fragile fingerprint path.

### Reproduction (observed, with evidence)
- Order `6ce58aa2` = **range 500–5000 CUP EnZona**, seller trade key = index **16** (`250fa732…`).
  → `app/log.txt:218` `NewOrder … trade_index=Some(16) trade_pubkey=250fa732`; `trades.trade_key_index = 16` ✅
[...]

PR #362, opened 2026-08-31
perf(ingest): cache trade-key misses so strangers' orders skip the DB

grunch:
Phase 2, PR 2.3 of the optimization plan (#348). Independent of #360/#361.

## Problem

`ingest_order_event` computes a content fingerprint for every Kind 38383 event and looks it up to decide whether the order is ours (`orders.rs:3137`). For every order belonging to somebody else — which is nearly all of them — that lookup misses the in-memory cache and falls through to storage (`lookup_trade_key_index`, `orders.rs:79-100`).

So the busiest path in the app performs **one storage round trip per relay event**, on ids that are never going to be present. On web that is an IndexedDB transaction. The existing cache cannot help: it only remembers hits.

## Change

Misses are remembered too.

This is only safe because absence here is **stable**: `store_trade_key_index` is the single path from absent to present anywhere in the crate, and it now clears the entry. That is the whole correctness argument, so it is written at both ends — at the set's definition and at the invalidation point.

Two deliberate choices worth reviewing:

- **A failed read is never cached.** Only `Ok(None)` records a miss; `Err` does not. An error is not evidence of absence, and caching it would strand a genuinely-ours order as "not ours" — after which every action on it gets signed with the wrong trade key and the daemon rejects it (the `InvalidPeer` class of failure, cf. #326).
- **The set drops everything when full**, rather than evicting oldest-first. It is a cache: an eviction costs one extra storage read. Tracking insertion order for entries nobody will ask about twice costs more than it saves.

I checked the other paths that change key state — the publish-failure rollback (`orders.rs:524`), the local→daemon id rebind (`orders.rs:1490-1495`), and `clear_trade_keys` (`identity.rs:304`). All of them move ids *towards* absent, so a remembered miss stays correct; none needs invalidation.

## Scope

The plan also suggested memoizing `local_trade_status` per ingest pass. Left out: that lookup goes through `get_trade_by_order_id`, which #350 gives a real index, so the remaining cost is a keyed read rather than a full scan. Worth re-measuring after #350 lands rather than adding pass-scoped state on spec.

## Test plan

- [x] New test: storing a key clears its recorded miss — the failure this guards against is silent and expensive, so it gets its own test rather than riding on the lookup path (where the positive cache would mask it).
- [x] New test: the miss set stays within its ceiling. Written against the pure `record_miss` helper rather than the global, so it cannot flake against tests running in parallel.
- [x] `cargo test` — 295 passed, 0 failed
- [x] `./scripts/frb-generate.sh --check` — clean, no bridge surface change


Catrya (dismissed):
Reviewed at `f99246d3`, merged locally against current `main` (`6261924`). The branch is 37 commits behind and **conflicts** — see the note at the end; after a trivial resolution: **331 Rust tests** pass, `cargo clippy --locked -- -D warnings` and `cargo check --locked --target wasm32-unknown-unknown` are clean.

**Approving.** The safety argument holds — I traced every path rather than taking it on trust — and I measured the benefit, which turns out to be much smaller than the description implies. That is worth fixing in the body, not a reason to hold the PR.

## The correctness claim checks out on every path

The whole argument rests on *"`store_trade_key_index` is the sole path from absent to present"*. I walked every writer:

| Path | Effect | Invalidates? |
|---|---|---|
| `create_order` (`:496-497`) — stores under local UUID and `ck` | absent→present | ✓ via `store_trade_key_index` |
| `take_order` (`:897`) | absent→present | ✓ |
| `bridge_fingerprint_trade_index` (`:3175-3176`) | lookup (records miss) → store | ✓ and in the right order |
| local→daemon rebind (`:1518-1524`) | `delrect** `save_trade_key(did)`,`map.remove(local)`, then `store_trade_key_index(&did, …)` | ✓ — the direct write is immediately followed by the store, which clears `did`'s miss |
| publish-failure rollback (`:553`) | remov | not needed (moves towards absent) |
| `clear_trade_keys` (`identity.rs:304`) |

The rebind was the one I expected to break it: the daemon id is *published in the 38383 event*, so it can have been
recorded as a miss before the rebind makes se line 1524 goes through`store_trade_key_index` even though line 1520 already wrote the DB directly.

Two more properties that hold and are worth stating in the code, since they are load-bearing:

- **No miss is recorded when there is no DB.** If `crate::db::app_db::db()` is `None` (early startup, or web before
IndexedDB opens) the whole `match` is skippcannot manufacture false negatives.
- **No `std::sync::RwLock` guard crosses an `await`.** The read guard lives inside the `is_ok_and(...)` temporary; the write guards live inside their `if let` blocks. Same shape as the existing `trade_key_map` access.

And the central decision — **never cache `Eued: a failed read is not evidence ofabsence, and caching it would strand a genuinely-ours order as somebody else's.

## Measurements

**What an avoided lookup costs.** Real `SqliteStorage`, 50 owned trade keys, 2000 lookups, `--release`:

| | per lookup |
|---|---|
| `get_trade_key` that misses (hits the DB) | **38,016 ns** (~38 µs) |
| `HashSet::contains` (the negative cache) | **327 ns** |
| **saved per avoided lookup** | **~37.7 µs

**How many are actually avoided.** I instrux` with counters and ran the app against alive daemon:

| lookup outcome | count |
|---|---|
| positive-cache hit | 171 |
| **DB read** | **98** |
| **negative-cache hit** | **3** |
| total | 272 |

So over a startup the negative cache avoideads (~3%), about **0.11 ms**. That isstructural, not bad luck: the cache only helps on the *second* lookup of the same absent key, and in the startup
burst each fingerprint is looked up exactlyand nobody asks again.

Where it does pay is `refetch_active_node_o switch), which re-ingests the whole book:every fingerprint is looked up again and all of those become negative-cache hits. I could not put a number on that —
my order book has 2 orders, so the list nevfresh does not fire.

**Which is why I would reword the Problem strip per relay event, on ids that are nevergoing to be present" describes the *cost* correctly but implies the cache removes it; it removes the repeats. Something like "removes the repeated lookup of the same absent fingerprint — which is what every book refetch doeswould match what it actually buys.
                                                                                                                  ## Notes
                                                                                                                  1. **Needs a rebase.** It conflicts with to/orders.rs`, and it is purely a test-modulecollision: #352 landed `daemon_message_dedup_recognizes_repeats_and_evicts_oldest` at the same spot. Keeping both blocks resolves it; I verified the result is green (331 tests, clippy, wasm).                                     2. **The branch the PR argues hardest for i.** `Err` → do not cache is the centralsafety decision, and nothing pins it: collate_trade_key_miss(...)` would leave thesuite green. I accept it is hard to reach — `OnceCell` with no way to inject a failingstore — but it is worth saying so in the PR rather than leaving the gap silent.                                   3. **The 4096 ceiling with drop-everything ooks at scale.** The key is the contentfingerprint, which includes the exact fiat s one distinct fingerprint per order. At theplan's own M4 target (10k+ orders) the set p, in exactly the scenario it exists for.Fine at 3k. Worth tying the constant to tha unexplained.

4. **On web the per-lookup saving is larger rather than 38 µs of SQLite — so the same3% is worth more there. Not measured; that needs the web build.

5. Good call on `the_miss_cache_stays_bounded` driving the pure `record_miss` against a local `HashSet` instead of the global.

PR #366, opened 2026-08-31
perf(ui): look orders up by id instead of scanning the whole book

grunch:
Phase 2, PR 2.7 of the optimization plan (#348). Independent of the other Phase 2 PRs.

## Problem

Four screens need exactly one order, and each found it by watching the **entire** order book and walking it:

```dart
final orders = ref.watch(orderBookProvider).valueOrNull ?? [];
final order = orders.where((o) => o.id == widget.orderId).firstOrNull;
```

- `take_order_screen.dart:210`
- `my_order_screen.dart:98`
- `trade_detail_screen.dart:542`
- `trade_state_header.dart:34` (the chat header)

So each screen paid an O(orders) scan **and a full rebuild** on every relay event, regardless of what changed — while displaying a single order that had usually not moved at all. On a book of a few thousand orders, opening a chat means re-scanning it every time anything anywhere updates.

## Change

- `orderBookIndexProvider` — the book keyed by id, built once per emission.
- `orderByIdProvider` — a family that `select`s one entry out of that index.

Screens now watch only their own order.

### The part that makes it work

`select` compares the selected value to decide whether to notify. `OrderItem` had no `==`, so comparison fell back to identity — and since every book emission rebuilds every `OrderItem`, identity always reports a change and `select` would filter nothing.

So this PR adds value equality to `OrderItem`. I deliberately left that out of #357, where it would have been speculative; here it is load-bearing, and the test pins exactly what it buys.

## Test plan

- [x] New test: resolves an order by id, and returns null for an unknown one.
- [x] New test — the one that matters: a screen watching order `a` is **not** rebuilt when order `b` changes, and **is** rebuilt when `a` itself changes. Without value equality on `OrderItem` the first assertion fails, so this test also guards the equality implementation against someone trimming it later.
- [x] `flutter test` — 259 passed, 0 failed (`main`: 257; this PR adds 2)
- [x] `flutter analyze` — 7 issues, identical to `main` (pre-existing `info` deprecations)
- [ ] Manual: open a chat and a trade detail while the book is busy, confirm the values still track the order

### DIFF
--- a/lib/features/chat/widgets/trade_state_header.dart
+++ b/lib/features/chat/widgets/trade_state_header.dart
@@ -31,8 +31,7 @@ final chatTradeOrderProvider =
   // Re-resolve when the polled status changes (e.g. active → fiatSent).
   ref.watch(tradeStatusProvider(orderId));
 
-  final book = ref.watch(orderBookProvider).valueOrNull;
-  final fromBook = book?.where((o) => o.id == orderId).firstOrNull;
+  final fromBook = ref.watch(orderByIdProvider(orderId));
   if (fromBook != null) return fromBook;
 
   try {
--- a/lib/features/home/providers/home_order_providers.dart
+++ b/lib/features/home/providers/home_order_providers.dart
@@ -108,6 +108,53 @@ class OrderItem {
     return v == v.truncateToDouble() ? v.toInt().toString() : v.toString();
   }
 
+  /// Value equality, so a screen watching one order can tell whether *its*
+  /// order actually moved. Every book emission rebuilds every [OrderItem],
+  /// so identity comparison always reports a change and
+  /// [orderByIdProvider]'s `select` would never filter anything out.
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      other is OrderItem &&
+          other.id == id &&
+          other.kind == kind &&
+          other.fiatAmount == fiatAmount &&
+          other.fiatAmountMin == fiatAmountMin &&
+          other.fiatAmountMax == fiatAmountMax &&
+          other.fiatCode == fiatCode &&
+          other.paymentMethod == paymentMethod &&
+          other.premium == premium &&
+          other.creatorPubkey == creatorPubkey &&
+          other.createdAt == createdAt &&
+          other.expiresAt == expiresAt &&
+          other.rating == rating &&
+          other.tradeCount == tradeCount &&
+          other.daysActive == daysActive &&
+          other.status == status &&
+          other.amountSats == amountSats &&
+          other.isMine == isMine;
+
+  @override
+  int get hashCode => Object.hashAll([
+        id,
+        kind,
+        fiatAmount,
+        fiatAmountMin,
+        fiatAmountMax,
+        fiatCode,
+        paymentMethod,
+        premium,
+        creatorPubkey,
+        createdAt,
+        expiresAt,
+        rating,
+        tradeCount,
+        daysActive,
+        status,
+        amountSats,
+        isMine,
+      ]);
+
   /// Map a Rust-bridge [OrderInfo] to an [OrderItem] for display.
   factory OrderItem.fromInfo(OrderInfo info) => OrderItem(
         id: info.id,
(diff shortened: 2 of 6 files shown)


Catrya (approved):
Reviewed at `96b78226`, merged locally against current `main` (`6261924`, which already includes #372). The branch is 37 commits behind and **merges clean** — including #372's rewrite of `take_order_screen.dart`, which auto-merges correctly. On that merge: `flutter analyze` clean, `flutter test` **295 passed, 0 failed** (293 on `main` + the 2 new ones).

**Approving.** No regressions, and the win is large and measurable. I have one measured suggestion about the mechanism, but it does not block.

## What I checked

**The value equality is complete.** `OrderItem` has exactly 17 fields (`id, kind, fiatAmount, fiatAmountMin, fiatAmountMax, fiatCode, paymentMethod, premium, creatorPubkey, createdAt, expiresAt, rating, tradeCount, daysActive, status, amountSats, isMine`) and both `==` and `hashCode` cover **all** of them. A field missing there is the classic version of this bug — the screen silently keeps stale values — and there isn't one.

**The test that matters is not vacuous.** I checked by mutation: removing `operator ==` from `OrderItem` makes `does not rebuild a screen whose order did not change` fail with `Expected: <1> / Actual: <2>` — exactly the "select filters nothing" failure the description predicts. So that test really does guard the equality implementation against a later trim.

**No behavioural change at the edges.** Before: `valueOrNull ?? []` + `.where().firstOrNull` ⇒ `null` while loading. Now: empty index ⇒ `null`. Same. And `orderByIdProvider` does notify on the `null → Some` transition (the order entering the book), which is what `chatTradeOrderProvider` depends on.

## Measurements

Deterministic harness: a 2000-order book, 100 emissions in which **only somebody else's order moves**, and a screen watching a different order that never changes.

| | |
|---|---|
| screen rebuilds on `main` | **100** |
| screen rebuilds with this PR | **1** |
| scan cost removed | 16.6 µs per emission per screen |
| **cost this PR adds**: building the index | **336 µs per emission**, shared across screens |

Two readings:

- **The real benefit is large and is exactly what the PR is after**: 99 of every 100 rebuilds disappear. And those are `TradeDetailScreen` rebuilds (a 1332-line widget tree), not a `Text` — the 16.6 µs scan figure badly understates what was actually being paid. The saving also **does not depend on book size**: with 2 orders in the book it is still 100 → 1.
- **But the chosen mechanism is not the cheap one.** Materializing the `Map` costs ~20x the scan it replaces (336 µs vs 16.6 µs). With one screen open this PR spends *more* CPU per emission than `main`; it would only pay for itself with ~20 order-watching screens mounted at once, which never happens. The net win comes entirely from the avoided rebuilds, not from the index.

There is an alternative that gives **the same filtering without the index** — verified in the same harness:

```dart
final orderByIdProvider = Provider.autoDispose.family<OrderItem?, String>((ref, id) =>
    ref.watch(orderBookProvider.select(
      (a) => a.valueOrNull?.where((o) => o.id == id).firstOrNull,
    )));
```

```
screen rebuilds  old=100   new(index)=1   alt(select-on-list)=1
```

Same result, 16.6 µs instead of 336 µs, and `orderBookIndexProvider` goes away. I am not blocking on this — the PR is already a big improvement over `main` — but if the index stays, its docstring should say it exists for the many-watchers case, because today it is the expensive half.

## Minor

1. **The value equality is fragile against a new field.** If someone adds a field to `OrderItem` and does not touch `==`, screens will silently keep the old value — the same failure the test prevents, arriving by another route. A comment on the field list ("keep `==`/`hashCode` in sync") helps; `freezed` is already in dev-dependencies if this is ever worth solving properly.
2. **The manual check is still unticked** (*"open a chat and a trade detail while the book is busy"*). With a 2-order book it would not show much anyway; the harness above covers it better than a look by hand.
3. The numbers in the description are stale (259 tests, "7 issues" from analyze); against today's `main` it is **295 tests** and **0 analyze issues**.

## What I could not verify

- **The real cost of one `TradeDetailScreen` rebuild**, which is what turns "99 rebuilds avoided" into milliseconds. I measured provider notifications, not widget-tree build time. That needs a `flutter test --profile` timeline — plan item 5.2, which does not exist yet.
- **Behaviour with a genuinely busy book**: the test node has 2 pending orders, so the busy-book scenario is synthetic here.

PR #368, opened 2026-08-31
fix(storage): key notifications by id and commit bulk updates once

grunch:
Phase 2, PR 2.9 of the optimization plan (#348). Independent of the other Phase 2 PRs.

## Problem

The notification store used `intMapStoreFactory` — auto-incrementing integer keys — while every operation actually addresses records by **notification id**. So each one had to search:

```dart
final finder = Finder(filter: Filter.equals('id', notification.id));
final existing = await _store.findFirst(client, finder: finder);
```

A full-store scan per save (`:80-89`) and per delete (`:110`).

`markAllAsRead` (`:242-250`) compounded it: one independent, **un-awaited** `save()` per notification, each doing its own scan and committing its own transaction. Marking N notifications read cost N scans and N commits, all on the UI isolate.

## Change

- Records are keyed by notification id — `put`/`delete` by key, no scan.
- New `saveAll` commits a batch in **one** transaction; `markAllAsRead` uses it.

## The migration, and two traps in it

The keyed records live under a **new store name** (`notifications_v2`) rather than a re-typed view of the old one. Both alternatives I tried first were wrong, and both failed loudly in tests:

1. Reading an int-keyed record through a `StoreRef<String, …>` throws `type 'int' is not a subtype of type 'String'`. Without a migration, an upgrading user either loses their notification history or crashes on open.
2. Pointing the old and new `StoreRef`s at the **same** physical store makes the migration's cleanup (`_legacyStore.delete(txn)`) wipe the records it just wrote — the store name is the same, only the Dart type differs.

`_migrateLegacyRecords` runs once per database inside a transaction, re-keys existing records, and clears the old store. From the second launch it is a no-op (an empty `find`).

The processed-events tombstone ledger is untouched — it was already correctly keyed by String.

## Test plan

- [x] New test: saving the same id twice leaves one record (the upsert contract that the `Finder` used to provide).
- [x] New test: deleting by id removes only that record.
- [x] New test: `saveAll` marks a whole batch read.
- [x] New test: **records written by the old int-keyed store survive the upgrade.** This is the one that matters — it caught both traps above, and it is the difference between a perf fix and silent data loss for every existing user.
- [x] `flutter test` — 261 passed, 0 failed (`main`: 257; this PR adds 4)
- [x] `flutter analyze` — 7 issues, identical to `main`
- [ ] Manual: upgrade over an existing install with notifications present, confirm they are still listed and still marked read correctly

### DIFF
--- a/lib/features/notifications/providers/notifications_provider.dart
+++ b/lib/features/notifications/providers/notifications_provider.dart
@@ -22,7 +22,13 @@ class SembastNotificationsStore {
         _pathOverride = path;
 
   static const _dbName = 'notifications.db';
-  static const _storeName = 'notifications';
+  /// The int-keyed store this feature shipped with.
+  static const _legacyStoreName = 'notifications';
+
+  /// Records keyed by notification id. A separate store name, not a re-typed
+  /// view of the old one: the two must not share physical records, or
+  /// migrating out of the old shape would delete what it just wrote.
+  static const _storeName = 'notifications_v2';
 
   /// Test seam: when set, bypasses platform factory/path resolution (e.g. an
   /// in-memory Sembast factory for restart/replay tests).
@@ -31,7 +37,14 @@ class SembastNotificationsStore {
 
   Database? _db;
   Completer<Database>? _opening;
-  final _store = intMapStoreFactory.store(_storeName);
+  /// Keyed by notification id. Before this, the store used auto-incrementing
+  /// integer keys and every write looked its record up with a `Finder` — a
+  /// full-store scan per save, and O(n) scans for an O(n) bulk update.
+  final _store = StoreRef<String, Map<String, Object?>>(_storeName);
+
+  /// The int-keyed shape this store used to have. Only read, and only once
+  /// per database, by [_migrateLegacyRecords].
+  final _legacyStore = intMapStoreFactory.store(_legacyStoreName);
 
   /// Tombstone / processed-event ledger: ids of externally-sourced events that
   /// have already been handled. It survives deletion of the notification record,
@@ -54,6 +67,7 @@ class SembastNotificationsStore {
         final dir = await appDataDirPath();
         db = await databaseFactoryIo.openDatabase(p.join(dir, _dbName));
       }
+      await _migrateLegacyRecords(db);
       _db = db;
       _opening!.complete(db);
       return db;
@@ -62,6 +76,27 @@ class SembastNotificationsStore {
       _opening = null;
       rethrow;
     }
+  }
+
+  /// Re-key records written by the previous int-keyed store.
+  ///
+  /// Reading an int-keyed record through a `StoreRef<String, ...>` throws, so
+  /// without this an upgrade would either lose the user's notification history
+  /// or fail on open. Runs inside one transaction and clears the old records,
+  /// so it is a no-op from the second launch onwards.
+  Future<void> _migrateLegacyRecords(Database db) async {
+    final legacy = await _legacyStore.find(db);
+    if (legacy.isEmpty) return;
+    await db.transaction((txn) async {
+      for (final record in legacy) {
+        final value = Map<String, Object?>.from(record.value);
+        final id = value['id'];
+        if (id is String) {
+          await _store.record(id).put(txn, value);
+        }
+      }
+      await _legacyStore.delete(txn);
+    });
   }
 
   Future<List<NotificationModel>> loadAll() async {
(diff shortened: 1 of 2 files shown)


Catrya (approved):
Reviewed at `33afdf8d`, merged locally against current `main` (`6261924`). The branch is 37 commits behind and **merges clean**. On that merge: `flutter analyze` clean, `flutter test` **297 passed, 0 failed** (293 on `main` + the 4 new ones).

**Approving.** The risky part here was the migration — the plan warned it is the difference between a perf fix and silently losing every user's notification history — and it is correct. I verified it beyond the unit tests: on a real on-disk database and in the running app.

## The migration

- **The migration test is not decorative.** Verified by mutation: disabling the `_migrateLegacyRecords(db)` call makes `adopts records left by the previously int-keyed store` fail with `Expected: ['legacy-1'] / Actual: []`.
- One correction to the description: it says that without a migration a user *"either loses their notification history or crashes on open"*. With the store rename this PR introduces, what I observed is the **silent loss** — `loadAll()` returns empty, no error. The crash belonged to the other alternative (re-typing the same store name), which this PR correctly rejected. The migration is needed either way, but the failure mode it prevents is the quiet one, which is the worse of the two.
- **Atomicity**: the re-key and the legacy wipe run in one `db.transaction`, so an interrupted upgrade cannot leave half-migrated or duplicated records.
- **Idempotence**: `_legacyStore.find(db)` returns empty from the second launch and returns before opening a transaction.
- **Startup ordering**: it runs inside `_open()` before `_db` is assigned and before the `_opening` completer completes, so no concurrent caller can observe a half-migrated store.
- **No orphaned consumers**: there is no other reference to `'notifications'`, `intMapStoreFactory` or `notifications.db` anywhere in `lib/` or `test/` outside this file.
- **The ordering risk does not exist**: `find` over String keys returns in key order rather than insertion order, but the consumer (`notifications_provider.dart:181-182`) sorts by `timestamp` descending after loading, so the list the user sees is unchanged.

## Verified on disk and in the app, not just in memory

The tests use `databaseFactoryMemoryFs`, which is right for CI but does not exercise the production factory. I ran the real upgrade path:

1. With **`main`'s** code, wrote 3 notifications into a scratch file through `databaseFactoryIo`. Raw keys came out as integers, as the old store produces:
   ```
   {"key":1  {"key":2  {"key":3
   ```
2. Opened **the same file** with this PR's code:
   ```
   UPGRADE recovered=3 ids=[vieja-1, vieja-2, vieja-3]
   UPGRADE titles=[Notificación previa 1, Notificación previa 2, Notificación previa 3]
   UPGRADE second open=3
   ```
   and the keys in the file became strings:
   ```
   {"key":"vieja-1"  {"key":"vieja-2"  {"key":"vieja-3"
   ```

Then the visual half. I backed up the real `notifications.db`, seeded it with three notifications written by `main`'s int-keyed code, and launched the app with this PR:

- The notifications screen listed all three intact, unread, in the right order — records written by the old store, read by the new one, nothing lost.
- **Marcar todo como leído** (the `saveAll` half) cleared the unread markers, and the file afterwards held exactly:
  ```
  {"key":"prueba-migracion-1" ... "isRead":true}}
  {"key":"prueba-migracion-2" ... "isRead":true}}
  {"key":"prueba-migracion-3" ... "isRead":true}}
  ```
  Keyed by id, read state persisted through the single-transaction batch.

The original database was restored afterwards. So the unticked manual item in the test plan — *"upgrade over an existing install with notifications present"* — is covered, with the caveat below.

## Measurements

Sembast memory-fs, 200 notifications, `main` vs this PR:

| | `main` | this PR |
|---|---|---|
| persisting `markAllAsRead` for 200 notifications | **166 ms** (832 µs each) | **41 ms** (208 µs) |
| a single `save` into a store of 200 | **1,700 µs** | **813 µs** |

~4x on the bulk path, ~2x on a single save. Debug-VM numbers, so the absolutes do not transfer to a device — but they size the problem being fixed: **166 ms of work on the UI isolate** for one "mark all read" tap is roughly ten dropped frames, and it was fire-and-forget, so failures were swallowed.

## Minor

1. **Real bug: the `$e` is escaped.** In `markAllAsRead` (`notifications_provider.dart:291`):
   ```dart
   debugPrint('NotificationsNotifier: failed to persist markAllAsRead: \$e');
   ```
   In Dart `'\$e'` prints the literal `$e`, not the exception; the previous version interpolated. Since the write is still fire-and-forget, that `debugPrint` is the **only** signal that persistence failed — and it now says nothing useful. Drop the backslash.
`NotificationModel.id` is a non-nullable `String` (`notification_model.dart:34`), so every legacy record carries one. But a record that failed the check would be skipped and then removed by the unconditional `_legacyStore.delete(txn)` — silently. An `else` that logs costs one line and turns a silent loss into a warning.

3. **Downgrade path**: anyone installing an older build after upgrading will read `notifications` (now empty) and see their history gone. Not something I would block a beta on, but worth stating in the PR.

4. **Small leak on a failed open**: if `_migrateLegacyRecords` throws, the already-opened `Database` is never assigned to `_db` nor closed, and the next `_open()` opens another. An `await db.close()` in the `catch` handles it.

5. The numbers in the description are stale (261 tests, "7 issues" from analyze); against today's `main` it is **297 tests** and **0 analyze issues**.

PR #382, opened 2026-09-03
Fix/381 session fallback

### LINKED ISSUE #381: Peer chat: post-restart session rebuild depends on relay replay session-dependent functions should fall back to the trade row
## Context

Follow-up to #334 (fixed in #345). That fix persists `counterparty_pubkey` on the trade row (the durable peer record) and creates/updates the in-memory session at peer reveal, for both roles. Item 2 of #334's proposed fix having the session-dependent functions derive from the trade row when no session exists  was intentionally left out of that PR's scope. This issue tracks it.

## Current behavior after a restart

Sessions are memory-only, so a restart empties them. Today they are rebuilt by **relay replay**: the global DM subscription carries no
`since`, so the whole daemon-message history is re-dispatched on startup and the first replayed reveal re-runs the full capture (durable write + session + chat subscription).

Incoming chat is additionally covered without a session: `resubscribe_active_chats` re-derives the chat keys straight from the row
(`get_active_trade_keys(trade.trade_key_index)` + `trade.counterparty_pubkey`) and re-arms the listener.

## The gap

**Sending** (`send_message`, `send_file`, `download_attachment`) still requires a session. Between startup and the arrival of the replayed reveal or permanently, if every relay in the pool has pruned the trade's daemon messages the maker and taker of a live trade can receive but not send: `send_message` takes its local-only branch (stores, returns `Ok`, never publishes) and `send_file` / `download_attachment` fail with `SessionNotFound`.

This is the same silent failure mode that made #334 hard to detect: the message appears in the sender's chat, the peer never receives it, no error anywhere.

Everything needed to serve those calls is already on the row: the trade key index and the counterparty pubkey. mostrix works this way by
construction no session concept; chat state is derived on demand fromthe persisted order.

## Proposed fix

When a session-dependent chat function finds no session, derive it from the trade row exactly as `resubscribe_active_chats` already does, create the session from the result (populating the cache), and proceed. The session stays a pure cache which is what the rest of the codebase already assumes.

Not in scope: web, where there is no trades store yet (#233) and the in-memory session remains the only chat-identity store.

Forte11Cuba:
Close #381

**Depends on #345** — merge after it; the diff collapses to this PR's own commits once it lands.

A restart empties the memory-only sessions, leaving the send path broken until a relay replays the peer reveal — or permanently, if the history was pruned. `send_message`, `send_file` and `download_attachment` now rebuild the session from the persisted trade row (index + counterparty, ECDH re-derived, session re-cached), gated by the same liveness/poison guard as the startup resubscription so pre-#334 rows never resurrect. The session is inserted fully populated in one write-lock pass, so no concurrent send can observe a keyless intermediate. Web (#233) is unchanged: no trades store, replay-only.

Covered by unit tests with injected keys, plus a mutation-validated `#[ignore]`d seam test driving `send_message` end to end from a row with no session — including a poisoned-row leg that fails if the safety gate is removed.

## Summary by CodeRabbit

- **New Features**
  - Chat sessions can now be reconstructed from saved trade information when unavailable in memory.
  - Peer details are captured and retained consistently across trade messages, enabling more reliable chat setup.
  - Messages and attachments can continue using recovered sessions when possible.

- **Bug Fixes**
  - Improved handling of invalid or incomplete peer information and interrupted session recovery.
  - Prevented incorrect counterparty details from being stored.
  - Ensured recovered chat sessions are created atomically for more reliable messaging.

- **Documentation**
  - Updated messaging and trade contracts to describe session recovery, peer discovery, and related error conditions.


Catrya (changes requested):
**Changes requested.** The fix is real — I proved it end to end against a live relay — but the branch does not compile against `main`, its diff is two PRs, and the one line that keeps it safe is held by nothing.

  ## First: this PR is #345 + #381, not #381

  Five of its seven commits are **PR #345's, with identical SHAs** — `0d2799f`, `cdcc690`, `ea50c8e`, `de3d6b9`, `8f11528`. #345 is still open and unmerged, and because #382's base is `main` rather than #345's branch, GitHub shows the union:

  ```
  #382 as listed:        1117 / -87   across 10 files
  #382's own delta:       374 / -16   across  4 files  (b6a3041 + 42311e5)
  ```

  Merging #382 merges #345 wholesale under a title about a different issue, and leaves #345 as a zombie. Either rebase this onto `fix/334-persist-counterparty` so the PR shows only its own two commits, or land #345 first — after which this diff shrinks on its own. Please don't merge #382 while #345 is unreviewed.

  Everything below is about the two commits that are actually #382's.

  ## The fix works — measured, not argued

  I ran a probe against `ws://localhost:7000` that reproduces the exact restart shape (row persisted, session gone), 
: it silently bundles unmerged PR #345, doesn't compile against main, and its safety guard has no test coverage. Next step is writing the English review message for you to paste.

❯
## What I verified

- The branch alone: `cargo test --locked` → 300 passed.
- The merge onto `16266d9`: 1 conflict + 26 path errors; after fixing, 368 passed / clippy / wasm all clean.
- Both `#[ignore]`d tests pass together under `--ignored`, as the PR claims. The other 8 failures in that batch are pre-existing — `main` fails the same 8.
- Five mutations (three call sites, the gate, atomicity), results above.
- Two live probes against `ws://localhost:7000` with the counterfactual for each.
- Web is genuinely unchanged: `IndexedDbStorage::get_trade_by_order_id` returns `Ok(None)`, so `session_or_rebuild` short-circuits before any derivation.
- 29 rows of a real trades table against the eligibility gate.calls `send_message`, then **reads the relay as the peer** and decrypts with the peer's own `K_conv`:

  | | `sender_pubkey` | kind-14 events on the relay | peer decrypts |
  |---|---|---|---|
  | **with the fallback** | our trade key | 1 | ✅ the message |
  | **without it** (`get_session` only) | `""` | **0** | ✗ |

  That is issue #381 exactly: `Ok(...)` returned, message shown locally, nothing on the wire. The fallback closes it.

  ## Blocking 1 — the safety gate is held by nothing

  `chat_still_relevant` inside `session_or_rebuild` is what stops a poisoned pre-#334 row from being resurrected. I deleted it:

  ```
  cargo test --locked        → 368 passed, 0 failed     ← unchanged
  cargo test -- --ignored    → send_message_rebuilds… ok ←

Catrya (dismissed):
tACK

PR #388, opened 2026-09-04
fix(disputes): persist the origin marker on a late-accepted open

grunch:
Follow-up to #256, which merged while this fix was in flight (CodeRabbit's last round on that PR; the reply there cites a commit that never reached `main`).

## Problem

`record_late_acceptance` — the path for an `open_dispute` that timed out and was accepted by the daemon later — builds the record with `initiated_by_me: true` but never wrote `dispute_mine:<order_id>`. The on-time path does. So after a restart, rehydration restored that dispute as **peer-opened**, which is exactly the wrong description #256 persisted the origin to avoid.

## Change

`record_late_acceptance` calls `persist_dispute_origin` on its success path, same as `open_dispute`. Best-effort like every other persistence write.

`a_late_acceptance_is_reconciled_into_a_record` now initialises the process-wide store (any `SqliteStorage` serves, first caller wins — same pattern as the rehydration tests) and reads the marker back, then clears its own keys.

## Test plan

- [x] `cargo test --locked` — 385 passed
- [x] `cargo clippy --locked -- -D warnings` — clean

https://claude.ai/code/session_01MV8X5ZDur4kqUEG1s6REiE

## Summary by CodeRabbit

* **Bug Fixes**
  * Late-accepted disputes now retain their correct origin after a restart.
  * Dispute state is rehydrated consistently as initiator-opened rather than peer-opened.

### DIFF
--- a/rust/src/api/disputes.rs
+++ b/rust/src/api/disputes.rs
@@ -495,9 +495,13 @@ pub(crate) async fn record_late_acceptance(trade_id: &str, dispute_id: Option<St
         .await;
 
     match result {
-        Ok(()) => log::info!(
-            "[disputes] reconciled late daemon acceptance for trade={trade_id}"
-        ),
+        Ok(()) => {
+            // Same as the on-time path: this side opened it, and the replay
+            // can never restore that, so the marker must be written here too
+            // or a restart would rehydrate the record as peer-opened.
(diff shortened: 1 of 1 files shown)


Catrya (approved):
tACK

Issue #395, opened 2026-09-05
A trades row carries two identities: the primary key and data.order.id diverge


Catrya:
`trades.id` (the primary key) and `data.order.id` (inside the serialised
`TradeInfo`) are not the same value. On a real database, **10 of 36 rows**
differ — all of them taker rows:

```
PK=77c5bea1  data.order.id=97c10465  idx=57  role=Buyer   status=WaitingBuyerInvoice
PK=a3a29d01  data.order.id=d4969423  idx=53  role=Buyer   status=Dispute
PK=1425f5fe  data.order.id=727dc4df  idx=55  role=Buyer   status=Active
…
```

This is not historical residue: the first row above was created by `take_order`
minutes before it was read. The behaviour is current.

The queries split along that line:

| query | keys on |
|---|---|
| `save_trade` | binds `trade.id` → the **primary key** |
| `get_trade` | `WHERE id = ?` → the **primary key** |
| `get_trade_by_order_id` | `json_extract(data, '$.order.id')` |
| `delete_trade_by_order_id` | `json_extract(data, '$.order.id')` |
| `update_trade_fields` | `json_extract(data, '$.order.id')` |

Everything the daemon-message path touches uses the JSON id, consistently — so
nothing is broken today. `get_trade` has **no callers**, and the Dart side reads
`trade.order.id` in all 12 of its uses.

## Why it is worth an issue anyway

It is a trap with the pin already pulled. `get_trade(id)` is a public method on
the `Storage` trait that silently returns `None` for those 10 rows, and the next
person to reach for the obvious accessor gets an empty result with no error and
no log line. The row count grows by one per take.

It also makes the table hard to reason about: two columns claim to identify the
same thing, and which one is right depends on which function you are in.

## Proposal

Pick one and be explicit:

- **Make the daemon id the primary key** and migrate the existing rows. The JSON
  keeps whatever local id is genuinely needed (if any) as a named field, and
  every query keys on the same thing. Needs a migration and a decision about
  rows whose daemon id is not yet known.
- **Or keep the local id as the key and remove `get_trade`**, renaming the
  remaining accessors so the two identities are visible at the call site rather
  than implied.

Either way the invariant should be stated in `specs/004` and pinned by a test,
because the current state is not documented anywhere — it was found by comparing
the column against the JSON.

## Context

Found while investigating #393. An earlier analysis in that investigation
reached a wrong conclusion precisely because it keyed on the primary key and
[...]

Issue #394, opened 2026-09-05
A daemon message for an order with no trade row has no defined behaviour


Catrya:
When a kind-14 daemon message arrives for an order that has no row in `trades`,
the handler computes the new status, writes it, and moves on. The write matches
no row and does nothing. The app then emits a `TradeUpdate` for a trade that
does not exist.

It never asks *why* there is no row, and there are two opposite answers:

**Deleted on purpose.** A `Canceled` before the trade went active wipes the row
(`cancellation_wipes_history`, `orders.rs:1986`) — there is no meaningful
history to keep. The replayed messages for that order are noise and should be
dropped.

**Never written.** Only two places create a trade row: `create_order`
(`orders.rs:814`) and `take_order` (`orders.rs:1088`), both *after* the daemon
confirms. When that confirmation times out at 10 s and the daemon proceeded
anyway, nothing is persisted. Here the message is not noise — **it is the only
recovery signal that exists**, and it is discarded.

Measured on a real database, two consecutive starts:

```
9 and then 7 × "update_trade_fields matched no row"
```

all for trades a cancel had wiped, one of them (`afb8d76e`) three times in a
single start — three writes to nothing, three `TradeUpdate`s to the UI.

## Why the fingerprint path exists, and why it should go

Because DM-driven recovery was never implemented, maker ownership is restored
by guessing: `create_order` stores a **content fingerprint**
(`kind|fiat_code|fiat_amount|min|max|payment_method`, `orders.rs:653-666`)
before publishing, and on cold start every Kind 38383 event is re-hashed and
compared (`orders.rs:3826-3852`).

The failure mode is not "does not recover" — it is **recovers wrong**:

```rust
if let Some(trade_idx) = lookup_trade_key_index(&ck).await {
    info.is_mine = true;
    bridge_fingerprint_trade_index(&info.id, trade_idx).await;
```

Two open orders sharing those fields collide — and a taken range order is
republished on the wire as a plain fixed-amount order, so it collides with any
equivalent fixed order. That is #326: index 21 bound where 16 belonged, and
`release`/`cancel`/`rate` signed with the wrong key (`InvalidPeer`).
`bridge_fingerprint_trade_index` was then guarded to not overwrite an
authoritative binding, but **`is_mine = true` is still set unconditionally**, so
a stranger's order can still be shown as ours.

The DM path fails the other way: if the message does not arrive, nothing
happens; if it arrives, ownership is proven by decryption, not by resemblance.

[...]

PR #396, opened 2026-09-05
fix(orders): refuse status writes from an out-of-order kind-14 replay

### LINKED ISSUE #393: Startup kind-14 backlog is applied newest-first, walking trade status backwards across a restart
## Summary

On every app start, the global kind-14 catch-up subscription replays the daemon's
full message history. Relays serve stored events **newest-first**, and the dispatch
path applies each message without looking at its timestamp — so the **oldest**
message is applied **last** and wins. A trade's persisted status ends up at whatever
state it was in at the *beginning* of its life.

Reproduced deterministically on Linux, `main` @ 8c02eea. Not platform-specific.

## Reproduction

1. Create a sell order, have it taken, pay the hold invoice, open a dispute.
2. Close the app.
3. Reopen it.
4. My Trades shows the trade as `waiting-buyer-invoice`.

## Evidence

Order `625eef22-73ee-4c1d-a24d-4818a46322c4`.

**Session 1 — the progression is correct:**

```
23:23:30  →WaitingBuyerInvoice   src=kind14/WaitingBuyerInvoice
23:23:41  →WaitingPayment        src=kind14/PayInvoice
23:26:45  →Active                src=kind14/BuyerTookOrder
23:27:37  →Dispute               src=kind14/DisputeInitiatedByYou
          (app closed)
```

**Session 2 — the restart, 23:28:42.** First, proof the persisted status was still
correct at boot:

```
status order=625eef22 wire=InProgress local=Dispute applies=false src=38383/book
                                      ^^^^^^^^^^^^^ persisted state was Dispute
```

Then the kind-14 backlog replays, in exactly reverse chronological order:

```
→Dispute              (newest, applied first)
→Active
→WaitingPayment
→WaitingBuyerInvoice  (oldest, applied last ⇒ wins)
NewOrder              (no matching pending create — leaving state untouched)
```

**Resulting DB row** (`~/.local/share/mostro/mostro.db`, table `trades`) — internally
contradictory, which is the tell:

```
status                    = WaitingBuyerInvoice
data.order.status         = "WaitingBuyerInvoice"
data.hold_invoice         = "lnbcrt4110n1p4fk5hj…"   ← arrived with PayInvoice
data.counterparty_pubkey  = "3718ce14…"              ← arrived with BuyerTookOrder
```

`settings` also holds `dispute_mine:625eef22-… = 1`, so the dispute was genuinely
opened and accepted. The fields moved forward; only the status walked back.

## Root cause

Four pieces:

1. **The global kind-14 filter deliberately carries no `since` and no `limit`**
   (`rust/src/api/orders.rs:3336-3350`, "this is the offline catch-up channel"), so
   the entire history replays on every boot.
2. **Dedup is in-memory only** (`is_duplicate_daemon_message`), so every replayed
   message is processed as new after a restart.
3. **`dispatch_mostro_message` never receives the event's `created_at`**
   (`orders.rs:1531-1536`). The ordering information is discarded at the boundary —
   the handler cannot tell a replayed message from a live one.
4. **The only guard covers hard-terminal statuses.**
   `status_sync_blocked_by_terminal` (`orders.rs:2449-2467`) blocks only when the
   local status is in `is_hard_terminal` (`rust/src/mostro/status.rs:88-99`:
   `Canceled`, `CanceledByAdmin`, `CooperativelyCanceled`, `Expired`, `Success`,
   `SettledByAdmin`, `CompletedByAdmin`). `Dispute`, `Active`, `FiatSent` and
   `WaitingPayment` are unprotected and get overwritten freely.

`wire_status_applies` — the guard that does work, visible as `applies=false` in the
log above — is only wired into the two Kind 38383 ingest paths (`orders.rs:2836`,
`orders.rs:3749`). It never runs on the kind-14 path.

## This is not new behaviour that was overlooked

The docstring on `status_sync_blocked_by_terminal` already describes this exact bug:

> *"Relays deliver the startup backlog newest-first, so a progression message that
> would move a finished trade is an out-of-order replay, not a real transition —
> applying it walks the status backwards."*

And the commit that introduced that guard (`7582162`, 2026-08-07) says which cases
were left open, on purpose:

> *"SettledHoldInvoice and Dispute stay open on purpose: they still progress to
> Success / admin resolutions."*

The reasoning is sound — a dispute really can still advance — but without a timestamp
comparison it leaves the hole this issue reports.

## When it was introduced

**2026-08-07, PR #292.** Before that, the startup path built the trade-key map only
to shape the subscription filter and never seeded `global_dm_keys()`, the map the
event loop decrypts against — only `refresh_subscriptions_for_active_node` seeded it,
and that does not run on a normal boot. Every replayed kind-14 from a previous
session was therefore dropped as `no-matching-p-tag map=0`, which is what we
diagnosed as "cause 3" in #277. The backlog reached the app but never applied, so it
could not corrupt anything.

`57d87e9 fix(orders): rehydrate kind-14 decryption coverage at startup` fixed the
seeding. Nineteen minutes later, in the same PR, `7582162` added the terminal-only
guard — the reordering surfaced immediately and was patched halfway.

The `Dispute` case only became reachable in practice once #275 (2026-09-01) and #256
(2026-09-03) made disputes usable end to end.

## Scope — it is not one order

Status writes per order, extracted from a single restart log:

| order | applied sequence | ends at |
|---|---|---|
| `625eef22` | Dispute → Active → WaitingPayment → WaitingBuyerInvoice | **WaitingBuyerInvoice** |
| `916d0dc5` | Active → WaitingBuyerInvoice → WaitingPayment | **WaitingPayment** |
| `0ed2bc2f` | Canceled → WaitingBuyerInvoice | **WaitingBuyerInvoice** |
| `e737ffe7` | WaitingBuyerInvoice → Canceled → WaitingPayment | **WaitingPayment** |

The last two cross a hard-terminal status, which should have been blocked. Second
hole: `current_local_status` (`orders.rs:2435-2442`) reads the trades row and falls
back to the order book. A trade canceled before going active is **deleted** from the
DB (`orders.rs:1986`, `cancellation_wipes_history`), so the guard loses its memory,
returns `None`, and fails open. For those the DB row is already gone so the `UPDATE`
is a no-op, but `order_book().update_order_status` and `emit_trade_update` still fire.

## Suggested fix

1. **Root fix:** thread the event's `created_at` into `dispatch_mostro_message` and
   reject a status write whose event is older than the last one applied to that order
   (persisting `last_status_event_at` per order). This closes the whole class,
   including the terminal cases, and removes the need for the guard to enumerate
   statuses.
2. **Complementary:** a monotonic-transition guard on the kind-14 path, analogous to
   `wire_status_applies`, rather than a terminal-only allowlist.
3. Bounding the catch-up window (MostriX subscribes per trade key with `limit(1)` or
   `since(persisted_ts)` — `src/util/dm_utils/dm_helpers.rs:56-60`) shrinks the
   backlog but **does not fix the ordering inside it**, so it is not a substitute
   for (1).

## Side finding (same area, worth folding in or splitting out)

`update_trade_fields` (`rust/src/db/sqlite.rs:587-642`) matches on
`WHERE json_extract(data, '$.order.id') = ?` and never checks `rows_affected`, so a
write against a non-existent trade silently succeeds. Both callers report failures
with `log::warn!`, which this app never emits (`install_log_bridge` is not wired, so
the `log` facade has a no-op logger). A status sync that writes nothing is currently
invisible. Should report `rows_affected == 0` via `blog_warn`.

Catrya:
Closes #393.

## Problem

The global kind-14 subscription carries no `since`, so every start replays the
node's full history for an order, and relays serve stored events
**newest-first**. The dispatch arms applied each message as if it had just
arrived, so the **oldest** one landed last and won.

A disputed trade came back as `waiting-buyer-invoice` after a restart, with
every state in between written to the DB and emitted to the UI on the way down:

```
23:27:37  →Dispute              (session 1, correct)
          ── restart ──
23:28:42  wire=InProgress local=Dispute applies=false   ← the persisted state was still right
23:28:42  →Dispute              (newest, applied first)
23:28:42  →Active
23:28:42  →WaitingPayment
23:28:42  →WaitingBuyerInvoice  (oldest, applied last ⇒ wins)
```

Introduced on 2026-08-07 by #292: before it, the startup path never seeded
`global_dm_keys()`, so every replayed message was dropped undecrypted and the
backlog could not corrupt anything. Nineteen minutes later, in the same PR,
`7582162` added a guard for the case it had just made visible — but only for
hard-terminal statuses, and its own message records that `SettledHoldInvoice`
and `Dispute` were left out on purpose because they still progress. Sound
reasoning, missing the piece that makes it safe: comparing timestamps.

## Change

**`created_at` was already there and thrown away.** `UnwrappedMessage` carries
the kind-14 event's own timestamp (`unwrap_message_nip44` sets it from
`event.created_at`, the key relays order by) and `dispatch_mostro_message`
discarded it with `created_at: _`. No signature plumbing was needed.

- **`refactor(orders)`** — bind it and report it as `age` on the per-message log
  line. No behaviour change; it makes a replayed message distinguishable from a
  live one in the field.
- **`fix(orders)`** — persist a per-order high-water mark,
  `status_cursor:<order_id>`, and refuse any status write from a **strictly
  older** message. Strictly older, because the daemon emits several messages for
  one order within the same second. Same shape as the `chat_cursor:` that
  already bounds the other channel, so no schema or trait change.
- The hard-terminal guard stays. It is not redundant: it reads the in-memory
  book, so it is the only one of the two that still protects a session with no
  durable store (web, #233).
- The cursor is deliberately **not** cleared with the trade row. A cancel before
  the trade went active wipes that row, which is exactly where the terminal
  guard goes blind; living in settings, the cursor outlives the wipe.
- **`docs(specs)`** — the orders contract described the terminal guard as the
  whole story and quoted the gap without saying what covered it. Split into the
  two rules, and record that the bulk filter is unbounded on purpose so the
  missing `since` is not read as an oversight.
- **`fix(db)` + `test(db)`** — collateral finding, folded in because it is the
  reason this class of bug was invisible: `update_trade_fields` matched on
  `WHERE json_extract(...)` and discarded the result, so a write against a row
  that is gone returned `Ok(())`; and the seven callers reported failures with
  `log::warn!`, which this app never emits (`install_log_bridge` is not wired,
  so the global filter stays `Off`). Now it warns on `rows_affected == 0` and
  reports through `blog_warn`.

## Test plan

Unit tests, all pinned by mutation (reverting the code under test makes them
fail):

- [x] `a_replayed_backlog_cannot_walk_the_status_backwards` — replays a backlog
      newest-first through the real dispatch path and asserts the book, the
      persisted row, the cursor, and that no intermediate `TradeUpdate` is
      emitted. Without the fix: `left: WaitingBuyerInvoice, right: Dispute`.
- [x] `a_trade_update_that_matches_no_row_is_reported`
- [x] `a_trade_update_that_fails_is_not_reported_as_a_write`
- [x] `cargo test --locked` — 406 passed
- [x] `cargo clippy --locked -- -D warnings` — clean
- [x] `cargo check --locked --target wasm32-unknown-unknown` — clean

**Validated against a real database**, not only the tests. Three orders already
corrupted by this bug healed on the first launch, with the outcome predicted
before running:

| order | before | after |
|---|---|---|
| `625eef22` | `WaitingBuyerInvoice` | `Dispute` |
| `916d0dc5` | `WaitingPayment` | `Active` |
| `84599f9e` | `WaitingBuyerInvoice` | `SettledHoldInvoice` |

Same startup, before and after:

| | messages | status writes | refused (order) | refused (terminal) |
|---|---|---|---|---|
| before | 198 | **78** | 0 | 53 |
| after, 1st start | 187 | **39** | 37 | 47 |
| after, 2nd start | 187 | **22** | 54 | 47 |

Half the status writes were backwards steps. The second start moved **no**
status at all — the cursor is durable, so refusal happens from the first
message. Backlog ages ranged from 5.7 h to 706 h, now visible on every line.

The guard does not over-block: `d4969423` received its messages in ascending
order (ages 100338 → 100319 → 100303 → 100243) and applied all four writes,
ending at `Dispute`. Only strictly older messages are refused.

A full take → restart cycle was run as well. The live reply logged `age=0s`, the
same message came back in the next start's replay at `age=35s` and was applied
once, and no trade changed status across the restart.
order (ages 100338 → 100319 → 100303 → 100243) and applied all four writes,
ending at `Dispute`. Only strictly older messages are refused.

A full take → restart cycle was run as well. The live reply logged `age=0s`, the
same message came back in the next start's replay at `age=35s` and was applied
once, and no trade changed status across the restart.

## Two residual findings, deliberately not fixed here

Both pre-existed and were silent; the new `rows_affected == 0` warning is what
surfaced them. Filed separately:

- **#394** — a daemon message for an order with no trade row has no defined
  behaviour. 7 orders whose row a cancel deleted still have their history
  replayed and re-applied on every start; the write lands nowhere and a
  `TradeUpdate` is still emitted. The cursor trims what it can (9 warnings → 7
  on the second start) but this is not an ordering problem: the trade does not
  exist. The same issue covers the opposite case — a create or take whose
  confirmation timed out while the daemon proceeded, where the message is the
  only recovery signal there is — and the content-fingerprint heuristic that
  stands in for it today. Also carries the idempotent-rewrite note: 10 of the 22
  remaining writes rewrite the value the row already holds.
- **#395** — `trades.id` and `data.order.id` diverge in 10 of 36 rows. Not
  reached by this PR's paths (they all key on the JSON id), but `get_trade`
  keys on the primary key and would return `None` for those rows.

Neither is a regression.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented order statuses from moving backward when delayed or replayed updates are received.
  - Preserved completed and canceled order states against later conflicting updates.
  - Improved reliability when recovering order history after downtime.
  - Added clearer warnings when order updates cannot be matched or saved.

- **Documentation**
  - Clarified order-status synchronization, replay behavior, ordering rules, and terminal-state handling.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1547,12 +1547,18 @@ async fn dispatch_mostro_message(
     //
     // A forger cannot sign a kind-14 event as the node, so `sender == mostro`
     // is the authoritative check.
+    //
+    // `created_at` is the kind-14 event's own timestamp, not an inner field
+    // (`unwrap_message_nip44` sets it from `event.created_at`) — the very key
+    // relays order their stored events by. It is the only thing that separates
+    // a live daemon reply from one being replayed out of a startup backlog,
+    // and until now it was dropped here.
     let mostro_core::nip59::UnwrappedMessage {
         message: msg,
         sender,
         identity: _,
         signature: _,
-        created_at: _,
+        created_at: event_created_at,
     } = unwrapped;
 
     // Daemon authentication: the kind-14 event author (`sender`) must be the
@@ -1622,9 +1628,21 @@ async fn dispatch_mostro_message(
         Some(other) => format!("{other:?}"),
         None => "None".to_string(),
     };
+    // `age` is the event's timestamp against the local clock. A live reply reads
+    // ~0; the global kind-14 feed carries no `since`, so on every start it
+    // replays the node's full history and those read hours or days. Relays hand
+    // that backlog back newest-first while the arms below apply each message as
+    // if it had just arrived, so a large age marks writes that are about to
+    // overwrite fresher state. Negative means the node's clock runs ahead.
+    let event_age_secs = crate::rt::unix_now().saturating_sub(event_created_at.as_secs() as i64);
     crate::api::logging::blog_info("daemon-msg", format!(
-        "action={:?} order_id={:?} trade_index={:?} trade_pubkey={} payload={}",
-        kind.action, kind.id, kind.trade_index, &trade_pubkey_hex[..8], payload_desc
+        "action={:?} order_id={:?} trade_index={:?} trade_pubkey={} age={}s payload={}",
+        kind.action,
+        kind.id,
+        kind.trade_index,
+        &trade_pubkey_hex[..8],
+        event_age_secs,
+        payload_desc
     ));
 
     // Everything below is serialized against other handlers of this order id:
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1345,7 +1345,13 @@ pub async fn cancel_order(order_id: String) -> Result<()> {
             )
             .await
         {
-            log::warn!("[orders] failed to optimistically update cancel status for {order_id}: {e}");
+            crate::api::logging::blog_warn(
+                "orders",
+                format!(
+                    "cancel status not persisted for order={}: {e}",
+                    crate::api::logging::short_id(&order_id),
+                ),
+            );
         }
     }
 
@@ -2036,7 +2042,13 @@ async fn dispatch_mostro_message(
                             )
                             .await
                         {
-                            log::warn!("[orders] failed to sync Canceled status for {oid}: {e}");
+                            crate::api::logging::blog_warn(
+                                "orders",
+                                format!(
+                                    "Canceled status not persisted for order={}: {e}",
+                                    crate::api::logging::short_id(&oid),
+                                ),
+                            );
                         }
                     }
                 }
@@ -2103,8 +2115,12 @@ async fn dispatch_mostro_message(
                         .update_trade_fields(&order_id, Some(new_status.clone()), None, None)
                         .await
                     {
-                        log::warn!(
-                            "[orders] failed to sync status for order={order_id}: {e}"
+                        crate::api::logging::blog_warn(
+                            "orders",
+                            format!(
+                                "status not persisted for order={}: {e}",
+                                crate::api::logging::short_id(&order_id),
+                            ),
                         );
                     }
                 }
@@ -2247,8 +2263,12 @@ async fn dispatch_mostro_message(
                     )
                     .await
                 {
-                    log::warn!(
-                        "[orders] failed to save hold invoice for order={order_id}: {e}"
+                    crate::api::logging::blog_warn(
+                        "orders",
+                        format!(
+                            "hold invoice and status not persisted for order={}: {e}",
+                            crate::api::logging::short_id(&order_id),
+                        ),
                     );
                 }
             }
@@ -2306,8 +2326,12 @@ async fn dispatch_mostro_message(
                         .update_trade_fields(&order_id, Some(status.clone()), None, None)
                         .await
                     {
-                        log::warn!(
-                            "[orders] failed to sync trade status for order={order_id}: {e}"
+                        crate::api::logging::blog_warn(
+                            "orders",
+                            format!(
+                                "status not persisted for order={}: {e}",
+                                crate::api::logging::short_id(&order_id),
+                            ),
                         );
                     }
                 }
@@ -2964,9 +2988,12 @@ async fn subscribe_single_order(order_id: &str) {
                                     )
                                     .await
                                 {
-                                    log::warn!(
-                                        "[orders] failed to sync d-tag trade status for order={}: {e}",
-                                        order.id
+                                    crate::api::logging::blog_warn(
+                                        "orders",
+                                        format!(
+                                            "d-tag status not persisted for order={}: {e}",
+                                            crate::api::logging::short_id(&order.id),
+                                        ),
                                     );
                                 }
                             }
@@ -3879,9 +3906,12 @@ async fn ingest_order_event_with(event: &nostr_sdk::prelude::Event, publish: Pub
                             )
                             .await
                         {
-                            log::warn!(
-                                "[orders] failed to sync trade status for order={}: {e}",
-                                info.id
+                            crate::api::logging::blog_warn(
+                                "orders",
+                                format!(
+                                    "status not persisted for order={}: {e}",
+                                    crate::api::logging::short_id(&info.id),
+                                ),
                             );
                         }
                     }
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -2531,25 +2531,52 @@ async fn load_status_cursor(order_id: &str) -> Option<i64> {
 /// Record that a daemon message dated `event_created_at` was allowed to write
 /// `order_id`'s status.
 ///
-/// Clamped to the local clock for the same reason the chat cursor is: a node
-/// dating an event in the future would otherwise push the mark past every
-/// message that follows and freeze the order's status for good.
+/// The mark is stored **raw**, in the node's own time domain, because that is
+/// the domain [`status_write_blocked`] compares against. Clamping it to the
+/// local clock — as the chat cursor does — is wrong here: there the cursor is a
+/// subscription `since`, where a low value only asks for more than needed, but
+/// here it is an ordering comparator. With a local clock behind the node's, a
+/// newest event at 3000 would store 1000 and a *later, older* event at 2000
+/// would then pass `2000 < 1000` and overwrite it — the very replay regression
+/// this guard exists to stop (PR #396 review).
+///
+/// The freedom that costs is bounded by a skew check instead: an event dated
+/// implausibly far ahead of the local clock does not move the mark, so one
+/// malformed timestamp cannot silence an order's status for good. Same
+/// tolerance the chat envelope already applies to node-adjacent events. A node
+/// whose clock is genuinely further ahead makes the guard inert for that order
+/// rather than wrong — logged, because that is a silent degradation otherwise.
+/// Ordering between the node's own events survives any uniform skew: they all
+/// carry the same clock.
 ///
 /// The read-modify-write is safe because every caller runs under this order's
-/// dispatch lock. The `>=` check is not redundant with [`status_write_blocked`]:
-/// the clamp can pull an accepted event *below* the current mark when the local
-/// clock runs behind the node's, and the mark must never move backwards.
+/// dispatch lock.
 ///
 /// Best-effort, like every other write here: losing it costs the durability of
 /// the guard, not its correctness within the session.
 async fn record_status_event(order_id: &str, event_created_at: i64) {
-    let accepted = event_created_at.min(crate::rt::unix_now());
-    if load_status_cursor(order_id).await.is_some_and(|c| c >= accepted) {
+    let horizon = crate::rt::unix_now()
+        .saturating_add(crate::nostr::transport::MAX_CLOCK_SKEW_SECS as i64);
+    if event_created_at > horizon {
+        crate::api::logging::blog_warn(
+            "orders",
+            format!(
+                "status cursor not advanced for order={}: event is {}s ahead of the local clock",
+                crate::api::logging::short_id(order_id),
+                event_created_at.saturating_sub(horizon),
+            ),
+        );
+        return;
+    }
+    if load_status_cursor(order_id)
+        .await
+        .is_some_and(|c| c >= event_created_at)
+    {
         return;
     }
     if let Some(db) = crate::db::app_db::db() {
         let key = crate::db::settings_keys::status_cursor(order_id);
-        if let Err(e) = db.set_setting(&key, &accepted.to_string()).await {
+        if let Err(e) = db.set_setting(&key, &event_created_at.to_string()).await {
             crate::api::logging::blog_warn(
                 "orders",
                 format!(
@@ -2577,6 +2604,18 @@ async fn record_status_event(order_id: &str, event_created_at: i64) {
 /// Strictly older is what is refused: the daemon emits several messages for one
 /// order within the same second (the `PayInvoice` reputation follow-up, for
 /// one), and those are genuine, in-order traffic.
+///
+/// That strictness is also why the callers record the mark *before* the writes
+/// they gate, rather than after a successful one (PR #396 review). The mark
+/// answers "which message have I decided to accept", not "which write
+/// succeeded" — every write here is best-effort, `record_status_event`
+/// included, and there is no transaction spanning the settings key and the
+/// trade row. Recording afterwards would leave the mark unmoved when a write
+/// fails, and then a *strictly older* message from the same backlog would be
+/// applied over the row that just failed to update — trading a fault that
+/// heals for one that corrupts. It heals because the failed message is not
+/// blocked on its next delivery: its timestamp equals the mark, and equal
+/// passes. The global feed carries no `since`, so the next start replays it.
 async fn status_write_blocked(
     order_id: &str,
     action: &mostro_core::message::Action,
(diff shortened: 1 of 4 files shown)


Catrya:
@coderabbitai review

rust/src/api/orders.rs
```diff
@@ -1954,9 +1979,10 @@ async fn dispatch_mostro_message(
                 // and completed — must not overwrite the terminal outcome.
                 // The wipe path below is unaffected: it starts from
                 // pending/waiting, which are not terminal.
-                if status_sync_blocked_by_terminal(&oid, &kind.action).await {
+                if status_write_blocked(&oid, &kind.action, event_ts).await {
                     return;
                 }
+                record_status_event(&oid, event_ts).await;
```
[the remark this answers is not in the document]
Catrya: Not applied. The mark answers which message was accepted, not which write succeeded — every write in this module is best-effort, record_status_event included, and no transaction spans the settings key and the trade row, so "commit both atomically" isn't available.

The described fault also heals on its own: the guard refuses strictly older messages, so the message whose write failed is not blocked on its next delivery — its timestamp equals the mark, and equal passes. The global feed carries no since, so the next start replays it.

Recording afterwards would be worse: with the mark unmoved after a failed write, a strictly older message from the same backlog would be applied over the row that just failed to update. That trades a fault that heals for one that corrupts. The reasoning is now a comment on status_write_blocked.

rust/src/db/mod.rs
```diff
@@ -62,6 +62,28 @@ pub mod settings_keys {
     pub fn dispute_mine(order_id: &str) -> String {
         format!("{DISPUTE_MINE_PREFIX}{order_id}")
     }
+
+    /// Per-order status replay cursor — the `created_at` (unix seconds,
+    /// decimal string) of the newest daemon message whose status write was
+    /// applied, clamped to the local clock. Full key is
```
grunch, line 68: _📝 Docs_ | _🟡 Minor_ | _⚡ Quick win_

**Stale: the mark is no longer clamped to the local clock.**

`c5b229b` changed `record_status_event` to store the timestamp **raw** (node's time domain) and bound it with a skew horizon instead; this doc, the `record_status_event` rustdoc and the spec now disagree with each other. Since this is the key's authoritative description, it will mislead the next reader into re-introducing the clamp.

```suggestion
    /// Per-order status replay cursor — the `created_at` (unix seconds,
    /// decimal string) of the newest daemon message whose status write was
    /// applied, stored raw in the node's time domain (not clamped to the
    /// local clock; see `record_status_event`). Full key is
```

rust/src/db/sqlite.rs
```diff
@@ -719,6 +735,69 @@ mod tests {
...
 
+    /// A status sync that matches no row used to be indistinguishable from one
+    /// that wrote: SQLite reports success for an `UPDATE` that touches nothing,
+    /// the result was discarded, and the caller's only failure report was a
+    /// `log::warn!` this app never emits. The write is unrecoverable either
+    /// way — the point is that it stops being silent.
+    #[tokio::test]
+    async fn a_trade_update_that_matches_no_row_is_reported() {
+        // The verbosity filter defaults to `Off`, so nothing reaches any sink
+        // until this runs — the same call `init_app` makes in the app.
```
grunch, line 746: _📝 Docs_ | _🔵 Nitpick_

**`init_app` never runs in the app.**

`init_app` carries `#[frb(init)]` but lives in `lib.rs`, outside `crate::api`, so codegen skips it: the generated `executeRustInitializers()` is empty and nothing in Dart calls `installLogBridge`. The test is right to install the bridge itself, but the comment's "the same call `init_app` makes in the app" describes a call that does not happen. Suggest: `// ... until this runs. (The app never calls it — see the follow-up on wiring \`#[frb(init)]\`.)` and open that follow-up, since it leaves 193 `log::*` call sites crate-wide silently dropped.

rust/src/api/orders.rs
```diff
@@ -2466,6 +2514,132 @@ async fn status_sync_blocked_by_terminal(
...
+            "orders",
+            format!(
+                "status cursor not advanced for order={}: event is {}s ahead of the local clock",
+                crate::api::logging::short_id(order_id),
+                event_created_at.saturating_sub(horizon),
+            ),
+        );
+        return;
+    }
+    if load_status_cursor(order_id)
```
grunch, line 2571: _⚡ Efficiency_ | _🔵 Nitpick_

**Cursor is read twice per daemon message.**

`status_write_blocked` loads the cursor, and `record_status_event` loads it again a few lines later in every caller — two settings-table round trips per message, on the hot startup-replay path (198 messages in your own measurement). Both run under the per-order dispatch lock, so the second read cannot observe anything new.

Consider having `status_write_blocked` return the cursor it loaded (e.g. `Result<Option<i64>, Blocked>` or an `enum { Blocked, Allowed(Option<i64>) }`) and passing it into `record_status_event`, or fold the record into a single `apply_status_event(order_id, action, event_ts) -> bool` that does check-then-record in one place. Also removes the possibility of a future arm calling one without the other.

rust/src/api/orders.rs
```diff
@@ -5913,6 +6093,235 @@ mod tests {
...
+        );
+    }
+
+    /// One malformed timestamp must not be able to silence an order for good:
+    /// an event far beyond the skew tolerance is still applied, but it does not
+    /// move the mark, so the messages that follow it are not all refused.
+    #[tokio::test]
+    async fn an_absurdly_future_event_does_not_move_the_cursor() {
+        let path = std::env::temp_dir()
+            .join(format!("mostro_status_skew_{}.db", std::process::id()));
```
grunch, line 6309: _🧪 Tests_ | _🔵 Nitpick_

**Same DB filename as `a_cursor_behind_the_local_clock_still_orders_future_dated_events` (L6226).**

Harmless today because `init_db` is a `get_or_try_init` singleton and only the first caller's path is ever opened, but the duplicated name reads as a copy-paste slip and would matter if the tests were ever run in separate processes. Use a distinct suffix (`mostro_status_skew_horizon_`).

While here: the three new tests plus the existing ones repeat the same ~25-line `TradeInfo` literal. A `fn seed_trade_row(db, &order_info) -> TradeInfo` next to `dummy_order_info` would cut the noise and keep the assertions in view.

grunch (approved):
**Actionable comments posted: 4** (1 minor, 3 nitpicks). Nothing blocking.

## Verdict: ✅ Approve

The fix is correct and well-scoped: the per-order `status_cursor:` high-water mark, compared raw in the node's time domain and bounded by `MAX_CLOCK_SKEW_SECS`, closes the newest-first replay regression, and the terminal guard is rightly kept as the web fallback. The `rows_affected == 0` warning in `update_trade_fields` is a genuine collateral improvement.

### Verified locally (head `c5b229b`)

| check | result |
|---|---|
| `cargo test --locked` | 408 passed, 0 failed, 18 ignored |
| `cargo clippy --locked -- -D warnings` | clean |
| `dispatch_mostro_message` status-writing arms (`Canceled`, small-order sync, `AddInvoice`, `PayInvoice`, `status_for_action`) | all five gated by `status_write_blocked` + `record_status_event` |
| `unwrap_mostro_message` (daemon path) | has **no** future-date rejection (only `mostro_unwrap` does), so the horizon branch in `record_status_event` is reachable in production and the skew test models a real path |
| `init_db` is a `get_or_try_init` singleton | the three new tests share one store; per-test file names are cosmetic |

### Prior CodeRabbit threads

- **Cursor recorded before the write** — the author's rebuttal holds: the mark records the accepted message, not the completed write; equal timestamps pass, so a failed write self-heals on the next replay, whereas recording afterwards would let a strictly older message overwrite the row. Agree, no change needed.
- **Cursor clamped to the local clock** — resolved by `c5b229b` (raw mark + skew horizon) and pinned by `a_cursor_behind_the_local_clock_still_orders_future_dated_events` and `an_absurdly_future_event_does_not_move_the_cursor`.

### One premise in the PR body is imprecise (outcome still correct)

The body and commit `5dfb4dc` say *"`install_log_bridge` is not wired, so the global filter stays `Off`"*. What is actually true:

- `init_app` (`#[frb(init)]`) lives in `lib.rs`, outside `crate::api`, so FRB never picks it up — the generated `executeRustInitializers()` is empty and no Dart code calls `installLogBridge`. **So no logger is ever installed and every `log::warn!`/`info!`/`error!` goes to the `log` facade's no-op logger.** That part is right, and it is worse than the PR states: 94 such call sites remain in `orders.rs` alone and 193 across the crate.
- The *filter* does not stay `Off`, though: `set_logging_enabled` → `set_verbose_logging` raises `log::max_level()` from settings, which is exactly why `blog_*` (which calls `BRIDGE_LOGGER` directly) is observable while `log::*` is not.

Net effect of this PR's `log::warn!` → `blog_warn` moves is as described. Please file a follow-up to make `#[frb(init)]` effective (move `init_app` under `crate::api`, or call `installLogBridge` from Dart at startup) so the remaining 193 facade calls stop being dead; that is a real observability gap this PR only worked around.

### Accepted residual (no action)

- **Same-second ties.** Equal timestamps pass by design, so two status messages for one order inside the same second, replayed newest-first, still let the older one win. Terminal outcomes are covered by the terminal guard; non-terminal same-second pairs are rare and documented in the spec. Worth a note in #394's scope if it ever shows up in the field.

### Docs

`specs/004-mostro-p2p-client/contracts/orders.md` correctly splits the two rules. One stale sentence in `db/mod.rs` contradicts it (inline).

PR #398, opened 2026-09-06
fix(ios): declare NSCameraUsageDescription

Pivii:
## Problem

`mobile_scanner` opens the camera from the NWC wallet screen (`lib/features/settings/screens/connect_wallet_screen.dart:108` → `lib/shared/widgets/platform_aware_qr_scanner.dart`), but `ios/Runner/Info.plist` carries no `NSCameraUsageDescription`.

iOS terminates the process on the spot. **Connecting a Lightning wallet by QR code is impossible on iPhone today.**

## Reproduction

iPhone 15 Pro Max (iOS 26.6.1) and iPhone 17 Pro simulator (iOS 26.5), both on `main`:

Menu → Settings → **NWC wallet** → **Scan QR** → the app dies immediately (`Lost connection to device`).

The system log names the cause verbatim:

```
Runner[54164] [com.apple.TCC:access] This app has crashed because it attempted to
access privacy-sensitive data without a usage description.  The app's Info.plist
must contain an NSCameraUsageDescription key with a string value explaining to
the user how the app uses this data.
```

Read it back with:

```sh
xcrun simctl spawn <udid> log show --predicate 'process == "Runner"' --last 3m --style compact
```

## Why it also blocks the App Store

App Review guideline 2.1(a): *"We will reject incomplete app bundles and binaries that crash or exhibit obvious technical problems."*

## The change

Two lines in `ios/Runner/Info.plist`. Verified in the built bundle, and the scanner now opens normally after the permission prompt.

## One open question for maintainers

The string is **English-only**. iOS purpose strings live in `Info.plist`, outside the Flutter l10n pipeline; localising them requires per-locale `InfoPlist.strings` files, which this repository does not have. That felt like too much for a crash fix, but I am happy to add that infrastructure in a follow-up if you would rather ship it localised in all five languages.

---

Context: I am working on making the iOS target build and keep working. v2 does compile and run on iOS unmodified — a fuller baseline report will follow as an issue.

### DIFF
--- a/ios/Runner/Info.plist
+++ b/ios/Runner/Info.plist
@@ -45,5 +45,7 @@
 	<true/>
 	<key>UIApplicationSupportsIndirectInputEvents</key>
 	<true/>
+	<key>NSCameraUsageDescription</key>
+	<string>Mostro uses the camera to scan QR codes when you connect a Lightning wallet.</string>
 </dict>
 </plist>


21Mill (commented):
**LGTM.** The key is required, not merely advisable: `mobile_scanner: ^7.2.0` (pubspec.yaml:42) is used by `lib/shared/widgets/platform_aware_qr_scanner.dart`, and without `NSCameraUsageDescription` iOS terminates the process on the first camera access rather than showing a denial. `ios/Runner/Info.plist` previously carried no `NS*UsageDescription` key at all, while Android already declares `CAMERA` (AndroidManifest.xml:7) — so this closes a real platform gap.

Scope looks right: the scanner has a single call site (`lib/features/settings/screens/connect_wallet_screen.dart:108`), so the "connect a Lightning wallet" wording is accurate. There is no `analyzeImage`/`pickImage` usage anywhere in `lib/`, so no `NSPhotoLibraryUsageDescription` is needed — good that it wasn't added speculatively.

Two non-blocking notes, neither of which should hold this up:

1. **The string is unlocalized.** The app ships `app_{de,en,es,fr,it}.arb`, but there is no `*.lproj` containing `InfoPlist.strings`, so this prompt will read English for every locale. That's pre-existing (there's no Info.plist localization infrastructure yet) — worth a follow-up issue rather than a change here.

2. **Deployment-target mismatch nearby.** `ios/Podfile:2` and `:45` pin `12.0` while `ios/Runner.xcodeproj/project.pbxproj:349` sets `IPHONEOS_DEPLOYMENT_TARGET = 13.0`, and `mobile_scanner` 7.x requires iOS 15+. Unrelated to this diff, but it's the next thing likely to break an iOS build once the plist is fixed — probably worth its own issue.

I could not verify how the prompt actually renders on device; that's cosmetic and the text reads fine as written.

Catrya (approved):
**Approved.** Minimal, correct fix for a real crash, and the scope is right — no speculative keys added.

#### Your open question: ship it English-only, follow-up for the rest

Localising this belongs in its own PR, and it is more work than it looks. It is not just the missing `InfoPlist.strings`: `ios/Runner` has **no per-locale `*.lproj` at all** (only `Base.lproj`, holding the two storyboards), and the bundle does not declare `CFBundleLocalizations`. So localising one purpose string means standing up the iOS bundle localisation infrastructure from scratch for five locales — out of proportion in a crash fix, and better reviewed on its own.

One thing worth recording in that follow-up: `CLAUDE.md` states that **all** user-facing strings are Dart-level (`lib/l10n/app_*.arb` → `AppLocalizations`). This is the first user-facing string in the repo that *cannot* be — iOS reads it from the bundle before the Flutter engine exists. That exception deserves a line in the guidelines, or the next person will "fix" it by moving it into an `.arb`.

## What I verified

- **The plist parses** (`plistlib`, 18 keys) and the new key is well-formed. `CFBundleDisplayName` is `Mostro`, so "Mostro uses the camera…" matches the name the user actually sees.
- **The wording covers the whole camera surface.** `PlatformAwareQrScanner` has exactly one call site — `connect_wallet_screen.dart:108` — so "when you connect a Lightning wallet" is the complete truth today, not a partial one.
- **No `NSPhotoLibraryUsageDescription` is needed**: zero uses of `analyzeImage`, `pickImage` or `ImagePicker` anywhere in `lib/`. Good that it wasn't added speculatively.
- **`main` carried no `NS*UsageDescription` key at all** — this is the first.
- **Android parity**: `AndroidManifest.xml:7` already declares `android.permission.CAMERA`, so the gap was iOS-only.
- **Merge with today's `main`**: clean (22 commits behind).

## Non-blocking

**1. The same class of App Review risk is already in the tree.** `permission_handler: ^11.4.0` and `file_picker: ^8.3.7` are declared in `pubspec.yaml` and used **nowhere** — zero references across the whole repo apart from docs and the generated Windows plugin registrant. `mobile_scanner` requests camera access itself (`AVCaptureDevice`), so `permission_handler` is not carrying this flow either. But its pod is integrated regardless and compiles *every* handler (photos, contacts, location, microphone…) unless the `Podfile` narrows them with the `GCC_PREPROCESSOR_DEFINITIONS` / `PERMISSION_*` block — which this `Podfile` does not have. The binary then references privacy APIs with no matching usage strings, which is the same rejection family this PR argues for the camera. Fits your announced iOS baseline report; the cleanest fix is probably deleting both dead dependencies.

**2. CI has not run on this branch.** The workflow run sits in `action_required` (fork PR), so a maintainer needs to approve it — the green check currently visible is CodeRabbit's, not CI's.

**3. On the deployment-target note from the earlier review:** I confirm the file-level mismatch — `ios/Podfile` pins `12.0` (lines 2 and 45) while `project.pbxproj` sets `13.0` in all three configurations. I could **not** confirm the "`mobile_scanner` 7.x requires iOS 15+" part (no 7.x podspec available here), and since you built and ran the app on both a device and a simulator, it is not breaking anything today. I would file it as a file-consistency item in the baseline issue rather than as an imminent break.

## What I could not verify

- Anything on iOS first-hand — I'm on Linux, so the crash, the TCC log line and the permission prompt rest on your reproduction, which is well documented (device, simulator, iOS versions, and the `simctl log show` command to read it back).
- How the prompt renders, and whether the sentence truncates on smaller screens.

PR #399, opened 2026-09-06
fix(ios): apply Runner.entitlements to the Runner target

Pivii:
## Problem

`ios/Runner/Runner.entitlements` exists and declares a keychain access group for `flutter_secure_storage`:

```xml
<!-- Keychain access group for flutter_secure_storage (PIN, session data) -->
<key>keychain-access-groups</key>
<array><string>$(AppIdentifierPrefix)foundation.mostro.app</string></array>
```

But `CODE_SIGN_ENTITLEMENTS` is not set on any Runner build configuration, and the file is not referenced anywhere in `project.pbxproj`. **Xcode never reads it.** The Flutter template does not create this file, so it was clearly written on purpose and then not wired up.

## Evidence

Entitlements read back from a signed device build of `main`:

```
application-identifier
com.apple.developer.team-identifier
get-task-allow
```

After this change, `keychain-access-groups` is present as declared. Checked with:

```sh
codesign -d --entitlements :- <path>/Runner.app | plutil -p -
```

## Why fix it now, given nothing is broken

Nothing is broken today: iOS falls back to the app's default access group, so `flutter_secure_storage` works either way, and I saw no `errSec` in a full session on device.

The reason to fix it now is the next step. `firebase_messaging` is already a dependency, so `aps-environment` will have to be added to this same file to enable APNs. Doing so would **silently do nothing**, with no error at build time, at signing time or at runtime. That is an expensive thing to debug, and a one-line change to prevent.

## The change

Three lines: `CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;` on the Runner target's Debug, Release and Profile configurations. The `RunnerTests` target is untouched.

Validated on an iPhone 15 Pro Max (iOS 26.6.1): app builds, signs, installs and runs; identity and settings survive a cold restart.

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -361,6 +361,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
@@ -540,6 +541,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
@@ -562,6 +564,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
+				CODE_SIGN_ENTITLEMENTS = Runner/Runner.entitlements;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;


21Mill (commented):
**LGTM.** Verified the premise: `CODE_SIGN_ENTITLEMENTS` appears nowhere in `ios/Runner.xcodeproj/project.pbxproj` on `main`, so "Xcode never reads it" is exactly right — the file was dead weight. The three additions land on the Runner target's Debug/Release/Profile configs (each immediately above `PRODUCT_BUNDLE_IDENTIFIER = foundation.mostro.app` at :371, :550, :572), and the `RunnerTests` configs (:387, :404, :419) are untouched, as described.

The thing worth stating explicitly, since it's the one way a change like this can hurt existing users: adding `keychain-access-groups` makes its **first entry the default access group** for items stored without an explicit group — which is how `flutter_secure_storage` stores them in `lib/core/services/identity_service.dart`. Had the declared group not matched the app identifier, installed users would have lost access to their existing PIN and session data on upgrade. Here the group is `$(AppIdentifierPrefix)foundation.mostro.app` and the bundle id is `foundation.mostro.app`, so the default group is unchanged and **no migration is needed**. Worth noting for whoever revisits this file later: that match is load-bearing, not cosmetic.

Supporting evidence that this is an oversight rather than a deliberate omission: `macos/Runner.xcodeproj/project.pbxproj` already sets `CODE_SIGN_ENTITLEMENTS` on all three of its configurations (:478, :610, :630). iOS was simply the platform that got missed.

One correction to the rationale, not to the code: the PR says `docs/firebase-setup.md` points towards adding `aps-environment`. That file is 11 lines long and mentions iOS exactly once — "ensure `GoogleService-Info.plist` is in `ios/Runner/`" — with no reference to APNs, push, or entitlements. The underlying argument stands on its own without the citation, since `firebase_messaging: ^15.1.3` is already a dependency and will need `aps-environment` in this very file. I'd just drop the reference so it doesn't send a future reader looking for guidance that isn't there.

Also flagging for maintainers rather than for this PR: there are no iOS or macOS workflows under `.github/workflows/`, so nothing in CI would have caught the dangling entitlements file, and nothing will catch it if the setting is dropped again. The author's `codesign -d --entitlements :-` check is the right verification and, as things stand, the only one available.

I don't have an iOS device, so the on-device half of the evidence (the before/after entitlements dump) is taken at face value; everything above is from the tree.

Pivii:
Good catch on the `docs/firebase-setup.md` reference — you are right, that file says nothing about APNs or entitlements. I have removed it from the description; the argument stands on `firebase_messaging` already being a dependency.

The point about `keychain-access-groups` becoming the default access group is the one I had underweighted, and it is the right thing to have on the record: the match between the declared group and the bundle identifier is what makes this safe for installed users, and it would not be obvious to whoever edits that file next.

Also agreed on the missing Apple CI — nothing would catch a regression here. I am putting together a broader iOS baseline issue and will include it there.

Catrya (approved):
**Approved.** Three lines, and they land exactly where the description says.

## What it fixes

`ios/Runner/Runner.entitlements` has been in the tree since the Phase 1/2 scaffold, but no build configuration ever pointed at it — I confirmed there is no `CODE_SIGN_ENTITLEMENTS` and no file reference anywhere in `main`'s `project.pbxproj`, so "Xcode never reads it" is literal.

## Verified

**The three lines are on the Runner target and only there.** Rather than trust the hunk headers, I parsed the `pbxproj` at the head and mapped every `XCBuildConfiguration` to its target through `PRODUCT_BUNDLE_IDENTIFIER`:

| configuration | entitlements | bundle id |
|---|---|---|
| Debug `97C14706` | **yes** | `foundation.mostro.app` |
| Release `97C14707` | **yes** | `foundation.mostro.app` |
| Profile `249021D4` | **yes** | `foundation.mostro.app` |
| Debug / Release / Profile `331C8088` `331C8089` `331C808A` | no | `foundation.mostro.app.RunnerTests` |

`RunnerTests` is untouched, as stated.

**This cannot strand anything already in the keychain — which is the question a change to entitlements has to answer.** The declared group is `$(AppIdentifierPrefix)foundation.mostro.app` and the Runner's bundle id is `foundation.mostro.app`, so the expanded group (`<TeamID>.foundation.mostro.app`) is exactly the application-identifier group iOS already uses by default. Declaring `keychain-access-groups` makes its first entry the default group for new items; here that first entry is the group already in use, so items written before and after the change live in the same place. No migration, no orphaned data.

That matters more than it looks: `IdentityService` (`identity_service.dart:19`) is the only `FlutterSecureStorage(` in the codebase and passes no `IOSOptions` or `groupId`, so **the mnemonic depends entirely on that default group**. Had the declared group not matched the bundle id, this three-line PR would have made existing identities unreachable with nothing to warn anyone. Worth a sentence in the description — it is the first thing a reviewer will worry about.

**Supply-chain pass on the build system.** For changes to files that execute at build time this repo reads the whole branch diff locally rather than the rendered patch, and checks the `pbxproj` for the things that hide well there:

- `git diff $(git merge-base ... origin/main)..<head>` → exactly 3 added lines, 0 deletions, all three identical.
- No new build phases (in particular no `PBXShellScriptBuildPhase`, which would run arbitrary code on every maintainer's machine and in CI), no change to signing identity or team, no settings applied to targets the PR does not name.
- The file being switched on was audited and traced: `Runner.entitlements` was added by @grunch in `eaf79f8` (Phase 1/2 scaffold, March 2026) and last touched by the bundle-ID bot — it is not something introduced alongside this change. Its contents are `keychain-access-groups` and nothing else.

**Other checks:** merges cleanly with today's `main` (22 commits behind), and stacks cleanly with #398 in either order — they touch different files. The forward-looking argument holds too: `firebase_core: ^3.6.0` and `firebase_messaging: ^15.1.3` are already dependencies, so `aps-environment` will need this file to be read, and without this change adding it would fail silently at every stage. That is the right reason to fix a thing that is not visibly broken.

## Non-blocking

1. **The entitlements file still is not a file reference in the project.** The build setting is enough to build and sign, but the file stays invisible in Xcode's navigator until someone toggles a capability in the UI (at which point Xcode adds it). Optional.
2. **Nothing automated covers this.** There is no iOS job in CI, and this PR's run sits in `action_required` because it comes from a fork — the green check currently on the page is CodeRabbit's, not CI's. For a `pbxproj`, which is exactly the file where a bad merge goes unnoticed, that is worth noting in the iOS baseline issue you mentioned.

## What I could not verify

I am on Linux: I did not build, sign, or read the entitlements out of a binary. The `codesign -d --entitlements :- | plutil -p -` before/after evidence is the right proof and I have no reason to doubt it, but it is yours, not mine. Everything above is static verification of the project file, the entitlements plist and the Dart side that consumes the keychain.

PR #400, opened 2026-09-06
feat: generate the launcher icons from the Mostro artwork

Pivii:
## Problem

Both platforms ship Flutter's stock icon. Not "a placeholder that looks close" — byte-identical to the template files:

| File | md5 vs Flutter 3.38.2 template |
| --- | --- |
| `ios/Runner/Assets.xcassets/AppIcon.appiconset/Icon-App-1024x1024@1x.png` | identical |
| `android/.../mipmap-{m,h,xh,xxh,xxxh}dpi/ic_launcher.png` | identical, all five |

So this is not an iOS gap — it affects Android too, and both are fixed here.

## The change

Wire up `flutter_launcher_icons` and generate the icons from the official artwork, which is the same image the v1 app uses (`MostroP2P/mobile`, `assets/images/launcher-icon.png`). The configuration mirrors v1's, with two deliberate differences described below.

## Two details worth reviewing

**1. iOS gets its own pre-composited source** (`assets/images/launcher-icon-ios.png`).

The shared artwork is a disc on a transparent square. iOS forbids an alpha channel in app icons, and `remove_alpha_ios` composites onto **white**, which leaves white corners once iOS applies its rounded mask — visible on a home screen.

Pre-compositing onto the brand background (`lib/core/app_theme.dart` `bgWell`, `#1D212C`) makes the disc merge into a full-bleed square.

It also sidesteps a bug in `flutter_launcher_icons` 0.14.4: it turns near-transparent pixels into saturated green. A source pixel of `(0,0,0,alpha=1)` came out as `(32,254,43)` — **1070 such pixels**, forming a 1px green ring around the disc. Feeding it an already-opaque image via `image_path_ios` leaves it nothing to do but resize.

Worth knowing: v1's shipped iOS icon has **both** artefacts, since it is generated the same way. This PR does not reproduce them.

**2. `adaptive_icon_background` is `#1D212C`, not v1's `#2D2D2D`**, so the disc merges into the Android adaptive icon instead of floating on a slightly different grey.

## Note for whoever regenerates these

`dart run flutter_launcher_icons` also rewrites `ASSETCATALOG_COMPILER_GENERATE_SWIFT_ASSET_SYMBOL_EXTENSIONS` in `project.pbxproj` from `YES` to `AppIcon` — an invalid value for a boolean setting. That edit is reverted in this PR and should be reverted after any regeneration.

## Validation

Built, signed, installed and visually checked on an iPhone 15 Pro Max (iOS 26.6.1): correct icon, no white corners, no ring. Generated iOS icon measured at 0 corrupted pixels, corners exactly `(29,33,44)`, no alpha channel.


Catrya (approved):
**Approved.** Merges cleanly with today's `main` and stacks cleanly with #399 and #398.

## Binary audit

38 files, 33 of them PNG, so the first pass was on the bytes themselves — signature, chunk structure, CRCs, and anything past `IEND`:

- All 33 have a valid PNG signature and **correct CRCs on every chunk**.
- **Zero bytes after `IEND`** in all 33 — the usual place for a payload to ride along in an image.
- No unexpected ancillary chunks. The only one present is `tEXt` = `Software\0www.inkscape.org`, inherited byte-for-byte from the original v1 artwork.
- Every file's dimensions match its filename, without exception.
- Every iOS icon is **RGB with no alpha channel** (Apple's requirement, and your stated intent); the Android ones are RGBA, as they should be.

**Provenance of the artwork — the check that matters most here:** `assets/images/launcher-icon.png` is **byte-identical** to `MostroP2P/mobile`'s `assets/images/launcher-icon.png` (md5 `6ae6aa966784f716918d3c50a061e319` on both). Not a lookalike; the same file.

**And the iOS source is provably derived from it.** I decoded both 1024×1024 PNGs and composited the v1 artwork over `#1D212C` myself:

```
pixels deviating by more than 2/255:  0 of 1,048,576
maximum deviation:                    0
iOS PNG corners: (29,33,44) on all four     centre: (140,197,65)
```

Exactly the v1 art over the brand background, with nothing else hidden in it. And `#1D212C` is confirmed as `bgWell` — `lib/core/app_theme.dart:236`, `bgWell: Color(0xFF1D212C)`.

**Nothing in the diff executes.** The non-binary changes are `pubspec.yaml` (one dev dependency plus a config block), `AndroidManifest.xml` (a single line, `@mipmap/ic_launcher` → `@mipmap/launcher_icon`), new `colors.xml` and `launcher_icon.xml`, and a regenerated `Contents.json`. **`project.pbxproj` is not in the diff**, which confirms the generator's invalid `ASSETCATALOG_COMPILER_GENERATE_SWIFT_ASSET_SYMBOL_EXTENSIONS` edit really was reverted rather than quietly shipped.

## Claims in the description, checked

**"Byte-identical to the Flutter 3.38.2 template" — correct**, verified against the local SDK at the version CI pins:

| file | repo (`main`) | Flutter template | |
|---|---|---|---|
| `mipmap-mdpi/ic_launcher.png` | `62703444…` | `62703444…` | identical |
| hdpi / xhdpi / xxhdpi / xxxhdpi | | | identical, all four |
| `Icon-App-1024x1024@1x.png` | `c785f893…` | `c785f893…` | identical |

All six. The table is exact, including the point that this was never an iOS-only gap.

**The asset catalog is internally consistent:** `Contents.json` is valid JSON, 25 entries, 21 distinct filenames, **no reference to a missing file and no orphaned PNG**, and the 1024 marketing icon is present.

**The dependency is in the right place** — `dev_dependencies`, so the generator does not ship with the app.

## Non-blocking

1. **The stock Flutter icon still ships inside the APK.** The five `android/app/src/main/res/mipmap-*/ic_launcher.png` are not deleted and nothing references them any more: the manifest points at `@mipmap/launcher_icon`, and there is no other `ic_launcher` reference in `lib/` or the manifest (the `ic_launcher_foreground` / `ic_launcher_background` names that *are* referenced belong to the new adaptive icon). Since the stated goal is to stop shipping the template icon, a `git rm` of those five finishes the job.
2. **The `pbxproj` gotcha is recorded where nobody will read it.** That `dart run flutter_launcher_icons` rewrites `ASSETCATALOG_COMPILER_GENERATE_SWIFT_ASSET_SYMBOL_EXTENSIONS` to an invalid value, and that it must be reverted, is exactly what the next person regenerating icons needs — and they will not find it in the body of a PR merged months earlier. The `flutter_launcher_icons:` block in `pubspec.yaml` already carries a "Regenerate with: …" comment; one more line there puts the warning in front of them.
3. `^0.14.2` resolves to 0.14.4 and its green-ring bug, but the output is committed and the iOS source is already opaque, so it does not affect what ships.

## What I could not verify

- **Nothing visual on a device** — I'm on Linux, with no macOS or Xcode. That the icon looks right under iOS's rounded mask and in Android's adaptive shape rests on your device check.
- **That each generated size is a correct downscale** of the original: I verified the 1024 source pixel by pixel, but did not re-scale the remaining 20 to compare.
- The adaptive icon's rendering (foreground over `#1D212C`) across different Android launchers.

PR #405, opened 2026-09-08
fix(#389): guard startup so a failure names the step instead of a blank page

### LINKED ISSUE #389: Web: startup has no top-level guard — any failure before runApp is a silent blank page
**Type:** bug / web startup robustness
**Priority:** low severity individually, but it is the failure mode with no floor

Follow-up to #227. That issue had two halves: fix the specific cause (an
unparseable browser locale), and make startup survivable regardless of the
cause. #370 does the first. This issue carries the second, which is the part
that generalises.

## The problem

`lib/main.dart` is, in full:

```dart
Future<void> main() => bootstrapAndRun();
```

No guard of any kind. Inside `bootstrapAndRun` (`lib/core/app_bootstrap.dart`),
the stretch between `WidgetsFlutterBinding.ensureInitialized()` and `runApp`
(line 189) runs several steps unprotected — `RustLib.init()`,
`SharedPreferences.getInstance()`, `setLoggingEnabled` — and the Firebase call
is guarded by `on UnsupportedError catch (e)` (line 53), which is too narrow
for a call into a third-party JS SDK: any `FirebaseException`, network error or
malformed config propagates straight out.

Anything that throws in that window produces the worst possible shape for a
failure: **blank page, nothing logged, no retry, no degraded mode.** The user
sees nothing and we get nothing to debug. #227 was exactly this, and it took a
full stack-trace hunt to find a one-line cause.

Everything *after* `runApp` is already written to degrade — DB init, identity
and node rehydration all catch and continue (12 `catch` blocks in that file).
This early stretch is the outlier, not the norm.

## Acceptance criteria

Both are carried over verbatim from #227, where they remain unticked:

- [ ] No unhandled exception escapes `main()`; a failure is logged rather than
      thrown, and the app either starts degraded or shows an error surface
- [ ] `Firebase.initializeApp` cannot take down startup for any reason
      (`app_bootstrap.dart:53` — broaden `on UnsupportedError`)

## Notes

- The goal is **not** to fix another specific bug. It is that the *next*
  unknown failure in startup shows something instead of nothing.
- Web is where this bites hardest (a blank `<body>` with no console output),
  but the guard is platform-independent and should live in
  `bootstrapAndRun` / `main`, not behind `kIsWeb`.
- Worth deciding as part of this: what a failed startup should actually render.
  A minimal error scaffold that names the failing step beats a blank page even
  if nothing else can run.

## References

- #227 — the original issue; these two criteria are its unticked half
- #370 — fixes the locale cause, deliberately scoped to that

Matobi98:
Closes #389. Carries the unticked half of #227.

## The problem

`main.dart` is `Future<void> main() => bootstrapAndRun();` with no guard, and the stretch of `bootstrapAndRun` before `runApp` runs a dozen steps. Anything that throws there means `runApp` never runs: Flutter paints nothing, and the page is not broken — it is absent, with no message anywhere. Everything *after* `runApp` already degrades; this stretch was the outlier.

#227 is the precedent that motivated this, not a case it covers: that crash fires inside the engine's own `CanvasKitRenderer.initialize`, before `main()` runs, which is why #370 fixed it in `web/index.html` and stated plainly that no app-level `try/catch` could reach it.

## The fix

`lib/core/startup_sequence.dart` holds the sequence: `currentStep`, plus two ways to run a step.

* **`optional(name, body)`** — a step the app can open without. A failure is logged under a shared `[startup]` prefix and startup continues, degraded.
* **`required(name, body)`** — a step it cannot. The failure propagates to the guard, `runGuarded` (same file), which paints `StartupFailureApp` naming the step and reports the cause to the web smoke test.

Both set the label on entry, and that is the reason there are two: with only `optional`, the mandatory stretches in between carry no label of their own and run under whichever optional step finished last — so a failure there names a step that succeeded. That was the shape of this PR after round 1, caught in review.

Every step, classified by what the app can still do without it:

| Step | Without it | |
|---|---|---|
| `registering font licences` | Flutter's licence page is missing the bundled fonts' notices | optional |
| `setting up notifications` | no push notifications | optional |
| `loading the engine` | no protocol, keys, relays or chat | **required** |
| `reading your settings` | no language, walkthrough state or wallet | **required** |
| `applying your log settings` | default log verbosity | optional |
| `opening the local database` | memory-only session, no persistence | optional |
| `selecting the Mostro node` | the compiled-in default node | labelled |
| `mirroring trade key indices` | the secure-storage copy is absent | labelled |
| `loading your identity` | secure storage unavailable | labelled |
| `subscribing to bond notices` | no in-app notice for a slash | optional |
| `subscribing to bond claim updates` | no live update of a claim; My Trades and the trade detail still list claims | optional |
| `subscribing to trade updates` | no in-app cards for trade updates; trade screens open their own subscription | optional |
| `subscribing to chat messages` | no in-app cards for chat messages; chat opens its own per-trade subscription | optional |
| `connecting to the network` | opens, but nothing that needs relays works | optional |
| `reading relay status` | one diagnostic line missing | optional |
| `building the interface` | nothing to show | **required** |

The three "labelled" ones keep their own bespoke handlers — one of them reports to the web bridge probe — so they announce their name and handle their own failure rather than being forced through a helper.

The two card subscriptions come from #474 (merged into `main` while this PR was open). They feed only `EventCards`; `trade_state_provider.dart` and `chat_providers.dart` subscribe on their own, so the app works without them.

**The web smoke test now sees late failures.** `mostroBridgeReady` used to be set at the first Rust call (`selecting the Mostro node`), and `smoke.mjs` stopped watching there, so anything failing afterwards passed CI. That gap predates this PR: measured on `86b74a7` with a `throw` before the interface is built, the smoke test passed 3/3, finishing at ~1.2 s while the failure arrived at ~1.6 s. The guard made it worse, turning the throw into a `debugPrint`. `markBridgeReady()` now runs at the very end of startup, after `runApp`; any failure before it leaves only `mostroBridgeError`, which the smoke test already waits for. `smoke.mjs` reports `startup finished` / `startup failed: …`.

**`reading your settings` is a deliberate `required`.** It could degrade to defaults, but then the app would open looking freshly installed: the walkthrough again, the device language instead of the one the user picked, and no attempt to reconnect a saved wallet (NWC, not available on web). Showing that as if nothing were wrong misleads about data loss in an app people trade money through. All three measured by forcing the step to read an empty store.

**`opening the local database` runs on the web too.** Since #408 that is where web persistence lives: with this step broken, a trade taken in Chrome is gone after a reload. `startup_guard_test.dart` fails if it ends up back inside a `!kIsWeb` block.

**`connecting to the network` stays optional, measured with it forced to fail.** The app opens, and the secret words can still be viewed and backed up — they come from secure storage, not from relays. Nothing that needs relays works, though: adding a relay in Settings fails, because the relay API goes through a pool that never started; the order book spins with no message, and My Trades shows empty although the trades are stored. Those screens have no "disconnected" state yet — that is what makes this degradation confusing, not the decision to open.

## The failure surface

`lib/core/startup_failure.dart`. "Mostro could not start", the step that failed, and the cause underneath it — selectable, capped at 300 characters (counted in characters, so an emoji is never cut in half). Each copy button sits to the right of what it copies, in one aligned column:

* **Copy details** puts the step and the cause on the clipboard. On Android, iOS and Linux there is no console for the person hitting this to read, and dragging to select inside a scrolling view is awkward on a phone.
* **Copy link** copies `https://github.com/MostroP2P/app/issues`, where to report it.

The write is awaited: when the clipboard refuses it (web without permission, non-secure context) the screen says "Could not copy" instead of a false "Copied".

A footer spells out the cost of the obvious way out: without the secret words, uninstalling or clearing app data loses access to open trades, disputes and reputation. Checked against the code — trade keys derive from the mnemonic (`m/44'/1237'/38383'/0/N`), the daemon restores sessions from the identity key, and ratings sit on the master pubkey — so the screen warns rather than recommending it.

No localization, no app theme, no Rust, no SharedPreferences — any of those can be what failed, and a rescue surface that needs what broke is a second blank page. Colours are hard-coded for the same reason (the cause is `#8C94A8`, 5.3:1 on the background), and `CLAUDE.md` records the l10n exception so nobody "fixes" it. The content scrolls, so a long cause on a short screen moves out of the way rather than painting Flutter's striped overflow warning.

It answers any initial route with exactly one page. On the web the initial route is the browser's URL, and Flutter's default handling pushes one route per path segment: `/orders/abc/detail` would otherwise stack four identical copies, with back popping to a clone.

No retry button: `RustLib.init` throws when called twice, so retrying after a failure past that point would fail differently and confuse the report.

Platform-independent, as the issue asks — no `kIsWeb` in the guard itself. Web is where it bites hardest, because the in-app log viewer lives inside the app that did not start.

## Automated tests

* [x] `flutter analyze` — clean
* [x] `flutter test` — **1344 passed** after merging `main` (`fd94e2b`), including **27 from this PR**: 8 over `StartupSequence`, 14 over the failure surface, 3 over the guard and 2 over `openDatabase`
* [x] `cargo test` — 661 passed (pre-commit hook)
* [x] Web smoke test on the merge, release bundle: passes with `SMOKE_BOND_STORE=1 SMOKE_PUSH_WORKER=1`; with a `throw` injected before the interface is built it fails with `startup failed: Bad state: …`

What they pin, each mutation-checked (the fix reverted, the test seen red):

* the guard is **executed** through `runGuarded` with fakes for `runApp` and the bridge report: the failure surface is painted with the failing step and error, before CI is told. Moving the call out of the catch, swallowing the error earlier, flipping the order or not awaiting the body all fail it;
* the step reported is the one that failed, not the last that succeeded;
* a programming `Error` in an optional step is reported through `FlutterError` in debug while startup continues; an environment failure is not;
* `openDatabase` opens `mostro` on the web without asking for a directory;
* the clipboard failure message, the link button, character truncation, the aligned copy column, the footer and the title spacing.

One static grep remains, deliberately: that `_startup` calls `openDatabase` without a `!kIsWeb` around it. Executing that line would mean faking ~20 Rust calls to protect one line, and the test says so.

## Manual Testing

Round 3, by hand:

| Commit | What | Result |
|---|---|---|
| `fd94e2b` (merge with `main`) | full sell trade — `mostro-cli` sells, the app buys — local regtest on Linux | completed |
| `913c44f3` | failure screen on Linux with an injected `throw`: text, layout, both copy buttons, footer | as described above; copied text checked by pasting |
| `86b74a7` and this branch | web smoke test with the same injected `throw`: before the fix it passes, after it fails with the cause | confirmed |

Agent-run: the smoke test repeated on `913c44f3` and on the merge; an isolated Linux debug run, before the merge, to check that no optional step raised an `Error` (the two card subscriptions from #474 were not part of it).

Not tested by hand in round 3: the failure screen on web and Android, and whether #474's in-app cards appear during the trade.

Round 2, breaking one step at a time, on the code merged then:

| Broken | Observed |
|---|---|
| nothing | order book, and a trade survives a page reload |
| nothing | a full buy and a full sell against `mostro-cli`, both completed |
| font licences and bond claim updates | app opens, one `[startup]` line each |
| `RustLib.init` | "Mostro could not start: **loading the engine**", cause below it, copy button works |
| the container assembly | "**building the interface**" |
| `RustLib.init`, from a deep URL | one page; back does not land on a clone |
| relay pool init | app opens; secret words visible; adding a relay fails; order book spins |
| settings read as empty | walkthrough, device language, no wallet restore attempted |

From the earlier round, before the merge:

| Broken | Observed |
|---|---|
| the database | app opens, `[startup]` line, persistence gone until restored |
| `SharedPreferences` | "It failed while **reading your settings**" — a *different* screen |

Reloading the page 30 times on a release build gave no blank page. In a long-lived `flutter run` session one reload came up blank with a flutter_rust_bridge panic in `TradeKeyIndexStream`; a fresh dev server, on `main` and on this branch alike, gave none in 18 reloads each.

## Screenshots

Failure screen, rendered from the widget with the real fonts (phone 400×800 and desktop 1000×700):

pr405-screenshot-desktop
pr405-screenshot-phone

## Known, not addressed here

- With relay init broken the app opens but nothing that needs relays works, and the order book spins forever rather than saying it is offline. That surface has no "disconnected" state of its own.
- `[startup]` degradations reach `debugPrint` only. The in-app log viewer shows Rust's logs, and Dart has no way to write into it, so "grep `[startup]` to see what a run gave up on" holds only with a dev console attached.
- Follow-ups from review round 3 (issues to be linked): a failure while painting the first frame bypasses the guard and the ready flag; listeners started during startup outlive a failure; the step label can still go stale through the public setter.

### DIFF
--- a/lib/core/app_bootstrap.dart
+++ b/lib/core/app_bootstrap.dart
@@ -11,6 +11,7 @@ import 'package:mostro/core/storage/app_data_dir.dart'
 import 'package:mostro/core/app.dart';
 import 'package:mostro/core/mostro_defaults.dart';
 import 'package:mostro/core/startup_failure.dart';
+import 'package:mostro/core/startup_sequence.dart';
 import 'package:mostro/core/services/identity_service.dart';
 import 'package:mostro/core/test_environment.dart';
 import 'package:mostro/core/web/bridge_probe.dart';
@@ -45,78 +46,89 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
 /// defaults gone, an unreachable local relay fails the test instead of
 /// silently succeeding against a public one.
 Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
-  try {
-    await _startup(seedRelays: seedRelays);
-  } catch (e, st) {
-    // The failure surface calls runApp too, and that is the whole fix: today an
-    // exception in here means runApp never runs and Flutter paints nothing —
-    // the page is not broken, it is absent, with no message anywhere (#227).
-    debugPrint('[startup] fatal while $_currentStep: $e\n$st');
-    runApp(StartupFailureApp(step: _currentStep));
-  }
-}
-
-/// Names the startup step in progress, so a failure can say where it happened.
-///
-/// A file-level variable rather than a parameter: the only writers are the
-/// steps themselves and the only reader is the guard above.
-String _currentStep = 'starting up';
-
-/// Runs a startup step the app can do without: a failure is recorded and
-/// startup continues, so the app opens degraded rather than not at all.
-///
-/// One helper rather than a try/catch per step, so every degradation prints the
-/// same prefix — grepping `[startup]` lists everything a run gave up on, in
-/// order, which matters when one failure is the cause of the next.
-Future<void> _optional(String name, Future<void> Function() body) async {
-  _currentStep = name;
-  try {
-    await body();
-  } catch (e, st) {
-    debugPrint('[startup] $name failed — continuing without it: $e\n$st');
-  }
-}
-
-Future<void> _startup({List<String> seedRelays = const []}) async {
+  // Outside the guard on purpose: the rescue below paints through runApp, which
+  // needs the binding too. Catching a failure here would only let us try to
+  // render a screen that cannot render, so this one is honestly unguarded.
   WidgetsFlutterBinding.ensureInitialized();
 
-  // Push notifications only — the app trades, chats and settles without them.
-  // Two arms on purpose: the placeholder config is an expected state, not a
-  // failure, and folding both into one message would make every single run log
-  // a "failed" nobody reads by the time it means something.
-  _currentStep = 'setting up notifications';
+  final startup = StartupSequence();
   try {
-    await Firebase.initializeApp(
-      options: DefaultFirebaseOptions.currentPlatform,
-    );
-  } on UnsupportedError catch (e) {
-    debugPrint(
-      '[startup] Firebase not configured — push notifications disabled: $e',
-    );
+    await _startup(startup, seedRelays: seedRelays);
   } catch (e, st) {
-    // A call into a third-party JS SDK: config, network, or the SDK itself.
-    debugPrint(
-      '[startup] setting up notifications failed — continuing without it: $e\n$st',
-    );
+    // The failure surface calls runApp too, and that is the whole fix: without
+    // it an exception here means runApp never runs and Flutter paints nothing —
+    // the page is not broken, it is absent, with no message anywhere.
+    //
+    // #227 is the precedent that motivated this guard, not a case it covers:
+    // that crash fires inside the engine's own CanvasKitRenderer.initialize,
+    // before main() runs, which is why #370 fixed it in web/index.html and
+    // stated that no app-level try/catch could reach it.
+    debugPrint('[startup] fatal while ${startup.currentStep}: $e\n$st');
+    // The screen first: it is what a person is waiting for, and it is the whole
+    // point of this catch. Anything ahead of it that could throw would leave
+    // them with the blank page this exists to replace.
+    runApp(StartupFailureApp(step: startup.currentStep, error: e));
+    // Then CI. No-op off web; on web it hands test/web/smoke/smoke.mjs the
+    // cause, so the run stops with a reason instead of timing out waiting for
+    // a bridge that is never coming.
+    markBridgeFailed(e);
   }
+}
 
-  _currentStep = 'loading the engine';
-  await RustLib.init();
+Future<void> _startup(
+  StartupSequence startup, {
+  List<String> seedRelays = const [],
+}) async {
+  // Push notifications only — the app trades, chats and settles without them.
+  //
+  // The one optional step that handles anything itself: this repo ships a
+  // placeholder Firebase config, so every run throws UnsupportedError here.
+  // That is an expected state, not a failure, and letting it reach the helper
+  // would make every single run log a "failed" nobody reads by the time it
+  // means something. Caught below and reported as what it is; everything else
+  // falls through to the helper.
+  await startup.optional('setting up notifications', () async {
+    try {
+      await Firebase.initializeApp(
+        options: DefaultFirebaseOptions.currentPlatform,
+      );
+    } on UnsupportedError catch (e) {
+      // The placeholder config, which is an expected state rather than a
+      // failure. Swallowed here so it does not reach the helper's generic
+      // "failed" line, which every run would then log for nothing.
+      debugPrint(
+        '[startup] Firebase not configured — push notifications disabled: $e',
+      );
+    }
+    // Anything else — a third-party JS SDK's config, network or internals —
+    // falls through to the helper and is reported as the degradation it is.
+  });
+
+  await startup.required('loading the engine', RustLib.init);
 
   // Pre-read SharedPreferences so providers start with synchronous initial
   // values — eliminates the AsyncValue.loading() race that caused the router
   // to show the home screen before redirecting to /walkthrough on first launch.
-  _currentStep = 'reading your settings';
-  final prefs = await SharedPreferences.getInstance();
-  final firstRunComplete = prefs.getBool(kFirstRunCompleteKey) ?? false;
-  final backupDismissed = prefs.getBool(kBackupReminderDismissedKey) ?? false;
-  final backupActive = prefs.getBool(kBackupReminderActiveKey) ?? false;
-  final backupPending = backupActive && !backupDismissed;
-  final savedSettings = AppSettingsState.fromPrefs(prefs);
+  final (
+    prefs,
+    firstRunComplete,
+    backupPending,
+    savedSettings,
+  ) = await startup.required('reading your settings', () async {
+    final prefs = await SharedPreferences.getInstance();
+    final backupDismissed = prefs.getBool(kBackupReminderDismissedKey) ?? false;
+    final backupActive = prefs.getBool(kBackupReminderActiveKey) ?? false;
+    return (
+      prefs,
+      prefs.getBool(kFirstRunCompleteKey) ?? false,
+      backupActive && !backupDismissed,
+      AppSettingsState.fromPrefs(prefs),
+    );
+  });
 
   // Before any startup work below, so a failure in it is captured at the
   // verbosity the user asked for rather than the default.
-  await _optional('applying your log settings', () async {
+  await startup.optional('applying your log settings', () async {
     await settings_api.setLoggingEnabled(enabled: savedSettings.loggingEnabled);
   });
 
@@ -130,7 +142,7 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
   // arrive, and every Rust caller handles a missing database. Runs on the web
   // too: since #408 that is where web persistence lives, so skipping it there
   // would quietly take out every feature built on top of it.
-  await _optional('opening the local database', () async {
+  await startup.optional('opening the local database', () async {
     final location = databaseLocation(
       isWeb: kIsWeb,
       dataDir: kIsWeb ? null : await appDataDirPath(),
@@ -148,6 +160,9 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
   // end, so its outcome doubles as the web readiness probe CI waits on — see
   // lib/core/web/bridge_probe.dart (no-op off web).
   String activeMostroPubkey = defaultMostroPubkey;
+  // Named here rather than through a helper: the catch below does more than
+  // record the failure — it tells the web bridge probe, and CI reads that.
+  startup.currentStep = 'selecting the Mostro node';
   try {
     await settings_api.rehydrateActiveMostroNode();
     activeMostroPubkey = await settings_api.getMostroPubkey();
@@ -184,6 +199,9 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
   // Guarded like every other optional startup step: if the bridge is broken
   // the mirror is simply absent — the database copy is still the primary
   // record — rather than taking startup down before the UI renders.
+  // Named here rather than through a helper: this block keeps its own handler
+  // and its own log prefix, which the identity work is grouped under.
+  startup.currentStep = 'mirroring trade key indices';
   try {
     _mirrorTradeKeyIndex(await identity_api.onTradeKeyIndexChanged());
   } catch (e) {
@@ -192,6 +210,8 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
 
   // Initialize identity: creates on first launch, reloads on subsequent launches.
   // Must run before Nostr init so the identity key is available for relay auth.
+  // Named here rather than through a helper, for the same reason as above.
+  startup.currentStep = 'loading your identity';
   try {
     await IdentityService.initialize();
   } catch (e, st) {
@@ -204,7 +224,7 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
   // Tokio broadcast channel buffers any notice arriving during startup rather
   // than dropping it (a receiver must exist at send time).
   bond_api.BondSlashedStream? bondSlashedStream;
-  await _optional('subscribing to bond notices', () async {
+  await startup.optional('subscribing to bond notices', () async {
     bondSlashedStream = await bond_api.onBondSlashed();
   });
 
@@ -215,12 +235,12 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
   // connection state, and Settings can switch node or edit the relay list. An
   // app that opens offline can be fixed from inside; one that does not open
   // cannot be fixed at all.
-  await _optional('connecting to the network', () async {
+  await startup.optional('connecting to the network', () async {
     await nostr_api.initialize(relays: seedRelays.isEmpty ? null : seedRelays);
   });
 
   // Log initial relay state for diagnostics.
-  await _optional('reading relay status', () async {
+  await startup.optional('reading relay status', () async {
     final relays = await nostr_api.getRelays();
     final connState = await nostr_api.getConnectionState();
     debugPrint(
@@ -228,35 +248,46 @@ Future<void> _startup({List<String> seedRelays = const []}) async {
     );
   });
 
-  // Watch for connection state changes in background (logs appear in flutter output).
-  _watchConnectionState();
+  // Assembling the container, restoring the wallet and starting the watchers
+  // are one stretch with no natural place to stop. Unlabelled, they ran under
+  // the name of whichever optional step finished last, so a failure here named
+  // a step that had already succeeded (#405 review).
+  //
+  // runApp stays outside the wrapper: the label is still 'building the
+  // interface' when it runs, so it is already covered, and the guard sits above
+  // both either way. Purely so this reads as one statement rather than a
+  // closure with the whole tail inside it.
+  final container = await startup.required('building the interface', () async {
+    // Logs every relay connection state change (debug builds only).
+    _watchConnectionState();
 
-  final container = ProviderContainer(
-    overrides: [
-      firstRunProvider.overrideWith(
-        (ref) => FirstRunNotifier(initialValue: firstRunComplete),
-      ),
-      backupReminderProvider.overrideWith(
-        (ref) => BackupReminderNotifier(initialValue: backupPending),
-      ),
-      settingsProvider.overrideWith(
-        (ref) => SettingsNotifier(prefs: prefs, initial: savedSettings),
-      ),
-      nwcProvider.overrideWith((ref) => NwcNotifier(prefs: prefs)),
-      mostroPubkeyProvider.overrideWith((ref) => activeMostroPubkey),
-    ],
-  );
+    final container = ProviderContainer(
+      overrides: [
+        firstRunProvider.overrideWith(
+          (ref) => FirstRunNotifier(initialValue: firstRunComplete),
+        ),
+        backupReminderProvider.overrideWith(
+          (ref) => BackupReminderNotifier(initialValue: backupPending),
+        ),
+        settingsProvider.overrideWith(
+          (ref) => SettingsNotifier(prefs: prefs, initial: savedSettings),
+        ),
+        nwcProvider.overrideWith((ref) => NwcNotifier(prefs: prefs)),
+        mostroPubkeyProvider.overrideWith((ref) => activeMostroPubkey),
+      ],
+    );
 
-  // Restore NWC wallet connection if a URI was saved from a previous session.
-  final savedNwcUri = prefs.getString(kNwcUriKey);
-  if (savedNwcUri != null) {
-    _restoreNwcConnection(savedNwcUri, container);
-  }
+    // Restore NWC wallet connection if a URI was saved from a previous session.
+    final savedNwcUri = prefs.getString(kNwcUriKey);
+    if (savedNwcUri != null) {
+      _restoreNwcConnection(savedNwcUri, container);
+    }
 
-  final slashed = bondSlashedStream;
-  if (slashed != null) _consumeBondSlashed(slashed, container);
+    final slashed = bondSlashedStream;
+    if (slashed != null) _consumeBondSlashed(slashed, container);
+    return container;
+  });
 
-  _currentStep = 'building the interface';
   runApp(
     UncontrolledProviderScope(container: container, child: const MostroApp()),
   );
--- a/lib/core/app_bootstrap.dart
+++ b/lib/core/app_bootstrap.dart
@@ -10,6 +10,7 @@ import 'package:mostro/core/storage/app_data_dir.dart'
     if (dart.library.html) 'package:mostro/core/storage/app_data_dir_web.dart';
 import 'package:mostro/core/app.dart';
 import 'package:mostro/core/mostro_defaults.dart';
+import 'package:mostro/core/startup_failure.dart';
 import 'package:mostro/core/services/identity_service.dart';
 import 'package:mostro/core/test_environment.dart';
 import 'package:mostro/core/web/bridge_probe.dart';
@@ -44,24 +45,68 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
 /// defaults gone, an unreachable local relay fails the test instead of
 /// silently succeeding against a public one.
 Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
+  try {
+    await _startup(seedRelays: seedRelays);
+  } catch (e, st) {
+    // The failure surface calls runApp too, and that is the whole fix: today an
+    // exception in here means runApp never runs and Flutter paints nothing —
+    // the page is not broken, it is absent, with no message anywhere (#227).
+    debugPrint('[startup] fatal while $_currentStep: $e\n$st');
+    runApp(StartupFailureApp(step: _currentStep));
+  }
+}
+
+/// Names the startup step in progress, so a failure can say where it happened.
+///
+/// A file-level variable rather than a parameter: the only writers are the
+/// steps themselves and the only reader is the guard above.
+String _currentStep = 'starting up';
+
+/// Runs a startup step the app can do without: a failure is recorded and
+/// startup continues, so the app opens degraded rather than not at all.
+///
+/// One helper rather than a try/catch per step, so every degradation prints the
+/// same prefix — grepping `[startup]` lists everything a run gave up on, in
+/// order, which matters when one failure is the cause of the next.
+Future<void> _optional(String name, Future<void> Function() body) async {
+  _currentStep = name;
+  try {
+    await body();
+  } catch (e, st) {
+    debugPrint('[startup] $name failed — continuing without it: $e\n$st');
+  }
+}
+
+Future<void> _startup({List<String> seedRelays = const []}) async {
   WidgetsFlutterBinding.ensureInitialized();
 
-  // Initialize Firebase (no-op if firebase_options.dart is the placeholder).
+  // Push notifications only — the app trades, chats and settles without them.
+  // Two arms on purpose: the placeholder config is an expected state, not a
+  // failure, and folding both into one message would make every single run log
+  // a "failed" nobody reads by the time it means something.
+  _currentStep = 'setting up notifications';
   try {
     await Firebase.initializeApp(
       options: DefaultFirebaseOptions.currentPlatform,
     );
   } on UnsupportedError catch (e) {
     debugPrint(
-      '[main] Firebase not configured: $e — push notifications disabled.',
+      '[startup] Firebase not configured — push notifications disabled: $e',
+    );
+  } catch (e, st) {
+    // A call into a third-party JS SDK: config, network, or the SDK itself.
+    debugPrint(
+      '[startup] setting up notifications failed — continuing without it: $e\n$st',
     );
   }
 
+  _currentStep = 'loading the engine';
   await RustLib.init();
 
   // Pre-read SharedPreferences so providers start with synchronous initial
   // values — eliminates the AsyncValue.loading() race that caused the router
   // to show the home screen before redirecting to /walkthrough on first launch.
+  _currentStep = 'reading your settings';
   final prefs = await SharedPreferences.getInstance();
   final firstRunComplete = prefs.getBool(kFirstRunCompleteKey) ?? false;
   final backupDismissed = prefs.getBool(kBackupReminderDismissedKey) ?? false;
@@ -71,23 +116,27 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
 
   // Before any startup work below, so a failure in it is captured at the
   // verbosity the user asked for rather than the default.
-  await settings_api.setLoggingEnabled(enabled: savedSettings.loggingEnabled);
+  await _optional('applying your log settings', () async {
+    await settings_api.setLoggingEnabled(enabled: savedSettings.loggingEnabled);
+  });
 
-  // Initialize the persistent store (SQLite file off the web, IndexedDB
-  // database on it). Must come before any trade / order operations that read
-  // or write trade keys and trade records.
-  try {
+  // The persistent store (a SQLite file off the web, an IndexedDB database on
+  // it, #408). Must come before any trade / order operation that reads or
+  // writes trade keys and trade records.
+  //
+  // Optional, and it was already written that way before this guard existed:
+  // without it the session is memory-only — trade keys and roles do not
+  // survive a restart — but orders still browse and relay messages still
+  // arrive, and every Rust caller handles a missing database. Runs on the web
+  // too: since #408 that is where web persistence lives, so skipping it there
+  // would quietly take out every feature built on top of it.
+  await _optional('opening the local database', () async {
     final location = databaseLocation(
       isWeb: kIsWeb,
       dataDir: kIsWeb ? null : await appDataDirPath(),
     );
     await rust_api.initDb(path: location);
-  } catch (e, st) {
-    // DB init failure is non-fatal: trade-key and role persistence won't
-    // work for this session, but the app can still browse orders and relay
-    // messages.  All Rust callers already handle db() == None gracefully.
-    debugPrint('[main] DB init failed — running in memory-only mode: $e\n$st');
-  }
+  });
 
   // Load the persisted active Mostro node into the Rust override before the
   // relay pool starts, so the first subscription targets the user's selected
@@ -154,19 +203,30 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
   // Subscribe to bond-slashed notices BEFORE relay delivery starts, so the
   // Tokio broadcast channel buffers any notice arriving during startup rather
   // than dropping it (a receiver must exist at send time).
-  final bondSlashedStream = await bond_api.onBondSlashed();
+  bond_api.BondSlashedStream? bondSlashedStream;
+  await _optional('subscribing to bond notices', () async {
+    bondSlashedStream = await bond_api.onBondSlashed();
+  });
 
   // Initialize the Nostr relay pool. `null` means the compiled-in defaults
   // (config.rs); a non-empty seed list replaces them entirely.
   // This must happen before any Nostr/order API calls.
-  await nostr_api.initialize(relays: seedRelays.isEmpty ? null : seedRelays);
+  // Optional because the app already knows how to be disconnected: it shows
+  // connection state, and Settings can switch node or edit the relay list. An
+  // app that opens offline can be fixed from inside; one that does not open
+  // cannot be fixed at all.
+  await _optional('connecting to the network', () async {
+    await nostr_api.initialize(relays: seedRelays.isEmpty ? null : seedRelays);
+  });
 
   // Log initial relay state for diagnostics.
-  final relays = await nostr_api.getRelays();
-  final connState = await nostr_api.getConnectionState();
-  debugPrint(
-    '[main] relay pool initialized — state=$connState relays=${relays.map((r) => '${r.url}:${r.status}').join(', ')}',
-  );
+  await _optional('reading relay status', () async {
+    final relays = await nostr_api.getRelays();
+    final connState = await nostr_api.getConnectionState();
+    debugPrint(
+      '[main] relay pool initialized — state=$connState relays=${relays.map((r) => '${r.url}:${r.status}').join(', ')}',
+    );
+  });
 
   // Watch for connection state changes in background (logs appear in flutter output).
   _watchConnectionState();
@@ -193,8 +253,10 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
     _restoreNwcConnection(savedNwcUri, container);
   }
 
-  _consumeBondSlashed(bondSlashedStream, container);
+  final slashed = bondSlashedStream;
+  if (slashed != null) _consumeBondSlashed(slashed, container);
 
+  _currentStep = 'building the interface';
   runApp(
     UncontrolledProviderScope(container: container, child: const MostroApp()),
   );
--- a/lib/core/startup_failure.dart
+++ b/lib/core/startup_failure.dart
@@ -20,36 +20,44 @@ class StartupFailureApp extends StatelessWidget {
   Widget build(BuildContext context) {
     return MaterialApp(
       debugShowCheckedModeBanner: false,
-      home: Scaffold(
-        // Hard-coded rather than taken from the app theme: the theme is built
-        // from settings this screen exists to survive the loss of.
-        backgroundColor: const Color(0xFF1D212C),
-        body: Center(
-          child: Padding(
-            padding: const EdgeInsets.all(32),
-            child: Column(
-              mainAxisSize: MainAxisSize.min,
-              children: [
-                const Text(
-                  'Mostro could not start',
-                  textAlign: TextAlign.center,
-                  style: TextStyle(
-                    color: Colors.white,
-                    fontSize: 20,
-                    fontWeight: FontWeight.w600,
-                  ),
+      // Every route resolves here rather than using `home:`. On the web the
+      // initial route comes from the browser's URL, and a startup failure is
+      // most often met on a reload of some deep path — which `home:` alone
+      // answers with "Could not navigate to initial route" in the console
+      // before falling back. Harmless, but this screen exists to make a failed
+      // startup legible; it should not add noise of its own to the one log the
+      // person reporting it is about to read.
+      onGenerateRoute: (_) => MaterialPageRoute<void>(builder: _buildBody),
+    );
+  }
+
+  Widget _buildBody(BuildContext context) {
+    return Scaffold(
+      // Hard-coded rather than taken from the app theme: the theme is built
+      // from settings this screen exists to survive the loss of.
+      backgroundColor: const Color(0xFF1D212C),
+      body: Center(
+        child: Padding(
+          padding: const EdgeInsets.all(32),
+          child: Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              const Text(
+                'Mostro could not start',
+                textAlign: TextAlign.center,
+                style: TextStyle(
+                  color: Colors.white,
+                  fontSize: 20,
+                  fontWeight: FontWeight.w600,
                 ),
-                const SizedBox(height: 12),
-                Text(
-                  'It failed while $step.',
-                  textAlign: TextAlign.center,
-                  style: const TextStyle(
-                    color: Color(0xFFB0B6C3),
-                    fontSize: 15,
-                  ),
-                ),
-              ],
-            ),
+              ),
+              const SizedBox(height: 12),
+              Text(
+                'It failed while $step.',
+                textAlign: TextAlign.center,
+                style: const TextStyle(color: Color(0xFFB0B6C3), fontSize: 15),
+              ),
+            ],
           ),
         ),
       ),
--- a/lib/core/startup_failure.dart
+++ b/lib/core/startup_failure.dart
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 
-/// Shown when startup cannot reach `runApp`.
+/// Painted instead of the app when startup fails before it can be.
 ///
 /// Deliberately dependency-free: no localization, no app theme, no Rust, no
 /// SharedPreferences. Any of those can be the thing that failed, and a rescue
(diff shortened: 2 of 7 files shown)


grunch:
This branch conflicts with `main` since #408 (`d49f3c1`) merged; both edit `lib/core/app_bootstrap.dart`.

What to do on rebase:

- #408 replaced the `if (!kIsWeb) { … initDb(path: p.join(dataDir, 'mostro.db')) … }` block with an unconditional `initDb` call whose argument comes from `databaseLocation(isWeb: kIsWeb, dataDir: …)` in `lib/core/storage/db_location.dart`: on the web the store is now opened under a fixed IndexedDB name, off the web under the data directory as before. It also dropped the then-unused `package:path/path.dart` import. Keep that shape when you fold the store initialisation into your guarded startup steps: it should be one named step like the others, still non-fatal (the app can browse without persistence), and it must run on the web too, because every other web feature that persists anything depends on it now.
- `test/core/storage/db_location_test.dart` covers the location helper; a startup-guard test that names the failing step would be the natural companion.
- Re-run `flutter analyze && flutter test` after resolving. CI pins Flutter 3.38.2, so avoid matchers newer than that (`isSemantics` is one; `containsSemantics` works on both).

Matobi98:
Rebased onto `d49f3c1`, pushed as `d89fb01`.

The three conflicts were formatting — #408 rewrapped lines this branch had also edited. The database step didn't conflict at all, which is why the warning was worth giving: it merged in silently and would have shipped as the only step outside the sequence. It's now `_optional('opening the local database', …)`, unconditional, with a comment saying why the `!kIsWeb` guard is gone.

`test/core/startup_guard_test.dart` pins that statically — it fails if `initDb` ends up back inside a `!kIsWeb` block, mutation-checked.

Flutter version: fixed at the root rather than remembered — the other SDK is gone from this machine, so `flutter` is now whatever `.fvmrc` pins.

Verified after the rebase: analyze clean, 351 Flutter tests, 438 Rust tests, clippy and wasm clean. Manually in Chrome against the local regtest: app opens, a trade survives a reload, and breaking the database leaves the app up with its log line.

One aside: `.githooks/pre-commit` calls bare `flutter`, so on a machine using fvm it fails `analyze` on deprecations the pinned version doesn't have and blocks every commit. Worked around locally — say the word and I'll send a one-line fix.

lib/core/app_bootstrap.dart
```diff
@@ -44,24 +45,68 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
...
+/// steps themselves and the only reader is the guard above.
+String _currentStep = 'starting up';
+
+/// Runs a startup step the app can do without: a failure is recorded and
+/// startup continues, so the app opens degraded rather than not at all.
+///
+/// One helper rather than a try/catch per step, so every degradation prints the
+/// same prefix — grepping `[startup]` lists everything a run gave up on, in
+/// order, which matters when one failure is the cause of the next.
+Future<void> _optional(String name, Future<void> Function() body) async {
```
grunch: `_currentStep = name` is never restored. After this helper returns successfully, everything up to the next explicit assignment runs under the label of a step that already finished: the rehydrate/identity block as *opening the local database*, and `ProviderContainer(...)`, the NWC URI read and `_watchConnectionState()` as *reading relay status*. A throw there produces exactly the screen the PR calls worthless — one naming a step that did not fail.

Suggest a `_required(String name, Future<T> Function() body)` twin (sets the label, awaits, rethrows) and using it for `RustLib.init`, the preferences read and the container build, so no code runs "between labels". Making both helpers methods of a tiny `StartupSequence` class would also let the guard be unit-tested with fake bodies instead of source greps.

lib/core/app_bootstrap.dart
```diff
@@ -44,24 +45,68 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
 /// defaults gone, an unreachable local relay fails the test instead of
 /// silently succeeding against a public one.
 Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
+  try {
+    await _startup(seedRelays: seedRelays);
+  } catch (e, st) {
+    // The failure surface calls runApp too, and that is the whole fix: today an
+    // exception in here means runApp never runs and Flutter paints nothing —
+    // the page is not broken, it is absent, with no message anywhere (#227).
```
grunch: #227 is not an example of what this catch handles. #370 pinned it inside the engine's `CanvasKitRenderer.initialize`, before `main()` runs, and states that no app-level try/catch around `main()` can guard it. Cite it as the motivating precedent, not as a case covered here (same wording in the PR body and in `startup_guard_test.dart:18`).

lib/core/app_bootstrap.dart
```diff
@@ -44,24 +45,68 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
...
+    debugPrint('[startup] fatal while $_currentStep: $e\n$st');
+    runApp(StartupFailureApp(step: _currentStep));
```
grunch: Consider `markBridgeFailed(e);` before `runApp` here. It is a no-op off web, and on web it lets `test/web/smoke/smoke.mjs` fail immediately with the cause instead of waiting for the bridge-ready timeout.

lib/core/startup_failure.dart
```diff
@@ -0,0 +1,66 @@
...
+    return MaterialApp(
+      debugShowCheckedModeBanner: false,
+      // Every route resolves here rather than using `home:`. On the web the
+      // initial route comes from the browser's URL, and a startup failure is
+      // most often met on a reload of some deep path — which `home:` alone
+      // answers with "Could not navigate to initial route" in the console
+      // before falling back. Harmless, but this screen exists to make a failed
+      // startup legible; it should not add noise of its own to the one log the
+      // person reporting it is about to read.
+      onGenerateRoute: (_) => MaterialPageRoute<void>(builder: _buildBody),
```
grunch, line 48: Verified with `defaultRouteNameTestValue = '/orders/abc/detail'`: `Navigator.defaultGenerateInitialRoutes` resolves `/`, `/orders`, `/orders/abc`, `/orders/abc/detail` through this callback and pushes all four, so the failure page is stacked 4 deep and `canPop()` is true — browser/Android back pops to an identical page.

Also, the "Could not navigate to initial route" report this comment cites is inside an `assert(...)` in `navigator.dart`, so it never appears in a release build; the motivation only holds for debug.

```dart
onGenerateInitialRoutes: (_) => [MaterialPageRoute<void>(builder: _buildBody)],
```

yields exactly one route for any initial URL.

test/core/startup_failure_test.dart
```diff
@@ -0,0 +1,53 @@
...
+
+    testWidgets('says which step, not just that one failed', (tester) async {
+      await tester.pumpWidget(
+        const StartupFailureApp(step: 'reading your settings'),
+      );
+      expect(find.textContaining('reading your settings'), findsOneWidget);
+      expect(find.textContaining('loading the engine'), findsNothing);
+    });
+
+    testWidgets('answers a deep initial route, not just "/"', (tester) async {
```
grunch: This test never sets a deep route — `defaultRouteNameTestValue` is `/` by default, and wrapping in `MediaQuery` does not change it — so it is the fourth test again and passes with `home:` too. To test what the name says:

```dart
tester.binding.platformDispatcher.defaultRouteNameTestValue = '/orders/abc/detail';
addTearDown(tester.binding.platformDispatcher.clearDefaultRouteNameTestValue);
await tester.pumpWidget(const StartupFailureApp(step: 'loading the engine'));
expect(tester.takeException(), isNull);
expect(tester.state<NavigatorState>(find.byType(Navigator)).canPop(), isFalse);
```

With the current `onGenerateRoute` the last expectation fails (4 stacked routes); with `onGenerateInitialRoutes` it passes.

test/core/startup_guard_test.dart
```diff
@@ -0,0 +1,83 @@
...
+      // file — would quietly take out every web feature built on top of it.
+      final initDbAt = source.indexOf('rust_api.initDb(');
+      expect(
+        initDbAt,
+        greaterThanOrEqualTo(0),
+        reason: 'the database is never initialised',
+      );
+
+      final before = source.substring(0, initDbAt);
+      final enclosingWebGuard =
```
grunch: Named `enclosingWebGuard` but it does not check enclosure: it matches any `if (!kIsWeb) {` anywhere before `initDb`, so a legitimate native-only step added earlier in the file fails this test with a message about the database. Either scope the search to the `_optional('opening the local database'` block, or drop this in favour of a behavioural test on an extracted startup sequence.

grunch (changes requested):
Strict review. Verified on the branch (`d89fb01`): `flutter analyze` adds no new diagnostics, the 8 new tests pass. The direction is right and the change is small and well reasoned, but two findings contradict the PR's own bar ("a screen that named the same step regardless would pass a careless look and be worthless"), so requesting changes.

**HIGH — the step label goes stale after a successful optional step.** `_optional` sets `_currentStep = name` and never restores it. The node rehydration block, the trade-key mirror and `IdentityService.initialize` all run labelled *opening the local database*; `ProviderContainer(...)`, the NWC URI read and `_watchConnectionState()` run labelled *reading relay status* — a step that already finished fine. A throw there names a step that did not fail. Structural fix: a `_required(name, body)` twin for the mandatory stretches, so every stretch carries its own label and there is no code "between labels". (inline)

**HIGH — the "deep initial route" test sets no deep route, and the real behaviour on one is wrong.** The test is identical to the fourth one (`defaultRouteNameTestValue` is `/` by default). I ran it for real with `/orders/abc/detail`: the Navigator stacks **4** copies of the failure page and `canPop()` is true, so browser/Android back "navigates" to a clone. Also the "Could not navigate to initial route" message that motivates the change lives inside an `assert`, so it only ever exists in debug builds. Fix with `onGenerateInitialRoutes` and a test that actually sets the route. (inline)

**MEDIUM — #227 is cited as a case this guard catches. It is not.** #370 pinned it inside `CanvasKitRenderer.initialize`, before `main()` runs, and says so: "no app-level try/catch around main() can guard it". #227 is the precedent that motivates the guard, not a case it covers. Reword the comment in `bootstrapAndRun` and the PR body.

**MEDIUM — the guard logic has zero behavioural coverage.** All four tests in `startup_guard_test.dart` are greps over the source. Nothing checks that a failed optional step continues, that a failed required step reaches the screen, or that the reported step is the right one (exactly the bug above). Extract a small class with `optional`, `required` and `currentStep`, call it from `bootstrapAndRun`, and test it with fake bodies. The `kIsWeb` grep is also not an "enclosing" check (inline).

**MEDIUM — the screen shows the step but not the error.** On Android/iOS/Linux the user has no console. A secondary, selectable line with `error.toString()` makes the report actionable without adding a dependency. Suggestion, not blocking.

**LOW**
- If `WidgetsFlutterBinding.ensureInitialized()` throws, the rescue `runApp` throws too. Acceptable edge case, but worth a comment or moving that call outside the `try`.
- The description drifted after round 1: "3 new tests" (there are 8), the seven-step table omits the database step, and "no `kIsWeb` anywhere in the change" is no longer true. It becomes the merge reference, so fix it before merging.
- Hard-coded English is justified here but contradicts CLAUDE.md ("all user-facing strings are Dart-level l10n"). Add one line under *Translations* recording the exception so nobody "fixes" it.
- The last-resort catch does not call `markBridgeFailed(e)`; with it the web smoke test fails fast with the cause instead of timing out. (inline)

Matobi98:
Round 2 — pushed in `159af5a`. All five findings and the four low ones taken; three more turned up on the way.

**HIGH — the stale label.** Correct, and the fix is the `_required` twin you suggested, as methods on a small `StartupSequence` (`lib/core/startup_sequence.dart`) so the guard is unit-testable. Every stretch now announces its own name; the three blocks that keep bespoke handlers — one reports to the bridge probe — set it directly, with a line at each site saying why they don't go through a helper.

Verified in the app, not only in tests: breaking the container assembly used to produce "It failed while reading relay status". It now says "building the interface".

**HIGH — the deep initial route.** Reproduced your measurement before changing anything: `/orders/abc/detail` stacked 4 copies, `canPop()` true. `onGenerateInitialRoutes` gives 1 and `canPop()` false. `onGenerateRoute` has to stay — MaterialApp refuses to construct without one — so it's an addition, not a replacement, and there's a comment saying so. The test now sets `defaultRouteNameTestValue` and asserts on the navigator stack; the old one was indeed the fourth test again.

You were also right that the console message lived in an `assert`. The comment justified the change with debug-only noise; it now states the real reason.

**MEDIUM — #227 miscited.** Ours to have caught: #370 is where we established the crash fires before `main()`. Reworded in `app_bootstrap.dart` and the test, and in the PR body.

**MEDIUM — no behavioural coverage.** Six tests now drive `StartupSequence` with fake bodies, including the one that would have caught the stale label: an optional step succeeds, a required one throws, and the reported step must be the second. Mutation-checked — dropping the label from `required` reddens two of them.

Two greps remain, and only because running is not available: that the guard reaches `runApp` with the current step, and the `!kIsWeb` check, now scoped to the `opening the local database` block as you asked rather than searching the whole file.

**MEDIUM — the cause on screen.** Added, selectable, capped at 300 characters. Plus a "Copy details" button, since on a phone dragging to select inside a scrolling view is awkward: it puts the step *and* the cause on the clipboard, because the cause alone loses half of what makes a report actionable.

**LOW.** `ensureInitialized` moved outside the guard, with a comment saying it is honestly unguarded — the rescue paints through `runApp`, which needs it too. `markBridgeFailed(e)` added. Description rewritten: the counts were stale, the table was missing the database step, and "no `kIsWeb` anywhere" had stopped being true. `CLAUDE.md` records the hard-coded English as a deliberate exception.

## Three things found while doing this

**1. A long cause on a short screen overflowed** — the striped warning painted over the one screen that has to stay readable. My own addition, an hour old. The content scrolls now, with a test at 320×400.

**2. `markBridgeFailed` was sitting in front of `runApp` in the catch**, unguarded. If it threw there would be no rescue at all. The screen goes first now; CI gets told after.

**3. `[startup]` degradations are not visible anywhere a user can reach.** The design leans on "grep `[startup]` to see what a run gave up on", but those lines are `debugPrint`; the in-app log viewer shows Rust's logs, and Dart has no way to write into it. So the promise only holds with a dev console attached. Not this PR's to fix, and not a regression — but worth knowing before someone relies on it in a bug report.

Two more I did not act on, deliberately:

- `optional` catches everything, so a null-check or type error in an optional step is swallowed and startup continues in an unknown state. Narrowing it to exceptions would turn those into hard failures — a behaviour change with real risk, and one that predates this PR. Worth a decision by someone who knows the project better than I do.
- If `runApp` throws, the guard calls `runApp` again. Nearly unreachable, since build errors surface in the next frame and never reach this catch, but the path exists and nothing exercises it.

## Verified

`flutter analyze` clean, `flutter test` 360 passed, `cargo test` / clippy / `cargo check --target wasm32-unknown-unknown` clean.

Manually against the local regtest in Chrome, seven cases: the app opens; a trade survives a reload; the database broken leaves the app up with its log line and persistence gone; `RustLib.init` and `SharedPreferences` broken produce two *different* screens; the container assembly broken says "building the interface"; and a deep initial route leaves one page with back not landing on a clone.

One of those runs turned up something unplanned — a page reload produced `AnyhowException(AlreadyInitialized)` from the relay pool, which the guard logged and continued past. Before this PR that line was unwrapped.

Matobi98:
Merged `main` and re-ran the manual round on the result. Three corrections to what I wrote earlier:

- **The Firebase comment was wrong.** It said `currentPlatform` throws `UnsupportedError` on every run. Only Linux throws; web, Android, iOS, macOS and Windows return options. Fixed in the code comment.
- **"An app that opens offline can be fixed from inside" is false.** Measured with relay init forced to fail: Settings opens, but adding a relay fails, because the relay API goes through a pool that never started. It stays optional — the app opens and the secret words can still be backed up — but the comment now says what actually works and what does not.
- **"Before this round the container failure said 'reading relay status'"** I confirmed by reading the round-1 code (the label only changed after the container was built), not by running it.
- **"A page reload produced `AnyhowException(AlreadyInitialized)`"** — it was a hot restart, not a reload. A hot restart keeps the Rust side alive, so the relay pool is already initialised; a real reload starts it fresh. Dropped from the description.

The merge also brought two steps, both optional: `registering font licences` (it was left outside the guard by the merge, moved inside) and `subscribing to bond claim updates`. Description updated.

lib/core/app_bootstrap.dart
```diff
@@ -52,51 +54,114 @@ import 'package:mostro/features/notifications/providers/notifications_provider.d
...
+    // stated that no app-level try/catch could reach it.
+    debugPrint('[startup] fatal while ${startup.currentStep}: $e\n$st');
+    // The screen first: it is what a person is waiting for, and it is the whole
+    // point of this catch. Anything ahead of it that could throw would leave
+    // them with the blank page this exists to replace.
+    runApp(StartupFailureApp(step: startup.currentStep, error: e));
+    // Then CI. No-op off web; on web it hands test/web/smoke/smoke.mjs the
+    // cause, so the run stops with a reason instead of timing out waiting for
+    // a bridge that is never coming.
+    markBridgeFailed(e);
```
grunch: _🧪 CI Observability_ | _🟠 Major_

**A fatal startup failure after `markBridgeReady()` now passes the web smoke test.**

`markBridgeReady()` fires inside `selecting the Mostro node` (line 206). After it, `loading your identity`, `connecting to the network` and the `required` step `building the interface` still run. Following `test/web/smoke/smoke.mjs` for a throw in `building the interface`:

- **Stage 3** (lines 358–373) resolves `waitForFunction` as soon as `mostroBridgeReady === true`, then reads `mostroBridgeError` **once**. Identity init and relay-pool init sit between those two points and take far longer than that read. The error flag this guard sets later is never seen. Only stage 3b's post-reload check would see it, and 3b is opt-in (`SMOKE_BOND_STORE=1`).
- **Final checks** (lines 261–264, 443) fail on `console` messages of type `error` and on `pageerror`. Before this PR the throw escaped `main()` and reached them. Now it is caught, printed with `debugPrint` (a `console.log`, not an error) and turned into `StartupFailureApp`.

The result is a green smoke run on a bundle that opens on "Mostro could not start: building the interface". The PR fixes this failure for users, but it also hides it from CI, which used to catch it.

Two small options:

```dart
} catch (e, st) {
  runApp(StartupFailureApp(step: startup.currentStep, error: e));
  markBridgeFailed(e);
  // Keep it observable as an error: console.error on web, error log elsewhere.
  FlutterError.reportError(FlutterErrorDetails(
    exception: e, stack: st, library: 'startup',
    context: ErrorDescription('while ${startup.currentStep}'),
  ));
}
```

or publish a separate `mostroStartupFailed` global, and have the smoke test wait for `MostroApp`'s first frame (e.g. a known semantics label) before it declares success. The first keeps the change local; the second closes the gap for good.

<details><summary>🤖 Prompt for AI Agents</summary>

```
In lib/core/app_bootstrap.dart bootstrapAndRun catch block, after runApp(StartupFailureApp(...)) and markBridgeFailed(e), report the error through FlutterError.reportError (or console.error on web) so test/web/smoke/smoke.mjs's console/pageerror collectors fail the run. Optionally extend smoke.mjs stage 3 to re-read mostroBridgeError after the app has rendered.
```
</details>

test/core/startup_guard_test.dart
```diff
@@ -0,0 +1,66 @@
...
+    test('a failure reaches runApp, naming the step that failed', () {
+      // Without this the sequence throws, runApp never runs, and Flutter paints
+      // nothing — not a broken page, an absent one. #227 is the precedent that
+      // motivated the guard, not a case it covers: that crash happens inside
+      // the engine's own bootstrap, before main() runs (#370).
+      // Two independent substrings rather than one exact call: the literal
+      // broke the first time an argument was added to it, which is the failure
+      // mode of a grep and the reason only two are left in this file.
+      expect(
+        source.contains('runApp(StartupFailureApp('),
```
grunch: _🧪 Test Quality_ | _🟠 Major_

**The guard is the fix, and it is the one part without a behavioural test.**

`source.contains('runApp(StartupFailureApp(')` and `contains('step: startup.currentStep')` still pass when:
- the call moves out of the `catch` (into an unreachable branch, or a helper that is never called);
- a second `catch` above it swallows the error first;
- the order flips so `markBridgeFailed` runs first and throws;
- someone wraps `_startup` in `unawaited(...)`, so the `try` never catches anything.

Reaching `bootstrapAndRun` needs Rust, preferences and relays, but the guard doesn't. Extracting it takes a few lines:

```dart
@visibleForTesting
Future<void> runGuarded(
  Future<void> Function(StartupSequence) body, {
  void Function(Widget) run = runApp,
  void Function(Object) onFailed = markBridgeFailed,
}) async {
  final startup = StartupSequence();
  try {
    await body(startup);
  } catch (e, st) { /* same as today */ }
}
```

`bootstrapAndRun` becomes `runGuarded((s) => _startup(s, seedRelays: seedRelays))`. A test can then use a fake body that throws inside `required('building the interface', …)` and check that `run` received a `StartupFailureApp` with that step and error, and that `onFailed` ran after it. That covers all four cases above. The `kIsWeb` grep can stay.

lib/core/app_bootstrap.dart
```diff
@@ -174,55 +247,93 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
...
+  // Assembling the container, restoring the wallet and starting the watchers
+  // are one stretch with no natural place to stop. Unlabelled, they ran under
+  // the name of whichever optional step finished last, so a failure here named
+  // a step that had already succeeded (#405 review).
+  //
+  // runApp stays outside the wrapper: the label is still 'building the
+  // interface' when it runs, so it is already covered, and the guard sits above
+  // both either way. Purely so this reads as one statement rather than a
+  // closure with the whole tail inside it.
+  final container = await startup.required('building the interface', () async {
```
grunch, line 294: _🎯 Correctness of the claim_ | _🟡 Minor_

**`runApp` is only partly covered.**

The comment above says `runApp` *"is already covered"* because the label is still set. But the `try` in `bootstrapAndRun` only catches what `runApp` throws **synchronously** (binding and attach errors). A throw in `MostroApp.build`, in a provider read during the first build, or in `GoRouter`'s initial redirect happens later, in the frame pipeline. It goes to `FlutterError.onError` (red screen in debug, grey box in release) and never reaches this catch.

That scope is acceptable, but the comment and the table row ("building the interface → nothing to show") should say so, so nobody reads the guard as covering first-frame errors. If you want it covered: install a `FlutterError.onError` for the first frame only. It calls `runApp(StartupFailureApp(step: 'showing the first screen', …))`, and you remove it in `addPostFrameCallback`.

lib/core/app_bootstrap.dart
```diff
@@ -174,55 +247,93 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
...
-  // Resume = resync in Rust, then re-hydrate every notifier from the bridge
-  // (issue #308, docs/PUSH_NOTIFICATIONS.md §10). Attached before runApp so
-  // the first suspension is observed too.
-  AppLifecycleService(
-    onResume: ResumeResync(container: container).run,
-  ).attach();
+    // Resume = resync in Rust, then re-hydrate every notifier from the bridge
+    // (issue #308, docs/PUSH_NOTIFICATIONS.md §10). Attached before runApp so
+    // the first suspension is observed too.
+    AppLifecycleService(
```
grunch, line 345: _♻️ Resource lifecycle_ | _🟡 Minor_

**After a failure past this point, listeners outlive the failed startup.**

Today only the `return` follows `attach()`, but this stretch is the one likely to grow. If anything after `attach()` throws, or `runApp` below throws synchronously, the guard paints `StartupFailureApp` while these keep running:
- `AppLifecycleService`, whose `onResume` runs `ResumeResync` against a container that is never mounted;
- the `_consumeBondSlashed` / `_consumeBondClaims` loops;
- `_restoreNwcConnection`'s microtask.

The user sees "could not start", yet a background resume still calls into Rust and the notifiers. Any resulting async errors land unhandled in the log and bury the real cause.

Either attach the lifecycle service and start the consumers as the last thing before `runApp`, once the container is fully built, or keep handles and call `detach()` / `container.dispose()` in the guard's `catch`.

lib/core/startup_failure.dart
```diff
@@ -0,0 +1,134 @@
...
+                ),
+              ],
+              const SizedBox(height: 24),
+              // A tap rather than a text selection: on a phone there is no
+              // console and dragging to select inside a scrolling view is
+              // fiddly, so the one action this screen offers is handing the
+              // whole report over in a form that can be pasted into a message.
+              TextButton.icon(
+                onPressed: () {
+                  Clipboard.setData(ClipboardData(text: _report()));
```
grunch: _🎯 Robustness_ | _🟡 Minor_

**"Copied" is shown even when nothing was copied.**

`Clipboard.setData` returns a `Future` that is not awaited. On the web it goes through `navigator.clipboard.writeText`, which rejects outside a secure context, when permission is denied, and in some embedded webviews. When it rejects:
- the snackbar still says **Copied**, and the person pastes an empty report;
- the rejection becomes an unhandled async error on the one screen that should stay quiet.

The screen's only action should report what actually happened:

```dart
onPressed: () async {
  final messenger = ScaffoldMessenger.of(context);
  try {
    await Clipboard.setData(ClipboardData(text: _report()));
    messenger.showSnackBar(const SnackBar(content: Text('Copied')));
  } catch (_) {
    messenger.showSnackBar(const SnackBar(
      content: Text('Could not copy — select the text above instead'),
    ));
  }
},
```

Please add a test where the mock handler throws on `Clipboard.setData`.

lib/core/startup_failure.dart
```diff
@@ -0,0 +1,134 @@
...
+                textAlign: TextAlign.center,
+                style: TextStyle(
+                  color: Colors.white,
+                  fontSize: 20,
+                  fontWeight: FontWeight.w600,
+                ),
+              ),
+              const SizedBox(height: 12),
+              Text(
+                'It failed while $step.',
```
grunch: _🎨 UX_ | _🟡 Minor_

**The screen says what failed, but not what to do.**

The text is actionable for a maintainer, but the person who sees it only gets "Copy details" and nowhere to paste them. The two **required** steps that can fail from local state fail on every launch:
- `reading your settings`: a corrupt or incompatible preference;
- `loading the engine`: a mismatched wasm bundle.

There is no retry (fine, for the `RustLib.init` reason in the description), so the only way out is clearing site or app data, and nothing says so.

One hard-coded line keeps the no-dependency rule, for example: *"Please report this at github.com/MostroP2P/app/issues. If it keeps happening, clearing the app's data may help — you will need your secret words to restore your account."* Adjust the wording to what is safe to recommend. A dead end in an app that holds money should point to the way out.

lib/core/startup_sequence.dart
```diff
@@ -0,0 +1,38 @@
...
+  /// continues, degraded.
+  ///
+  /// Every degradation prints the same `[startup]` prefix, so grepping it lists
+  /// what a run gave up on, in order — which matters when one failure causes
+  /// the next.
+  Future<void> optional(String name, Future<void> Function() body) async {
+    currentStep = name;
+    try {
+      await body();
+    } catch (e, st) {
```
grunch, line 61: _🎯 Diagnostics_ | _🔵 Nit_

**`catch (e, st)` also swallows `Error`.**

That includes `TypeError`, `LateInitializationError`, `RangeError` and, in debug, `AssertionError`. These are programming bugs, not an environment problem, but today each becomes one `[startup]` line and the app opens half-initialised. Keeping release behaviour unchanged:

```dart
} catch (e, st) {
  debugPrint('[startup] $name failed — continuing without it: $e\n$st');
  assert(e is! Error, 'optional step "$name" hit a programming error: $e');
}
```

(or rethrow `Error` when `kDebugMode`). A bug in an optional step then fails loudly during development instead of degrading silently.

lib/core/startup_sequence.dart
```diff
@@ -0,0 +1,38 @@
...
+/// that catches it is at the end of the sequence and has no other way to know.
+///
+/// Both helpers set [currentStep] on entry, so every stretch of startup runs
+/// under its own name. That is the point of having two of them rather than one:
+/// with only [optional], the mandatory stretches in between would keep running
+/// under the label of whichever optional step finished last, and a failure
+/// there would name a step that succeeded (#405 review).
+class StartupSequence {
+  /// The step in progress. Read by the failure surface.
+  String currentStep = 'starting up';
```
grunch, line 49: _♻️ Maintainability_ | _🔵 Nit_

**The label can still go stale through the public setter.**

The round-1 bug was a stretch running under the previous step's name. `required` and `optional` fix it structurally, but the three steps with their own handlers (`selecting the Mostro node`, `mirroring trade key indices`, `loading your identity`) set `currentStep` by hand. The next step added with its own `catch` that forgets that line brings the bug back, and no test catches it.

A third helper sets the label and leaves error handling to the caller:

```dart
Future<void> handled(
  String name,
  Future<void> Function() body, {
  required void Function(Object e, StackTrace st) onError,
}) async {
  currentStep = name;
  try { await body(); } catch (e, st) { onError(e, st); }
}
```

Then make the field read-only (`String get currentStep`). The bridge-probe step becomes `startup.handled('selecting the Mostro node', …, onError: (e, _) { debugPrint(…); markBridgeFailed(e); })`.

lib/core/startup_failure.dart
```diff
@@ -0,0 +1,134 @@
...
+  String _report() {
+    final cause = error == null ? '' : '\n${_shortError(error!)}';
+    return 'Mostro could not start: $step$cause';
+  }
+
+  static String _shortError(Object error) {
+    final text = error.toString();
+    return text.length <= _maxErrorChars
+        ? text
+        : '${text.substring(0, _maxErrorChars)}…';
```
grunch: _🎯 Text handling / a11y_ | _🔵 Nit_

**Truncation and contrast.**

- `text.substring(0, 300)` counts UTF-16 code units, so cutting at an emoji or other astral character leaves a lone surrogate. It renders as a replacement glyph, and on the web it can throw when passed to JS. Use `text.characters.take(_maxErrorChars).toString()` instead; `package:characters` already comes with Flutter.
- The cause is 12 px `#7A8296` on `#1D212C`, which I compute at about **4.2:1**. That is below the WCAG AA minimum of 4.5:1 for small text, on the line people will be asked to read out. `#8C94A8` or lighter clears 4.5:1 without a visible change in tone.

grunch (changes requested):
## Review (strict pass, round 3)

The round-2 findings are resolved:
- every stretch now runs under its own label (`required` + `optional`, locked in by `startup_sequence_test.dart`);
- the deep-route test really sets `defaultRouteNameTestValue` and checks `canPop()`;
- the `!kIsWeb` grep is scoped to the database step;
- #227 is cited as a precedent;
- `markBridgeFailed` is called from the guard.

The step classification is careful. What remains is mostly *around* the guard: how CI observes it, how the failure screen behaves when copying fails, and whether the guard's own logic is really tested.

### Findings

| # | Severity | Where | Summary |
|---|---|---|---|
| 1 | 🟠 Major | `app_bootstrap.dart` guard | A fatal failure *after* `markBridgeReady()` now passes the web smoke test. It used to be an uncaught page error; now it is a `debugPrint` plus a flag the smoke test no longer reads |
| 2 | 🟠 Major | `startup_guard_test.dart` | The guard itself (catch → `runApp(StartupFailureApp)` → `markBridgeFailed`) is only checked by a substring grep, so a dead or misplaced call still passes. It can be extracted and tested without Rust |
| 3 | 🟡 Minor | `app_bootstrap.dart` `building the interface` | The comment says `runApp` is "already covered", but only a synchronous throw is: a `MostroApp` build error happens in the first frame, outside the `try` |
| 4 | 🟡 Minor | `app_bootstrap.dart` `building the interface` | A throw after `AppLifecycleService.attach()` or after the bond consumers start leaves them running on a container that is never mounted, and `ResumeResync` runs behind the failure screen |
| 5 | 🟡 Minor | `startup_failure.dart` Copy details | `Clipboard.setData` is not awaited. "Copied" shows even when the write is rejected (web without clipboard permission, non-secure context), and the rejection goes unhandled |
| 6 | 🟡 Minor | `startup_failure.dart` | No next step on the screen: nowhere to report, no hint about clearing app data. With `reading your settings` required, a corrupt preference is a permanent dead end |
| 7 | 🔵 Nit | `startup_sequence.dart` | `optional` also swallows `Error`s (`TypeError`, `LateInitializationError`, failed asserts), so programming bugs degrade silently during development |
| 8 | 🔵 Nit | `startup_sequence.dart` / `app_bootstrap.dart` | `currentStep` is public and mutable, and three steps set it by hand. The first new step with its own handler that forgets the line brings back the round-1 bug |
| 9 | 🔵 Nit | `startup_failure.dart` | `substring(0, 300)` can split a surrogate pair, and the cause text (`#7A8296` on `#1D212C`, 12 px) is ≈4.2:1, below WCAG AA 4.5:1 |

Checks on `54bf16f`:
- `flutter analyze` reports no issues on `startup_failure.dart`, `startup_sequence.dart` and the three new test files.
- The 17 new tests pass.
- A full-tree analyze in a fresh worktree fails only because the FRB bindings aren't generated there, which is unrelated to this PR.
- I did not run the smoke test. Finding 1 comes from reading `smoke.mjs`: stage 3 and its `console` / `pageerror` collectors.

**Verdict: request changes.** 1 and 2 should be fixed before merge. The rest can go in this PR or a follow-up issue.

Matobi98:
Thanks for the round-3 review. Six findings are fixed in this PR (1, 2, 5, 6, 7, 9). For 3, the comment is corrected here and first-frame coverage moves to a follow-up issue. 4 and 8 go to follow-up issues, as you suggested. Most of the fixes are measured rather than read off the code, so each point below says how it was checked.

**Verified by hand:** a full sell-order trade against local regtest, the failure screen on Linux (points 5 and 6), and the web smoke test before and after the fix (point 1). **Agent-run:** the unit and widget tests, mutation checks (each fix reverted to confirm its test goes red), the isolated debug run for point 7, the character-count and contrast checks for point 9, and a final smoke-test re-run after the last layout change.

### 1. Late startup failures now fail the smoke test
You were right about the gap, and it turned out to be older than this PR. I built the pre-PR commit (`86b74a7`) with a `throw` injected just before the interface is built, and the smoke test **passed** 3 out of 3 times. A timing probe showed why: the test finished its final checks at ~1.2 s, and the failure only arrived at ~1.6 s. `mostroBridgeReady` was set at the first Rust call, so nothing after that point was ever watched.

The guard made it worse, as you pointed out: once caught, the failure became a `debugPrint` (`console.log`), so even a test that waited would not have seen it.

Fix: `markBridgeReady()` now runs at the very end of startup, after `runApp`. A failure anywhere before it never sets the ready flag; the guard sets `mostroBridgeError` instead, and the smoke test already waits for either flag. `smoke.mjs` now reports `startup finished` / `startup failed: …`, and the comments in `bridge_probe*.dart`, `CLAUDE.md` and `web-build.yml` now describe the flag as "startup finished".

Result: with the same injected `throw`, the smoke test fails with `startup failed: Bad state: …`. Without it, the test passes, including `SMOKE_BOND_STORE=1`.

### 2. The guard is executed, not grepped
The try/catch moved into `runGuarded(body, run:, onFailed:)` in `startup_sequence.dart`, and `bootstrapAndRun` calls it. The tests drive it with a body that throws inside `required('building the interface', …)` and check that `StartupFailureApp` is painted with that step and error, and that `onFailed` runs after `run`. All four cases you listed turn the test red: the call moved out of the catch, an earlier catch swallowing the error, the order flipped, and the body left unawaited.

The same treatment went to the database step: `openDatabase(isWeb:, dataDir:, initDb:)` is now executed in `db_location_test.dart`, and it opens `mostro` on the web without asking for a directory. One grep stays, limited to the call site in `_startup`. Executing that line would mean faking ~20 Rust calls to protect one line, and the test comment says so.

### 3. `runApp` scope
The comment now states that the guard only catches what `runApp` throws synchronously; a first-frame failure bypasses both the guard and the ready flag. Covering the first frame goes to a follow-up issue.

### 4. Listeners outliving a failed startup
Follow-up issue. The change is small, but two of those starts depend on their current order (the bond consumers subscribe before relay delivery, and the lifecycle service is attached before `runApp`), so it needs its own verification.

### 5. Clipboard
The write is awaited; on failure the screen shows "Could not copy. Select the text instead." A test uses a clipboard handler that throws.

### 6. A way out
The screen now has a second copy button for `https://github.com/MostroP2P/app/issues`. A footer explains what is at stake rather than recommending that users clear their data: without the secret words, uninstalling or clearing app data loses access to open trades, disputes and reputation. I checked this against the code: trade keys derive from the mnemonic (`m/44'/1237'/38383'/0/N`), `restore_session` works from the identity key, and ratings are stored on the master pubkey. The layout was also reworked: each copy button sits right of what it copies, the buttons line up in one column, and the title is larger.

### 7. `Error` in optional steps
In debug, an `Error` is reported through `FlutterError.reportError` and startup continues. Release is unchanged. I went with a report rather than an `assert`/rethrow so that a step already failing this way on some platform doesn't block every debug run there. An isolated Linux debug run reached the end of startup with no `Error` reported.

### 8. Stale step label
Follow-up issue (a `handled` helper and a read-only `currentStep`). Nothing is broken today; it prevents a regression.

### 9. Truncation and contrast
Truncation now counts characters (`characters.take`), and a test puts an emoji at the cut. The cause colour is `#8C94A8`, which comes out at 5.3:1 on `#1D212C`.

I'll link the follow-up issues for 3, 4 and 8 here once they are open.

Matobi98:
Thanks for the round-3 review. Fixed here: 1, 2, 5, 6, 7, 9. Follow-up issues: #489 (3), #510 (4), #490 (8). Each point quotes your finding, then the change, then how I checked it: by hand in the app, or with a script I ran that breaks one line at a time and runs the tests (for what can't be done by hand, like a race).

### 1. Smoke test after `markBridgeReady()`
> A fatal failure after `markBridgeReady()` passes the web smoke test.

**Change:** `markBridgeReady()` now runs at the very end of startup, after `runApp`. A failure anywhere before it leaves only `mostroBridgeError`, which the smoke test already waits for.

**Checked by hand**, with a `throw` injected before the interface is built:
- pre-PR (`86b74a7`): `BUNDLE_DIR=… node smoke.mjs` → `web bundle smoke test passed.` — the gap was older than this PR;
- this branch: same command → `✗ web bundle smoke test failed: startup failed: Bad state: …`.

### 2. The guard is only grepped
> The guard itself is only checked by a substring grep.

**Change:** the try/catch moved into `runGuarded(body, run:, onFailed:)`; tests execute it with a body that throws. `openDatabase` is executed in a test too; one grep stays on its call site, since running it means faking ~20 Rust calls.

**Checked with the script**, against `startup_guard_test.dart`: each of your four cases — `run(...)` removed from the catch, `try { await body(startup); } catch (_) {}`, `run`/`onFailed` swapped, `body(startup)` without `await` — gives `[E] a failure reaches runApp, naming the step that failed`.

### 3. `runApp` "already covered"
> Only a synchronous throw is covered; a first-frame error happens outside the `try`.

**Change:** the comment now says so. Covering the first frame: #489.

### 4. Listeners outlive a failed startup
> A throw after `AppLifecycleService.attach()` leaves `ResumeResync` running behind the failure screen.

**Follow-up:** #510 — some of those starts depend on their order, so moving them needs its own verification.

**Checked by hand**, with a `throw` right after `.attach()`: `flutter run -d chrome --web-port=5555 …`, then switch tab and back → `[lifecycle] resync: online=true flushed=0 coalesced=false` behind the failure screen.

### 5. Clipboard not awaited
> "Copied" shows even when the write is rejected.

**Change:** the write is awaited; on failure the screen shows "Could not copy. Select the text instead."

**Checked by hand:** tapping copy → "Copied", and the pasted text holds the step and the cause. With the script, dropping the `await` gives `[E] says so when the copy fails instead of claiming it worked`.

### 6. No next step
> Nowhere to report, no hint about clearing app data.

**Change:** a second copy button for `https://github.com/MostroP2P/app/issues`, and a footer: without the secret words, uninstalling or clearing app data loses access to open trades, disputes and reputation (trade keys derive from the mnemonic, `m/44'/1237'/38383'/0/N`; `restore_session.rs:15`: `let master_key = event.identity.to_string();`; `rate_user.rs:115`: the rating goes to `normal_seller_idkey` / `normal_buyer_idkey`, the identity keys).

**Checked by hand:** "Copy link" pastes the URL; the footer sits at the bottom.

### 7. `optional` swallows `Error`s
> Programming bugs degrade silently.

**Change:** in debug an `Error` is reported through `FlutterError.reportError` and startup continues, so a step already failing on some platform doesn't block debug runs. Release unchanged.

**Checked with the script:** replacing the condition with `if (false)` gives `[E] optional reports a programming error loudly in debug`.

### 8. Stale step label
> `currentStep` is public and mutable; three steps set it by hand.

**Follow-up:** #490. Nothing is broken today; it prevents a regression.

### 9. Truncation and contrast
> `substring` can split a surrogate pair; the cause text is ≈4.2:1.

**Change:** truncation counts characters (`characters.take`); the colour is `#8C94A8`, 5.3:1 on `#1D212C`.

**Checked with the script:** going back to `substring` gives `[E] truncates by character, never inside an emoji`; the WCAG formula on `#8C94A8` / `#1D212C` prints `5.3`.

Catrya:
@Matobi98 please fix the conflicts

Matobi98:
Conflicts fixed, pushed as `9ed10fc6`.

One of them needed a decision, so here is what I did.

While this PR was open, #494 changed the same block: Firebase, `RustLib.init()` and `SharedPreferences` stopped running one after another and now start together. This PR names each startup step so a failure can say where it stopped, and `StartupSequence` holds that name in a single field. With the three running under it at once, a failure could report the name of a step that actually worked.

They still start together, but each one is awaited inside its own step:

```dart
final firebase = _initFirebase()..ignore();
final engine = RustLib.init()..ignore();
final preferences = SharedPreferences.getInstance()..ignore();

await startup.optional('setting up notifications', () => firebase);
await startup.required('loading the engine', () => engine);
```

`ignore()` is there because a future that fails before anything awaits it is reported as an unhandled error. It marks the error as handled, and the `await` below still throws.

One behaviour change worth flagging: Firebase is an optional step again. Reading `main`, a Firebase failure that is not `UnsupportedError` goes into the `.wait` and takes startup down with it — I did not run that case, it is how I read the code. Here it is an optional step, so the app opens without push notifications.

There is a test for this in `startup_sequence_test.dart`: work that starts before its step still fails under that step's name. It fails if `required` stops setting the name.

What I checked by hand:

- Took `librust.so` out of the built bundle and ran it. The failure screen named `loading the engine`:

  

- Same run on Linux, where Firebase has no configuration: the log says `[startup] Firebase not configured: ... push notifications disabled` and the app opens normally.
- Full sell flow against a local Polar regtest and mostrod: created the order from the CLI, took it in the app, paid the hold invoice, fiat sent, released. It completed.

`flutter test`: 1597 passing.

Matobi98:
Taken — _report() now sends the whole cause and _shortError stays on the display path only.
▎
▎ Checked by hand: forced a 412-character failure, and the screen shows it cut with an ellipsis while Copy details pastes it whole, ending in the last word. There is also a test for it: it asserts the copied text keeps the tail and carries no ellipsis, and it fails on the previous _report().

PR #406, opened 2026-09-08
fix(#404): cover every locale tag the sanitizer handles, and what it leaves behind

### LINKED ISSUE #404: Web smoke test: shadow navigator.languages so CI can cover en_US, "" and mixed locale lists
`web-build.yml` runs the release-bundle smoke test under an invalid locale as a regression guard for #227, plus a negative control that strips the sanitizer and requires the run to die with `RangeError: Incorrect locale information provided`. That guard only ever exercises `C`.

The reason is `smoke.mjs`'s use of Playwright's `locale` option, which normalizes the tag before the page sees it. Measured:

| `SMOKE_LOCALE` | what the page actually reports |
|---|---|
| `C` | `C` — reaches the page intact |
| `en_US` | `en-US` — normalized by Playwright |
| `""` | the system locale — the empty string never arrives |

So two of the three tags named in #227's acceptance criterion "the app starts and renders with the browser locale forced to `C`, `en_US` and `""`" are handled by the sanitizer but not covered by CI, and the mixed-list path (`["C","es-AR"]` → `["es-AR"]`) is covered by nothing at all.

This is a limitation of that one Playwright option, not of the browser. `context.addInitScript()` runs before any page script — including the sanitizer — so the smoke test can shadow the properties directly and reach the real code path:

```js
// smoke.mjs — opt-in knob, comma-separated, used verbatim (no normalisation)
const forced = process.env.SMOKE_NAVIGATOR_LANGUAGES;
if (forced !== undefined) {
  await page.addInitScript((langs) => {
    Object.defineProperty(navigator, 'languages', { get: () => langs, configurable: true });
    Object.defineProperty(navigator, 'language', { get: () => langs[0] ?? '', configurable: true });
  }, forced.split(','));
}
```

`configurable: true` matters: the sanitizer bails out when either property is already locked down, so a non-configurable shadow would make it skip the very path under test.

### Acceptance criteria

- [ ] `smoke.mjs` accepts `SMOKE_NAVIGATOR_LANGUAGES` and injects it via `addInitScript`, before the sanitizer runs
- [ ] `web-build.yml` exercises `C`, `en_US`, `""` and `C,es-AR` against the release bundle
- [ ] The negative control covers the same matrix, still requiring the `RangeError` signature
- [ ] The "CI only reaches C" caveat on #227's acceptance criteria is removed once it no longer holds

### Context

Follow-up from @grunch's review of #370 (finding 3), left out of that PR at his suggestion because it changes what `smoke.mjs` accepts as input and wants a matrix in the workflow — neither of which the sanitizer itself needed.

The sanitizer already handles all of these tags correctly; verified by hand in Chromium against `C`, `en_US`, `""`, `POSIX`, `en-US` and `C,es-AR`. What is missing is CI proving it, so a future regression in the sanitizer would only be caught for `C`.

Matobi98:
# fix(#404): cover every locale tag the sanitizer handles, and what it leaves behind

Closes #404. Follow-up to #370, from @grunch's third finding there.

No app code changes: `lib/`, `web/` and `rust/` are untouched, and the bundle that ships is byte-identical to main's. What changes is what CI checks before letting a change through.

## The gap

The locale guard added in #370 only ever exercised `C`. Playwright's `locale` option normalizes the tag before the page sees it — `en_US` arrives as a valid `en-US`, `""` falls back to the system locale — so two of the three tags named in #227's acceptance criterion were handled by the sanitizer but never reached it in CI. The mixed list (`["C","es-AR"]` → `["es-AR"]`) was tested by nothing at all.

That is a limit of that one Playwright option, not of the browser. `addInitScript` runs inside the page before any of its own scripts, so it can hand the engine a tag Playwright would never deliver.

## What this adds

**`SMOKE_NAVIGATOR_LANGUAGES`** shadows `navigator.language(s)` from an init script. Comma-separated, used verbatim, unset means "do not touch" — so every existing run is unaffected. `!== undefined` rather than a truthiness check, because the empty string is one of the broken tags. `configurable: true` is load-bearing, though not as a false-green guard: the sanitizer bails out when either property is locked down, so a non-configurable shadow makes it skip the very path under test and the positive run *fails*, with the same message a real regression produces. What it prevents is a red matrix that reads as a broken sanitizer when it is really a broken harness. Measured both ways on `C` and `C,es-AR` (review round).

**`SMOKE_EXPECT_LANGUAGES`** asserts what `navigator.languages` reads after the sanitizer ran. This is what makes the mixed list worth running — see below.

**The matrix** in `web-build.yml` runs `C`, `en_US`, `""` and `C,es-AR`, each against the shipped bundle (must pass, with the expected locale) and against the control bundle (must fail with a locale error).

**`selftest.mjs` holds that second check down** (review round). Two cases against the existing `healthy` fixture, which has no sanitizer, so whatever is injected is what `navigator.languages` still reads at the end: `C,es-AR` expecting `es-AR` must fail, `es-AR` expecting `es-AR` must pass. A comparison that never fires and one that always fires each pass one of those and fail the other, so neither case is redundant. The passing one also proves the init script reaches the page at all. `SMOKE_NAVIGATOR_LANGUAGES` is what makes this testable without a bundle — before it, the comparison was a gate nothing in CI held down.

**The control bundle is now built once** in its own step instead of being rebuilt inside the control, since two places would otherwise duplicate the strip and drift. The "did the strip actually happen" check moved with it: if the marker is ever renamed, the cut is a no-op and every control silently compares the bundle against itself.

**The `SMOKE_LOCALE=C` step and its control stay.** It is the one path where the tag reaches the page as the browser's own locale rather than as a shadow this repo installed.

## Two things this turned up

**1. The empty tag crashes with a different message.** Three of the four controls die with `Incorrect locale information provided`; the empty one dies with `First argument to Intl.Locale constructor can't be empty or missing`. Both are the engine refusing a locale before `runApp`, but the control introduced in #370 required the first string exactly — so the empty case would have reported *"the control failed, but not with the locale RangeError"*, a false negative on a control that worked perfectly.

That check is not wrong today: it only ever runs `C`, and for `C` the signature is right. It becomes too narrow the moment this PR starts feeding it the other tags, so widening it belongs here rather than in a follow-up. The accepted signatures are now a job-level `LOCALE_CRASH_SIGNATURE`, kept exact rather than loosened to something like `locale` — a lax pattern is what lets a control celebrate an unrelated crash.

**2. "The view mounted" cannot judge the mixed list.** With `C,es-AR`, a sanitizer that keeps `es-AR` and one that drops the whole list for the fallback both boot perfectly; only the second silently costs the user their language. Without reading the result back, that case tests nothing `C` does not already test.

Hence `SMOKE_EXPECT_LANGUAGES`. Mutation-checked: patching the sanitizer to discard the whole list whenever any tag is invalid leaves `C`, `en_US` and `""` green and fails only the mixed case, with `navigator.languages is "en-US", expected "es-AR"`.

## Verified

Against a release bundle built on this branch (`build-web.sh --release` + `flutter build web --release`, Flutter 3.38.2), running the matrix exactly as the workflow does:

| tag | expected after | with sanitizer | without |
|---|---|---|---|
| `C` | `en-US` | passes | dies with the locale signature |
| `en_US` | `en-US` | passes | dies with the locale signature |
| `""` | `en-US` | passes | dies with the locale signature |
| `C,es-AR` | **`es-AR`** | passes | dies with the locale signature |

Eight for eight. Plus `selftest.mjs` 5/5: the two new cases hold the `SMOKE_EXPECT_LANGUAGES` comparison down from both sides, and each is mutation-checked — `if (false)` fails the mismatch case only, `if (true)` fails the match case only, so neither is redundant and neither passes on a broken check. `pages_bundle_test.dart` also still passes 17/17, but it asserts the sanitizer's position in the bundle and the shape of these workflows; it does not reach the matrix or either new variable.

Verified by hand too, in a real browser window against that same bundle. With `C` forced, the control is a blank page with `RangeError: Incorrect locale information provided` while the shipped bundle loads. With `['C','es-AR']` injected, three bundles side by side: no sanitizer is blank, the shipped one opens in Spanish, and one with the sanitizer mutated to discard the whole list opens in English — no error, no blank page, nothing logged, the user simply loses their language. That middle window is what `SMOKE_EXPECT_LANGUAGES` exists to catch, and what every other assertion here reports as green.

## Cost

Four extra control runs at ~20s each — each waits out `SMOKE_TIMEOUT_MS`, set inline on the control invocation rather than on the step, for a view that will never mount — so roughly 80s added per PR. The positive runs are ~1s each. That is the price of the guard being self-validating; the alternative is a matrix that can go green for the wrong reason.

## For #227

Its acceptance criteria carry no caveat to remove: the *"the CI step only exercises `C`"* note lives in #370's body, which is a record of what was true when it merged and is better left alone. What #227 does have is two criteria this PR satisfies and that are still unticked — the app rendering under `C`, `en_US` and `""`, and a CI case asserting the view mounts under an invalid locale. Ticking those is the honest version of #404's fourth criterion; happy to do it, or leave it to whoever merges.

### DIFF
--- a/.github/workflows/web-build.yml
+++ b/.github/workflows/web-build.yml
@@ -43,6 +43,16 @@ jobs:
       contents: read
     # The first uncached run compiles std from source (-Z build-std), which is slow.
     timeout-minutes: 60
+    env:
+      # What a negative control must find in its output to count as "died of
+      # the bug the sanitizer fixes", rather than of any of the five unrelated
+      # reasons smoke.mjs exits non-zero for. Two messages because Intl.Locale
+      # words the empty tag differently from a malformed one — both are the
+      # engine refusing a locale before runApp, and matching only the first
+      # would report the empty case as an unexplained failure. Kept exact
+      # rather than loosened to something like "locale": a lax pattern is what
+      # lets a control celebrate an unrelated crash.
+      LOCALE_CRASH_SIGNATURE: "Incorrect locale information provided|Intl.Locale constructor can't be empty"
     steps:
       - uses: actions/checkout@v4
         with:
@@ -198,10 +208,11 @@ jobs:
       # navigator.language(s). This reruns the same smoke test forcing an
       # invalid locale so the fix can't silently regress.
       #
-      # Only "C" is actually covered here: Playwright normalizes the locale it
-      # is given before the page sees it, so "en_US" reaches the browser as a
-      # valid "en-US" and "" falls back to the default. The sanitizer handles
-      # all three; CI can only exercise the first.
+      # This step covers "C" only, and does so through Playwright's `locale`
+      # option — the one path where the tag reaches the page as the browser's
+      # own locale rather than as a shadow this repo installed. Kept for that
+      # reason. The other broken tags cannot travel this way (Playwright
+      # normalizes them first) and are covered by the locale matrix below.
       #
       # The crash itself only exists on Flutter >= 3.38, where the engine's
       # parseBrowserLanguages started building the locale through
@@ -209,6 +220,28 @@ jobs:
       # validated nothing. If FLUTTER_VERSION ever moves back below that, or
       # the engine stops using Intl, this step passes for the wrong reason —
       # revalidate it against a build without the sanitizer before trusting it.
+      # Built once and reused by every negative control below. Copying the
+      # built bundle and cutting the sanitizer out of its HTML — rather than
+      # building a second time — is what makes the comparison mean something:
+      # both sides come from one build, so the sanitizer is the only difference.
+      - name: Build the control bundle — the same build, sanitizer removed (#227)
+        run: |
+          set -euo pipefail
+          control_dir="${RUNNER_TEMP}/web-no-sanitizer"
+          rm -rf "$control_dir"
+          cp -r "${GITHUB_WORKSPACE}/build/web" "$control_dir"
+          # Delete the marker comment through the end of the inline script it
+          # introduces. The script contains no literal </script>, so the first
+          # one after the marker closes it.
+          sed -i '/<!-- locale-sanitizer/,/<\/script>/d' "$control_dir/index.html"
+          # Never let a silent no-op turn the controls into a rubber stamp: if
+          # the marker is ever renamed, this cuts nothing and every control
+          # below would be comparing the bundle against itself.
+          if grep -q 'locale-sanitizer' "$control_dir/index.html"; then
+            echo "::error::failed to strip the sanitizer — the controls prove nothing"
+            exit 1
+          fi
+
       - name: Smoke test the release bundle under an invalid locale (#227)
         working-directory: test/web/smoke
         env:
@@ -240,17 +273,6 @@ jobs:
         run: |
           set -euo pipefail
           control_dir="${RUNNER_TEMP}/web-no-sanitizer"
-          rm -rf "$control_dir"
-          cp -r "${GITHUB_WORKSPACE}/build/web" "$control_dir"
-          # Delete the marker comment through the end of the inline script it
-          # introduces. The script contains no literal </script>, so the first
-          # one after the marker closes it.
-          sed -i '/<!-- locale-sanitizer/,/<\/script>/d' "$control_dir/index.html"
-          # Never let a silent no-op turn this control into a rubber stamp.
-          if grep -q 'locale-sanitizer' "$control_dir/index.html"; then
-            echo "::error::failed to strip the sanitizer — the control proves nothing"
-            exit 1
-          fi
           control_log="${RUNNER_TEMP}/control-smoke.log"
           if BUNDLE_DIR="$control_dir" BASE_PATH="/${GITHUB_REPOSITORY#*/}/" node smoke.mjs >"$control_log" 2>&1; then
             cat "$control_log"
@@ -260,7 +282,7 @@ jobs:
           fi
           # It failed — but smoke.mjs fails for five reasons that have nothing
           # to do with this bug. Only the locale RangeError proves the control.
-          if ! grep -q 'Incorrect locale information provided' "$control_log"; then
+          if ! grep -Eq "$LOCALE_CRASH_SIGNATURE" "$control_log"; then
             cat "$control_log"
             echo "::error::the control failed, but not with the locale RangeError —"
             echo "::error::it proves nothing about this fix (see #227)"
@@ -271,6 +293,74 @@ jobs:
           # failure does not upload this expected one as its evidence.
           rm -f smoke-failure.png
           echo "control failed as expected — the locale guard is load-bearing"
+
+      # The two steps above only ever exercise "C". Playwright's `locale`
+      # option normalizes the tag before the page sees it, so "en_US" arrives
+      # as a valid "en-US" and "" falls back to the system locale — a limit of
+      # that option, not of the browser. SMOKE_NAVIGATOR_LANGUAGES shadows
+      # navigator.language(s) from an init script instead, which runs inside
+      # the page before any of its own scripts, so the engine is handed a tag
+      # Playwright would never deliver.
+      #
+      # That closes the gap on #227's second acceptance criterion, which names
+      # three tags, and covers the mixed list nothing tested at all: with
+      # "C,es-AR" the correct result is not the fallback but "es-AR" — the
+      # sanitizer must drop the broken tag and keep the good one.
+      #
+      # Which is why each case also states the locale it expects to find
+      # afterwards. "The view mounted" cannot tell a sanitizer that kept
+      # "es-AR" from one that dropped the whole list for the fallback: both
+      # boot fine, and only the second silently costs the user their language.
+      #
+      # Each tag runs twice: against the shipped bundle it must pass, against
+      # the control bundle it must fail with the locale signature. The control
+      # half costs ~20s per tag (it waits out SMOKE_TIMEOUT_MS for a view that
+      # will never mount), which is the price of the guard being self-validating.
+      - name: Locale matrix — every tag the sanitizer claims to handle (#227)
+        working-directory: test/web/smoke
+        env:
+          SMOKE_TIMEOUT_MS: "20000"
+        run: |
+          set -euo pipefail
+          control_dir="${RUNNER_TEMP}/web-no-sanitizer"
+          base="/${GITHUB_REPOSITORY#*/}/"
+          # Deliberately includes the empty string; it is one of the broken
+          # tags, and smoke.mjs distinguishes unset from empty for that reason.
+          # "<injected tags>|<languages expected after the sanitizer>".
+          for case in "C|en-US" "en_US|en-US" "|en-US" "C,es-AR|es-AR"; do
+            langs="${case%%|*}"
+            expected="${case##*|}"
+            label="${langs:-<empty>}"
+
+            echo "::group::locale [$label] — with the sanitizer, must pass and leave [$expected]"
+            SMOKE_NAVIGATOR_LANGUAGES="$langs" SMOKE_EXPECT_LANGUAGES="$expected" \
+              BUNDLE_DIR="${GITHUB_WORKSPACE}/build/web" \
+              BASE_PATH="$base" node smoke.mjs
+            echo "::endgroup::"
+
+            echo "::group::locale [$label] — without the sanitizer, must fail"
+            log="${RUNNER_TEMP}/control-${label//[^A-Za-z0-9]/_}.log"
+            if SMOKE_NAVIGATOR_LANGUAGES="$langs" BUNDLE_DIR="$control_dir" \
+                 BASE_PATH="$base" node smoke.mjs >"$log" 2>&1; then
+              cat "$log"
+              echo "::error::locale [$label] passed without the sanitizer —"
+              echo "::error::the matrix above is no longer testing this bug (see #227)"
+              exit 1
+            fi
+            if ! grep -Eq "$LOCALE_CRASH_SIGNATURE" "$log"; then
+              cat "$log"
+              echo "::error::locale [$label] failed, but not with a locale error —"
+              echo "::error::it proves nothing about this fix (see #227)"
+              exit 1
+            fi
+            cat "$log"
+            echo "::endgroup::"
+
+            # Each failing control leaves a screenshot behind; drop it so a
+            # later, real failure does not upload an expected one as evidence.
+            rm -f smoke-failure.png
+            echo "locale [$label]: sanitized to [$expected] with the sanitizer, dies without it"
+          done
 
       - name: Upload smoke test failure screenshot
         if: failure()
--- a/.github/workflows/web-build.yml
+++ b/.github/workflows/web-build.yml
@@ -199,27 +199,6 @@ jobs:
         env:
           BUNDLE_DIR: ${{ github.workspace }}/build/web
         run: BASE_PATH="/${GITHUB_REPOSITORY#*/}/" node smoke.mjs
-
-      # Regression guard for issue #227: an unparseable browser locale (the
-      # POSIX "C" locale, an underscored "en_US", empty) crashed the Flutter
-      # web engine's own bootstrap before runApp, with no error path — an
-      # unrecoverable blank page. Fixed by the locale sanitizer in
-      # web/index.html, which runs before flutter_bootstrap.js and normalizes
-      # navigator.language(s). This reruns the same smoke test forcing an
-      # invalid locale so the fix can't silently regress.
-      #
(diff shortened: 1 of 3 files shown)


Catrya (changes requested):
**Changes requested** — light. The matrix does what it says and I reproduced it end to end, but the new `SMOKE_EXPECT_LANGUAGES` check has no negative-side coverage in CI, and this PR is what makes that coverage possible.

## Blocking

**`SMOKE_EXPECT_LANGUAGES` is a gate nothing in CI holds down.** Two mutations against a release bundle built from this branch:

- Sanitizer patched to discard the whole list whenever any tag is invalid: `C`, `en_US` and `""` stay green, and only `C,es-AR` fails, with `navigator.languages is "en-US", expected "es-AR"`. Your mutation result, verbatim.
- Same broken sanitizer, with the new check neutered (`if (actual !== expected)` → `if (false)`): the mixed case **passes**, even though the user has lost `es-AR`. `selftest.mjs` still exits 0.

So if that comparison ever rots, the one case the check exists for goes green for the wrong reason, which is exactly what the body argues a matrix must not do. `selftest.mjs` exists for this ("a gate nobody tests is a gate that quietly stops gating"), and `SMOKE_NAVIGATOR_LANGUAGES` is what lets it test the check without a bundle. Both cases work against the existing `healthy` fixture:

```
SMOKE_NAVIGATOR_LANGUAGES=C,es-AR SMOKE_EXPECT_LANGUAGES=es-AR  -> exit 1: navigator.languages is "C,es-AR", expected "es-AR"
SMOKE_NAVIGATOR_LANGUAGES=es-AR   SMOKE_EXPECT_LANGUAGES=es-AR  -> exit 0
```

Two more entries in `CASES` (plus an `env` field), a few seconds of runtime. The passing one also proves the init script reaches the page at all.

## Non-blocking

- **Doc-comment hijack.** The new "Build the control bundle" step landed between the #227 comment block (`web-build.yml:203-222`, no blank line in between) and the step that block describes (`:245`). Those 20 lines now read as documentation of the copy-and-strip step, and the `SMOKE_LOCALE=C` step is left bare. Worth moving in the same round.
- **Right code, wrong reason, on `configurable: true`** (`smoke.mjs:187-190`, same wording in the body). A non-configurable shadow does not make the run "pass for the wrong reason". The sanitizer bails out as described, the engine gets `C`, and the positive run **fails** with `Incorrect locale information provided`. I measured that with `configurable: false` for both `C` and `C,es-AR`. The flag is still required, but what it prevents is a red matrix that looks like a sanitizer regression when it is really a harness bug. Worth stating it that way so the wrong reason doesn't become folklore.
- **The 20 s timeout also applies to the positive matrix runs.** The step-level `SMOKE_TIMEOUT_MS: "20000"` (`:321-322`) contradicts the comment at `:269-271` ("Do not 'unify' this with the positive steps"). Low risk in practice, since the positive run takes 1–2 s in CI, but scoping it inline to the control invocation keeps the two consistent.
- **The body claims more coverage than exists.** "`pages_bundle_test.dart` 17/17 and `selftest.mjs` 3/3, both of which cover this workflow and this script": neither one touches the matrix or either new variable, and both pass with or without them.
- **`Closes #404` will close it with its fourth criterion open** (removing the "CI only reaches C" caveat from #227). That's an issue edit, not a code change. Whoever merges can take care of it.

Matobi98:
Thanks — all six addressed. The two new `selftest.mjs` cases are in the diff; what follows is what the diff cannot show, plus one correction.

**Blocker — done.** `CASES` grows an `env` field and two entries against `healthy`: `C,es-AR`/`es-AR` must exit 1, `es-AR`/`es-AR` must exit 0. Mutation-checked both ways, since one case alone cannot tell the two failure modes apart: with `if (false)` only the first goes red, with `if (true)` only the second. 5/5.

**(a) Doc-comment hijack — moved.** The control-bundle step now sits above the #227 block, which is back against the `SMOKE_LOCALE=C` step it describes.

**(c) Timeout — scoped.** `SMOKE_TIMEOUT_MS` moved off the step and onto the control invocation inside the loop, so it no longer contradicts the comment two steps up.

**(b) Right code, wrong reason — you were right, and it was in two places.** The comment at `smoke.mjs:187` now says the positive run *fails* under a non-configurable shadow, and that what the flag prevents is a red matrix mistaken for a sanitizer regression. The same wording was in the PR body and is corrected there too.

**(d) Overclaimed coverage — fixed in the body.** It now reads `selftest.mjs` 5/5 with the mutation results, and says plainly that `pages_bundle_test.dart` 17/17 covers the sanitizer's position in the bundle, not the matrix or either new variable.

**(e) The caveat is not where #404 says it is.** #227's acceptance criteria carry no such note — the *"the CI step only exercises `C`"* line is in **#370's body**, which records what was true when it merged and is probably better left alone. What #227 does have is two criteria this PR satisfies and that are still unticked: the app rendering under `C`, `en_US` and `""`, and a CI case asserting the view mounts under an invalid locale. Ticking those looks like the honest version of that fourth criterion. Happy to do it if you agree, otherwise it stays with whoever merges.

Since the PR body has no diff, the four changes above that live in it are invisible from the Files tab — hence spelling them out.

One thing beyond the review: the `C,es-AR` case was also verified by hand, three bundles side by side under an injected `['C','es-AR']`. No sanitizer is blank; the shipped one opens in Spanish; one with the sanitizer mutated to discard the whole list opens in English, with no error and nothing logged. That middle window is the failure `SMOKE_EXPECT_LANGUAGES` exists for, and every other assertion here calls it green.

grunch (commented):
## Review — 0 CRITICAL, 0 HIGH. Nothing here blocks the merge.

Checked out the branch and read all three files in full plus the sanitizer in `web/index.html`. Where a claim was checkable I ran it rather than reading it:

| check | result |
|---|---|
| `node selftest.mjs` (5 cases, incl. the 2 new ones) | ✅ 5/5 |
| matrix loop shell semantics (`%%\|*`, `##*\|`, `${langs:-<empty>}`, log-name sanitizing) | ✅ all four cases expand correctly, no log-name collision |
| `SMOKE_NAVIGATOR_LANGUAGES=""` reaches the page as `[""]`, not as unset | ✅ confirmed in a real browser |
| no `${{ }}` interpolation inside any `run:` block | ✅ uses `$LOCALE_CRASH_SIGNATURE` from env — no script injection |

The shape is right: the per-tag negative control is what stops the matrix from going green for the wrong reason, and lifting the control build into its own step removes a real duplication.

---

## MEDIUM

### 1. `test/web/smoke/smoke.mjs` — the empty-`navigator.languages` branch still isn't covered

`forcedLanguages.split(',')` can never produce `[]`: `"".split(',')` is `[""]`. Verified. The sanitizer has an explicit branch for that state:

```js
var raw = (navigator.languages && navigator.languages.length)
  ? navigator.languages
  : [navigator.language];   // ← unreachable from this harness
```

A kiosk/embedded browser reporting `navigator.languages = []` is squarely the #227 scenario, so "every tag the sanitizer claims to handle" doesn't quite hold yet. Smallest fix: honour a sentinel (`SMOKE_NAVIGATOR_LANGUAGES="[]"` → inject an empty array), or add a separate `SMOKE_NAVIGATOR_LANGUAGE` so the singular can be driven with an empty list.

### 2. `.github/workflows/web-build.yml:52` — `LOCALE_CRASH_SIGNATURE` is pinned to two V8 message strings

`"Incorrect locale information provided|Intl.Locale constructor can't be empty"` is not API surface, it is V8 error text — and the Chromium that produces it is pinned by `PLAYWRIGHT_VERSION: "1.56.0"` thirty lines above, with nothing tying the two together. A Playwright bump that reasons those messages differently does not break the bundle; it breaks all five controls with *"failed, but not with a locale error"*, which reads as "the fix is broken". Worth a line next to `PLAYWRIGHT_VERSION` saying that pin also fixes the wording these signatures match.

### 3. `test/web/pages_bundle_test.dart` — nothing statically anchors the matrix or the two new knobs

The PR body already notes this ("it does not reach the matrix or either new variable"), but it is exactly the pattern this repo uses for every other workflow invariant (`expect(yaml, contains('selftest.mjs'))`, the enumerated fixtures). Today the entire matrix step can be deleted and `flutter test` stays 17/17 green. Two lines close it:

```dart
expect(yaml, contains('SMOKE_NAVIGATOR_LANGUAGES'));
expect(yaml, contains('SMOKE_EXPECT_LANGUAGES'));
```

### 4. The ~100s of controls are also paid on the production deploy

`deploy-pages.yml` calls `web-build.yml`, so the four matrix controls plus the single-tag control run when publishing to `mostro.network/app/` too. A negative control can contribute nothing there — it can only turn a healthy deploy red. The positive runs do carry their weight. Consider `if: ${{ !inputs.upload-pages-artifact }}` on the control halves.

### 5. The negative assertion is now duplicated — the same drift risk that justified extracting the build

The `if ... passed without the sanitizer ... fi` block, the `grep -Eq "$LOCALE_CRASH_SIGNATURE"`, and the `rm -f smoke-failure.png` now exist twice with near-identical error text. The argument for keeping the `SMOKE_LOCALE=C` step separate is right and I'd keep it — it is the one path where the tag arrives as the browser's own locale. But that justifies two *invocations*, not two copies of the *assertion*. A small helper under `test/web/smoke/`, or a shell function defined once in the job, keeps both paths and one source of truth.

---

## LOW / nit

### 6. `smoke.mjs` — the shadow isn't frozen, unlike the real property

The sanitizer itself documents why this matters: *"per the HTML spec navigator.languages returns a frozen array, so an unfrozen shadow would quietly accept a push() the genuine property rejects."* The harness installs exactly that unfrozen shadow, and returns the same array instance on every read. No consequence today (`raw.map()` doesn't mutate), but it is a divergence from the real thing in the very property under test. `Object.freeze(langs)` inside `addInitScript` costs nothing.

### 7. `smoke.mjs` — the empty case logs ambiguously

With `[""]` the run prints `✓ locale sanitized to []`, indistinguishable from a genuinely empty list (confirmed by running it). `JSON.stringify(actual)` instead of `[${actual}]` makes the log readable. Same trap in the comparison itself: `SMOKE_EXPECT_LANGUAGES=""` cannot tell `[]` from `[""]` — unused by the current matrix, but armed.

### 8. `web-build.yml:52` — `Intl.Locale` inside an ERE

The `.` is a wildcard, so a pattern whose comment declares it *"kept exact rather than loosened"* also matches `IntlXLocale`. Escape it (`Intl\.Locale`) if exactness is the point.

### 9. Implicit step-ordering dependency

Both consumers of `$control_dir` assume the step that builds it ran first. If the order is ever changed, the failure reads as `smoke.mjs` complaining "No bundle at …", not "the control bundle was never built". A `[[ -d "$control_dir" ]] || { echo "::error::control bundle missing"; exit 1; }` makes it say what it means.

### 10. `::group::` without a matching `::endgroup::` on the failure paths

Cosmetic — the log of a failing case stays collapsed inside its group.

### 11. `selftest.mjs` — the `expected: 1` cases also pass when Chromium fails to launch

I hit this by accident: with no browser installed, *"a locale left unsanitized fails the run"* passed anyway. That is a pre-existing property of the harness, and the `expected: 0` case this PR adds is precisely what compensates for it — good design. Only the `what` string of the first case promises slightly more than it proves on its own.

---

## Worth saying explicitly

- `!== undefined` rather than a truthiness check on both variables is correct, and it is *the* bug a hurried version of this PR would have shipped.
- The rewritten `configurable: true` comment from the review round is accurate: it is a false-**red** guard, not a false-green one. The first wording was wrong and the correction is the right one.
- The two new `selftest.mjs` cases hold the comparison down from both sides and each is mutation-checked, so neither is redundant. That is what separates this from "added an assertion".
- Moving `SMOKE_TIMEOUT_MS` off the step `env:` and onto the control invocation is right — the positive run needs the full 120s on a cold runner.
- Moving the strip verification next to the `sed`: without it, renaming the marker turns all five controls into rubber stamps.

Findings 1, 3 and 8 are small and self-contained if you want them in this PR; the rest are fine as follow-ups.

grunch (approved):
**Approve.** 0 CRITICAL, 0 HIGH — nothing in my findings blocks the merge.

The detailed review is in my comment above: 5 MEDIUM (the empty `navigator.languages = []` branch still uncovered, `LOCALE_CRASH_SIGNATURE` pinned to V8 message text with nothing tying it to `PLAYWRIGHT_VERSION`, no static anchor for the matrix in `pages_bundle_test.dart`, the controls also running on the production deploy, and the duplicated negative assertion) plus a handful of nits. Findings 1, 3 and 8 are small enough to fold into this PR if you want them; the rest are fine as follow-ups.

Verified rather than read: `selftest.mjs` 5/5, the matrix loop's shell expansion for all four tags, and that `SMOKE_NAVIGATOR_LANGUAGES=""` really does reach the page as `[""]`.

Note this does not unblock the PR on its own — @Catrya's `CHANGES_REQUESTED` from 2026-09-10 is still standing.

Issue #218, opened 2026-07-24
Restore: idempotent transactional reconstruction of trades, sessions and disputes


grunch:
Sub-issue of #142. **Depends on:** #215, #216, #217.

The heart of #142: turn a validated `RestoreData` snapshot into local state, as **one idempotent recovery operation**.

### Required behaviour

Applying the final snapshot is a single operation, in this order:

1. **Validate the whole payload first** — reject and abort before touching local state if anything is malformed or the version is unsupported.
2. Upsert trades / sessions / disputes from the snapshot (explicit `role` from the payload; `step` derived locally from the tested `(role, status)` mapping).
3. Persist every `order_id -> trade_index` mapping.
4. Raise `trade_key_index` to the daemon's last known index for this account (see #328 ). Not the maximum over the snapshot: that payload carries only non-finalized orders, so its maximum is a lower bound of the daemon's counter, and the next order created after the apply is rejected with `CantDo(InvalidTradeIndex)`.
5. Refresh the per-trade daemon-message subscriptions for the recovered trade keys — NIP-44 signed kind 14, this client speaks protocol v2 only and neither reads nor writes kind 1059. See 56747c7, which removed every gift-wrap path, and 67df1ab, which renamed what was left to match.)
6. Reprocess / refetch cached kind-38383 events so `is_mine` is corrected for orders that were already ingested before the restore ran.

**Only after all of those succeed** is recovery marked complete.

### Constraints

- **Idempotent**: applying the same snapshot twice must not duplicate trades, sessions, disputes or mappings.
- **Retryable**: a partial DB/key update (crash, kill, lost connection mid-apply) must be safely re-appliable without corrupting or duplicating state. Recovery is not marked complete on a partial apply.
- **Retroactive `is_mine`**: an order already sitting in the cache from a previous ingest must end up flagged as the user's. Flagging only at ingest time is not sufficient — this is an ordering gap that same-process tests do not catch.

### Acceptance criteria
- [ ] The per-trade kind-14 daemon-message subscriptions are live for all recovered trade keys after the apply
- [ ] A valid snapshot rebuilds trades, sessions and disputes with the correct role and step
- [ ] Every `order_id -> trade_index` mapping is persisted
- [ ] Orders already present in the cache before restore are corrected to `is_mine = true`
- [ ] Re-applying the same snapshot is a no-op
- [ ] An interrupted apply leaves recovery marked incomplete and succeeds on retry

grunch:
[...]

Issue #291, opened 2026-08-07
Relay liveness watchdog: detect and recover dead subscriptions


Catrya:
Split from #277 (its cause 2). Causes 1 and 3 are fixed; this tracks the
remaining one.

## Problem

The websocket can go completely silent — no Kind 38383, no kind-14, no
error, no reconnect — and the app never notices. Documented in #277: a
22-minute total silence during an active trade, recovered only by a manual
app restart; meanwhile the daemon's messages (the seller's turn to pay the
hold invoice) were missed and the order was canceled by timeout.

On Android we also observed both relays dropping together roughly every
60 seconds (connect/disconnect cycles in lockstep), which points to OS
power/network management killing the sockets rather than relay-side issues.

## Proposed direction

- Investigate what nostr-sdk already offers (relay ping / health status /
  auto-reconnect hooks) before building anything.
- A silence watchdog: if no event or EOSE arrives on the active
  subscriptions for N minutes, force a reconnect + resubscribe under the
  same stable subscription ids.
- Resubscribe on app resume (foreground transition), which is when the
  Android socket kills are most likely to have happened.

## Notes

- With #<PR de este fix> merged, a recovered subscription now replays and
  decrypts the missed kind-14 backlog correctly, so recovery is actually
  effective — before it, reconnecting was not enough.
- Related hygiene: #182 (subscription lifecycle cleanup).

codaMW:
Picking this up. Did the nostr-sdk investigation first, as suggested — findings and proposed approach:

**What the SDK (0.44.1) already handles:** the per-relay connection task auto-reconnects a dropped socket on its own (this is why our `spawn_status_monitor` sees relays cycle Connected→Disconnected→Connecting→Connected). `RelayPoolBuilder` also exposes a `Monitor` for status observation, and `try_connect(timeout)` forces connection attempts. So the ordinary drop-and-reconnect case is covered.

**What it does NOT handle — and is exactly this bug:** the *silent* case. Our status monitor polls `client.relay(url).status()`, which is connection *state*, not data *liveness*. When the OS half-kills a socket, the SDK can still report `Connected` while no events flow — the 22-minute total silence from #277 with no status transition, no error, no reconnect. Nothing in the SDK detects "connected but silent."

**Proposed approach:**
1. Track `last_event_at` on the pool, bumped by the existing notification loops (orders/messages already see every `RelayPoolNotification::Event`).
[...]

PR #407, opened 2026-09-09
feat: single-owner registry for per-trade subscriptions

### LINKED ISSUE #325: Single owner for per-trade subscription lifecycle (remaining scope of #182)
#182 is done as far as the leak goes: PR #255 unsubscribes on exit and gives the per-trade and single-order watchers deterministic ids. Its third task — the optional central registry — is not done, and this issue carries it.

## Status: not broken today, required before #218

Nothing is broken right now. Each `create_order` / `take_order` calls `derive_trade_key()` first, so every per-trade subscription gets a unique pubkey and therefore a unique id; `subscribe_single_order` has a single caller, in the successful-take path. Two watchers over the same subscription id cannot currently happen.

#324 does not change that either: it forces a pool `disconnect`/`connect`, which drives the existing `Online` path in `api::nostr` (order book and chat subscriptions). It does not re-arm per-trade watchers.single caller, in the successful-take path. Two watchers over the same subscription id cannot currently happen.

#324 does not change that either: it forces a pool `disconnect`/`connect`, which drives the existing `Online` path in `api::nostr` (order book and chat subscriptions). It does not re-arm per-trade watchers.

## Confirmed trigger: #218

#218 (`priority: high`, milestone M4) makes it reachable, by its own specification. Step 5 of the apply sequence is:

> 5. Refresh gift-wrap subscriptions for the recovered trade keys.

with the matching acceptance criterion:

> - [ ] Gift-wrap subscriptions are live for all recovered trade keys after the apply

and the whole operation is required to be **idempotent and retryable** — "applying the same snapshot twice must not duplicate trades, sessions, disputes or mappings", and "a partial DB/key update (crash, kill, lost connection mid-apply) must be safely re-appliable".

Two applies of the same snapshot therefore re-arm a watcher for a trade key that already has one: two tasks owning `mostro-trade-<pubkey>`. The failure is not theoretical — the older watcher's exit path unsubscribes the newer one's live subscription and purges its pending request, which surfaces as `NoDaemonResponse` on an order the daemon actually accepted. A restore is exactly the moment a user cannot afford that.

So this is an **ordering constraint, not a backlog item**: it has to land before, or together with, #218 — and it stays small if it does, since all #218 needs is for `subscribe_daemon_messages` to be idempotent per trade key under a single owner.

(Note for #218: its step 5 and acceptance criteria say "gift-wrap subscriptions", which no longer exists — `main` renamed that transport in `56747c7` / `67df1ab`. The subscription meant is the per-trade daemon-message one.)

## Other paths that would reach the same place

- **#291** — its stated direction includes "force a reconnect + resubscribe under the same stable subscription ids" and "Resubscribe on app resume". #324 covers only the pool half; a per-trade re-arm is the trigger.
- **#308** — resume resync / hydration, the same thing from the foreground transition.
- **#147** — a background service keeping subscriptions alive.

## The shape to copy

#182 dismissed the registry as "likely overkill" by comparing against v1's `lib/features/subscriptions/` — a manager with four subscription types, session listeners and bulk filters. Mostrix is the closer reference (same language, same `nostr-sdk`, per-trade subscriptions like ours), and there it is three maps plus one function:

- `subscribed_pubkeys: HashSet<PublicKey>` — idempotence by set membership, not by id
- `pubkey_to_subscription: HashMap<PublicKey, SubscriptionId>`
- `subscription_to_order: HashMap<SubscriptionId, (order_id, trade_index)>`
- `unsubscribe_dm_listener_subscriptions(client)` — tears down only the DM-listener subscriptions and leaves the order/dispute scheduler ones intact

What it buys is a **single owner** of subscription lifecycle. v1 gets the same property from one subscription per type; Mostrix from one DM-router task owning the maps. Neither needs per-task ownership guards, because in neither does more than one task own a subscription.

Our deterministic ids are an advantage over Mostrix here: the id is recomputable from the trade pubkey, so a registry would not need `pubkey_to_subscription` at all.

If PR #255 merges with its generation guard still in place (`SUBSCRIPTION_GENERATIONS` / `claim_subscription` / `owns_subscription`), that guard is a per-task stand-in for this ownership and should be removed as part of this work — it also claims *after* subscribing, which leaves the window it is meant to close open.

## Where it should live

Not in `api/orders.rs` — see #120. The relay/subscription layer (`rust/src/nostr/`) is the right home.

## Acceptance criteria

- [ ] Subscribing twice for the same trade key is idempotent: one relay-side subscription, one owner
- [ ] Re-arming a watcher for a trade key that already has a live one neither tears down the live subscription nor purges its pending request
- [ ] Teardown is targeted: closing per-trade subscriptions leaves the order-book and global DM subscriptions intact
- [ ] Applying the same restore snapshot twice (#218) leaves exactly one live subscription per recovered trade key

---

Supersedes the subscription-manager half of the closed #144 (its RequestId half shipped in #172).

Forte11Cuba:
Close #325 

Every per-trade watcher's exit path runs a destructive pair — `unsubscribe` on its deterministic id plus `purge_pending_request` — unconditionally. Today that is safe only by accident: every caller of `subscribe_daemon_messages` derives a fresh trade key first, so two watchers can never share an id. The restore apply (#218) breaks that invariant by design: it must be idempotent, and re-applying a snapshot re-arms watchers for trade keys that already have one. The older watcher's exit would then kill the newer one's live subscription and purge its pending record — surfacing as `NoDaemonResponse` on an order the daemon actually accepted, during a restore.

## Change

New `rust/src/nostr/subscriptions.rs` — a single-owner registry for the per-trade subscription lifecycle, following the repo's own `ACTIVE_CHATS` precedent (`messages.rs`) rather than importing Mostrix's shape:

- **Idempotence by membership**: `try_claim` admits exactly one owner per trade key. A second claim bounces — the live subscription and its pending record stay untouched.
- **A bounce is a lease refresh**: the bounce marks the key re-armed; the owner's idle-timeout exit consumes the mark and resets its timer instead of dismantling. Without this, a re-arm landing at minute 29 of the 30-minute idle window would be "covered" for seconds.
- **Teardown is atomic w.r.t. claims**: the destructive pair runs under the registry lock, so a concurrent claim either bounces before it (owner keeps running) or lands after it (key is free, subscribes from scratch) — never in between. Note: `ACTIVE_CHATS` releases *before* unsubscribing and leaves exactly that window open; this module documents why not to copy that ordering.
- **Shutdown vs idle**: Shutdown/closed-channel exits tear down unconditionally even with a re-arm mark pending — the watcher's receiver is dead, honoring the mark would spin on a closed channel. Post-reconnect coverage belongs to the re-arm paths (#218/#291).

`daemon_message_subscription_id` moves into the module (id + ownership together, in line with #120). `subscribe_single_order` is out of scope: single caller, not on the #218 path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of per-trade message subscriptions by preventing duplicate watchers and restoring idle subscriptions.
  - Pending requests are preserved during subscription recovery, with cleaner shutdown and channel-closure handling.
  - Improved invoice update handling, including bond requests and settled hold-invoice messages.
  - Improved cancellation and cleanup behavior across maker, inactive-take, and active-trade states.
  - Subscription identifiers are generated consistently to reduce conflicts and improve message routing.

- **Documentation**
  - Clarified subscription recovery, teardown, relay acceptance, and cancellation behavior.

- **Tests**
  - Added coverage for subscription races, recovery, cleanup, bond handling, and identifier uniqueness.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1600,9 +1600,12 @@ pub(crate) async fn subscribe_daemon_messages(
     // outlives this caller's interest. If the owner is still mid-setup, the
     // claim parks until the REQ is live (or takes over if that setup fails),
     // so a bounce always means real coverage. Claiming first means every
-    // early return below must release, and the success path must mark_live.
+    // early return below must release the guard and the success path must
+    // consume it via mark_live; a panic anywhere in between frees the claim
+    // from the guard's Drop instead of wedging the key in Setup.
     let trade_pubkey_hex = trade_pubkey.to_hex();
-    if !crate::nostr::subscriptions::claim(&trade_pubkey_hex, trade_index).await {
+    let Some(setup) = crate::nostr::subscriptions::claim(&trade_pubkey_hex, trade_index).await
+    else {
         crate::api::logging::blog_info(
             "orders",
             format!(
@@ -1611,20 +1614,20 @@ pub(crate) async fn subscribe_daemon_messages(
             ),
         );
         return;
-    }
+    };
 
     let recipient_keys = match crate::api::identity::get_active_trade_keys(trade_index).await {
         Ok(k) => k,
         Err(e) => {
             log::error!("[orders] subscribe_daemon_messages: no trade keys: {e}");
-            crate::nostr::subscriptions::release(&trade_pubkey_hex).await;
+            crate::nostr::subscriptions::release(setup).await;
             return;
         }
     };
 
     let Ok(pool) = crate::api::nostr::get_pool() else {
         log::warn!("[orders] subscribe_daemon_messages: relay pool not initialized");
-        crate::nostr::subscriptions::release(&trade_pubkey_hex).await;
+        crate::nostr::subscriptions::release(setup).await;
         return;
     };
     let client = pool.client();
@@ -1634,7 +1637,7 @@ pub(crate) async fn subscribe_daemon_messages(
             Ok(pk) => pk,
             Err(e) => {
                 log::error!("[orders] subscribe_daemon_messages: invalid mostro pubkey: {e}");
-                crate::nostr::subscriptions::release(&trade_pubkey_hex).await;
+                crate::nostr::subscriptions::release(setup).await;
                 return;
             }
         };
@@ -1668,7 +1671,7 @@ pub(crate) async fn subscribe_daemon_messages(
     let sub_id = crate::nostr::subscriptions::daemon_message_subscription_id(&trade_pubkey_hex);
     if let Err(e) = client.subscribe(filter).with_id(sub_id).await {
         log::warn!("[orders] subscribe_daemon_messages subscribe failed: {e}");
-        crate::nostr::subscriptions::release(&trade_pubkey_hex).await;
+        crate::nostr::subscriptions::release(setup).await;
         return;
     }
 
@@ -1677,7 +1680,7 @@ pub(crate) async fn subscribe_daemon_messages(
     // above was obtained before subscribing, so events arriving before the
     // watcher task spawns below sit buffered in the channel — nothing leaks
     // in the gap.
-    crate::nostr::subscriptions::mark_live(&trade_pubkey_hex).await;
+    crate::nostr::subscriptions::mark_live(setup).await;
 
     crate::api::logging::blog_info(
         "orders",
@@ -5589,10 +5592,12 @@ mod tests {
         let trade_pubkey = keys.public_key();
         let hex = trade_pubkey.to_hex();
 
-        assert!(crate::nostr::subscriptions::claim(&hex, 9).await);
+        let guard = crate::nostr::subscriptions::claim(&hex, 9)
+            .await
+            .expect("first claim owns the key");
         // Live, not Setup: against a mid-setup owner the re-arm below would
         // (correctly) park instead of bouncing, and this test would hang.
-        crate::nostr::subscriptions::mark_live(&hex).await;
+        crate::nostr::subscriptions::mark_live(guard).await;
         pending_requests().lock().unwrap().insert(
             hex.clone(),
             PendingRequest {
@@ -5610,12 +5615,12 @@ mod tests {
             "a bounced re-arm must not purge the live watcher's pending request"
         );
         assert!(
-            !crate::nostr::subscriptions::claim(&hex, 9).await,
+            crate::nostr::subscriptions::claim(&hex, 9).await.is_none(),
             "the original owner must still hold the key after a bounced re-arm"
         );
 
         pending_requests().lock().unwrap().remove(&hex);
-        crate::nostr::subscriptions::release(&hex).await;
+        crate::nostr::subscriptions::teardown(&nostr_sdk::prelude::Client::default(), &hex).await;
     }
 
     /// Nothing ever displays a stranger's finished order — the book filters to
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1668,9 +1668,15 @@ pub(crate) async fn subscribe_daemon_messages(
         .author(mostro_pubkey)
         .pubkey(trade_pubkey)
         .limit(0);
+    // `subscribe_accepted`, not a bare subscribe: the SDK reports a REQ
+    // that failed on every relay as an `Ok`, and a failed REQ is removed
+    // from the relay's registry, beyond reconnect resubscription's reach.
+    // Marking that Live would promise coverage that never exists — claims
+    // would bounce off it while the caller's request dies at its 10 s
+    // timeout with `NoDaemonResponse`.
     let sub_id = crate::nostr::subscriptions::daemon_message_subscription_id(&trade_pubkey_hex);
-    if let Err(e) = client.subscribe(filter).with_id(sub_id).await {
-        log::warn!("[orders] subscribe_daemon_messages subscribe failed: {e}");
+    if let Err(e) = subscribe_accepted(&client, sub_id, filter).await {
+        log::warn!("[orders] subscribe_daemon_messages: {e}");
         crate::nostr::subscriptions::release(setup).await;
         return;
     }
@@ -3973,26 +3979,20 @@ fn relay_list_subscription_id() -> nostr_sdk::prelude::SubscriptionId {
     nostr_sdk::prelude::SubscriptionId::new("mostro-relay-list")
 }
 
-/// Point the long-lived subscription `id` at `filter`, replacing whatever it
-/// carried before.
-///
-/// nostr-sdk 0.45 refuses a subscribe whose id already exists and keeps the
-/// old filters, so the id is closed first (a no-op when it was never open).
-/// The brief gap between CLOSE and REQ loses nothing: a node switch refetches
-/// the book right after, and the Kind-14 feed has no `since`, so its REQ
-/// replays history.
+/// Subscribe `filter` under `id`, failing when no relay accepted the REQ.
 ///
 /// The SDK reports per-relay failures inside an `Ok` output, which is how a
-/// rejected re-subscribe used to pass for a live one. No relay accepting it is
+/// rejected subscribe used to pass for a live one. Empty success is not a
+/// transient state, either: a REQ that failed on a relay is *removed* from
+/// that relay's subscription registry (nostr-sdk 0.45,
+/// `subscribe_long_lived`), so reconnect resubscription cannot revive it —
+/// the subscription exists nowhere and never will. No relay accepting it is
 /// an error here; a partial failure is logged.
-async fn replace_subscription(
+async fn subscribe_accepted(
     client: &nostr_sdk::prelude::Client,
     id: nostr_sdk::prelude::SubscriptionId,
     filter: nostr_sdk::prelude::Filter,
 ) -> Result<()> {
-    if let Err(e) = client.unsubscribe(&id).await {
-        log::warn!("[orders] closing {id} before re-subscribing failed: {e}");
-    }
     let output = client
         .subscribe(filter)
         .with_id(id.clone())
@@ -4015,6 +4015,25 @@ async fn replace_subscription(
         );
     }
     Ok(())
+}
+
+/// Point the long-lived subscription `id` at `filter`, replacing whatever it
+/// carried before.
+///
+/// nostr-sdk 0.45 refuses a subscribe whose id already exists and keeps the
+/// old filters, so the id is closed first (a no-op when it was never open).
+/// The brief gap between CLOSE and REQ loses nothing: a node switch refetches
+/// the book right after, and the Kind-14 feed has no `since`, so its REQ
+/// replays history.
+async fn replace_subscription(
+    client: &nostr_sdk::prelude::Client,
+    id: nostr_sdk::prelude::SubscriptionId,
+    filter: nostr_sdk::prelude::Filter,
+) -> Result<()> {
+    if let Err(e) = client.unsubscribe(&id).await {
+        log::warn!("[orders] closing {id} before re-subscribing failed: {e}");
+    }
+    subscribe_accepted(client, id, filter).await
 }
 
 /// (Re)subscribe the order-book (Kind 38383) and Mostro-reply (Kind 14)
@@ -5486,6 +5505,43 @@ mod tests {
         assert!(
             result.is_err(),
             "a subscription no relay accepted must not pass for a live one"
+        );
+    }
+
+    /// PR #407 CodeRabbit: the per-trade daemon REQ shares that guard via
+    /// `subscribe_accepted` — no relay accepting it must be an error, so
+    /// `subscribe_daemon_messages` releases its claim instead of marking a
+    /// coverage-less subscription Live (a failed REQ is also removed from
+    /// the relay's registry, beyond reconnect resubscription's reach).
+    #[tokio::test]
+    async fn a_per_trade_subscription_no_relay_accepts_is_an_error() {
+        use nostr_sdk::local_relay::MockRelay;
+        use nostr_sdk::prelude::{Client, Keys};
+
+        let relay = MockRelay::run().await.expect("mock relay");
+        let client = Client::new();
+        client
+            .add_relay(relay.url().await)
+            .await
+            .expect("add relay");
+        // Added but never connected: the only relay rejects the REQ.
+
+        let trade_pubkey = Keys::generate().public_key();
+        let filter = nostr_sdk::prelude::Filter::new()
+            .kind(nostr_sdk::prelude::Kind::PrivateDirectMessage)
+            .author(Keys::generate().public_key())
+            .pubkey(trade_pubkey)
+            .limit(0);
+        let result = subscribe_accepted(
+            &client,
+            crate::nostr::subscriptions::daemon_message_subscription_id(&trade_pubkey.to_hex()),
+            filter,
+        )
+        .await;
+
+        assert!(
+            result.is_err(),
+            "a per-trade REQ no relay accepted must not reach mark_live"
         );
     }
 
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1595,12 +1595,14 @@ pub(crate) async fn subscribe_daemon_messages(
 ) {
     // ── Synchronous setup: awaited by the caller ──
     // Single-owner claim before anything else (#325): if a watcher already
-    // owns this trade key, this call is a lease refresh — the live
-    // subscription and its pending record stay untouched, and the owner
-    // outlives this caller's interest. Claiming first means every early
-    // return below must release.
+    // owns this trade key with a live REQ, this call is a lease refresh —
+    // the subscription and its pending record stay untouched, and the owner
+    // outlives this caller's interest. If the owner is still mid-setup, the
+    // claim parks until the REQ is live (or takes over if that setup fails),
+    // so a bounce always means real coverage. Claiming first means every
+    // early return below must release, and the success path must mark_live.
     let trade_pubkey_hex = trade_pubkey.to_hex();
-    if !crate::nostr::subscriptions::try_claim(&trade_pubkey_hex).await {
+    if !crate::nostr::subscriptions::claim(&trade_pubkey_hex, trade_index).await {
         crate::api::logging::blog_info(
             "orders",
             format!(
@@ -1669,6 +1671,13 @@ pub(crate) async fn subscribe_daemon_messages(
         crate::nostr::subscriptions::release(&trade_pubkey_hex).await;
         return;
     }
+
+    // The REQ is active: advance the claim to Live so claims parked on this
+    // key stop waiting and bounce against real coverage (#325). The `rx`
+    // above was obtained before subscribing, so events arriving before the
+    // watcher task spawns below sit buffered in the channel — nothing leaks
+    // in the gap.
+    crate::nostr::subscriptions::mark_live(&trade_pubkey_hex).await;
 
     crate::api::logging::blog_info(
         "orders",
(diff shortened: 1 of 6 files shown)


grunch:
This branch conflicts with `main` since #408 (`d49f3c1`) merged; GitHub reports it as not mergeable. Both PRs edit `rust/src/api/orders.rs`.

What to do on rebase:

- Rebase onto `main` and resolve `rust/src/api/orders.rs`. #408 changed three regions there: the daemon-message status sync (after `status_for_action`, it now spawns `confirm_payout_completion` when a trade enters `SettledHoldInvoice`), the stale sweep (`sweep_action` gained a `local_status` argument and a `SyncSuccess` arm, and `run_stale_sweep_once` also examines `SettledHoldInvoice` trades), and two new helpers next to `fetch_public_order_status` (`newest_book_status`, `apply_payout_completed`). Your hunks around `subscribe_daemon_messages` and `orders_subscription_id` are adjacent to those, so expect textual conflicts rather than semantic ones.
- One semantic point to keep in mind: #408 relies on the per-trade public-status subscription your registry now owns. The seller learns of the completed payout only from the public kind-38383 `success` event (the daemon sends `PurchaseCompleted` to the buyer alone), and `confirm_payout_completion` exists precisely because that subscription sometimes misses the event. If the registry changes when a watcher is torn down, make sure the watcher for a trade that just reached `SettledHoldInvoice` stays alive until the trade is terminal, or the fallback in #408 will be doing all the work.
- Re-run `cd rust && cargo clippy -- -D warnings && cargo clippy --target wasm32-unknown-unknown -- -D warnings && cargo test` after resolving; the web target is checked by CI as well.

rust/src/api/orders.rs
```diff
@@ -1550,16 +1550,35 @@ pub(crate) async fn subscribe_daemon_messages(
     trade_index: u32,
 ) {
     // ── Synchronous setup: awaited by the caller ──
+    // Single-owner claim before anything else (#325): if a watcher already
+    // owns this trade key, this call is a lease refresh — the live
+    // subscription and its pending record stay untouched, and the owner
+    // outlives this caller's interest. Claiming first means every early
```
grunch, line 1556: **Coverage promised before it exists.** The claim happens before `get_active_trade_keys`, `get_pool` and `client.subscribe`. A second caller that bounces inside that window returns immediately believing a watcher is live. Two consequences:

1. If this owner's setup then fails, `release` removes the entry and silently drops the re-arm mark. The bounced caller has already inserted its pending record and goes on to publish, and nobody is listening: the same `NoDaemonResponse` this PR sets out to prevent.
2. Even on a successful setup, the bounced caller can publish before the `limit(0)` live-only REQ is active and miss the reply.

Suggestion: make the registry state `Setup | Live { rearmed }` and have the bounce await a `tokio::sync::Notify` until the owner reaches `Live`, retrying the claim if the owner released instead. Minimal alternative: have `release` return whether a mark was pending and log it as a warning so the loss is at least visible.

rust/src/api/orders.rs
```diff
@@ -5316,6 +5338,40 @@ mod tests {
...
+                request_id: 5,
+                trade_index: 9,
+                kind: PendingRequestKind::Take,
+                tx: None,
+            },
+        );
+
+        subscribe_daemon_messages(trade_pubkey, 9).await;
+
+        assert!(
```
grunch, line 5366: **This assertion holds with or without the registry.** In the test environment there is no identity and no relay pool, so a non-bouncing call exits through `get_active_trade_keys` or `get_pool`, and neither path purges the pending record. The test therefore cannot tell a bounce from an ordinary setup failure.

Add, right after the call:
```rust
assert!(
    !crate::nostr::subscriptions::try_claim(&hex).await,
    "the original owner must still hold the key after a bounced re-arm"
);
```
That fails if the entry was released (non-bounce path) and passes only when the call actually bounced.

rust/src/nostr/subscriptions.rs
```diff
@@ -0,0 +1,288 @@
...
+async fn dismantle(
+    client: &nostr_sdk::prelude::Client,
+    trade_pubkey_hex: &str,
+    map: &mut HashMap<String, bool>,
+) {
+    // Drop the relay-side REQ. Without this the task exits but the
+    // subscription lives on: relays cap concurrent REQs, and once past the
+    // cap they answer CLOSED — which can take the order-book feed down
+    // with it.
+    if let Err(e) = client
```
grunch, line 153: Non-blocking. I checked nostr-relay-pool 0.44: `Relay::unsubscribe` only takes the in-process subscriptions `RwLock` and queues the CLOSE, it never waits on the socket, so the module doc's "a short send" is accurate. Since this global lock serializes claims for every trade, worth a one-line note that a nostr-sdk bump must re-verify `unsubscribe` still does not block on network.

grunch (changes requested):
Reviewed the registry design and ran the branch locally: the 6 new tests pass and clippy adds no warnings in the touched files. The single-owner model is sound and I found no teardown path that bypasses the registry. Two items block the merge, the rest is non-blocking.

**Blocking**
- A bounce that lands while the owner is still between `try_claim` and `client.subscribe` is promised coverage that may never exist (inline comment).
- The new orders.rs test passes with or without the registry check, so it does not prove the bounce (inline comment).

**Non-blocking**
- Specs: CLAUDE.md asks for the matching contract to follow behavior changes. Nothing in `specs/` or `.specify/` describes the per-trade daemon subscription lifecycle (30-minute idle, single owner, lease refresh); only the chat listener has such a note (`specs/004-mostro-p2p-client/contracts/messages.md:62`). A short paragraph there or in the orders contract would keep them in sync.
- Branch is 9 commits behind `main`. No conflict (main only touched `create_order`), but please rebase before merge.
- Nit: the bounce ignores `trade_index`; the owner keeps the recipient keys of its own index. Same pubkey implies same index today, so a `debug_assert_eq!` or a log when they differ would make the assumption explicit.
- Nit: `&hex[..8]` in orders.rs vs `[..8.min(len)]` in the new module.

rust/src/nostr/subscriptions.rs
```diff
@@ -0,0 +1,544 @@
...
+                short(trade_pubkey_hex)
+            ),
+        );
+    }
+
+    // The subscription bounds the pending record's lifetime: once no reply
+    // can be delivered here anymore, a still-unconsumed record (request timed
+    // out and no genuine late reply ever arrived) is dead state — drop it,
+    // whatever attempt it belongs to.
+    crate::mostro::pending::purge_pending_request(trade_pubkey_hex);
```
grunch: 🔴 **Potential issue — a parked re-arm's pending record is purged out from under it.**

The atomicity argument in the module docs ("a concurrent claim either bounces first or waits and finds the key free — never in between") covers the relay-side REQ, but not the pending record. `purge_pending_request` removes *whatever* record sits under the key, and the two production callers register theirs **before** they call `subscribe_daemon_messages`:

- `create_order` (orders.rs ~L731-748): inserts `PendingRequest { tx: Some(conf_tx), .. }`, then subscribes.
- `take_order` (orders.rs ~L980-995): same ordering, "so the reply cannot race the bookkeeping".

Interleaving on a key `K` that already has a watcher (the #218 / #291 re-arm case this registry exists for):

1. Owner `W1` hits its idle timeout, `teardown_or_rearm` takes the registry lock, `rearmed == false`, enters `dismantle`.
2. Caller `C` inserts its record for `K` with a live waiter, calls `subscribe_daemon_messages(K)` → `claim` blocks on the lock.
3. `dismantle` runs `purge_pending_request(K)` → `C`'s record (live `tx`) is gone. Entry removed, lock released.
4. `C` claims fresh, subscribes, publishes. The daemon's reply reaches the new watcher, `take_matching_*` finds no record, the reply is dropped as stale, `C` times out with `NoDaemonResponse` on a request the daemon accepted.

Today no caller re-arms a covered key so this is unreachable, but it is reachable by construction the moment #218 lands — and the purge comment "no reply can be delivered to any attempt on this key" is false while a caller is parked on this very lock waiting to subscribe.

Minimal fix that keeps the registry's contract: only purge a record nobody is waiting on. A live `tx` means a caller is mid-request and will (re)subscribe; a detached record (the 10 s timeout sets `tx = None`) is the dead state this purge is meant for. The existing `pending_record_survives_a_lease_refresh_and_dies_with_teardown` test inserts `tx: None`, so it keeps passing.

```suggestion
    // The subscription bounds the pending record's lifetime: once no reply
    // can be delivered here anymore, a still-unconsumed record (request timed
    // out and no genuine late reply ever arrived) is dead state — drop it,
    // whatever attempt it belongs to. A record with a *live* waiter is not
    // dead: its caller registered it before calling
    // `subscribe_daemon_messages` (create_order / take_order ordering) and
    // may be parked on this very lock, about to subscribe from scratch —
    // purging it here would strand that caller with a NoDaemonResponse.
    crate::mostro::pending::purge_detached_pending_request(trade_pubkey_hex);
```

with, in `mostro/pending.rs`:

```rust
/// Like `purge_pending_request`, but leaves a record whose waiter is still
/// attached: that caller is mid-request and owns the record's lifetime.
pub(crate) fn purge_detached_pending_request(trade_pubkey_hex: &str) {
    if let Ok(mut m) = pending_requests().lock() {
        if m.get(trade_pubkey_hex).is_some_and(|p| p.tx.is_none()) {
            m.remove(trade_pubkey_hex);
        }
    }
}
```

Alternative: flip the two callers to register **after** `subscribe_daemon_messages` returns (safe now — a bounce only returns against a live REQ and `limit(0)` means nothing can be delivered before the publish) and document that ordering as a registry precondition. Either way, please add a test for step 1-4: claim + `mark_live`, insert a record with `tx: Some`, run `teardown_or_rearm` with no mark, assert the record survives.

rust/src/nostr/subscriptions.rs
```diff
@@ -0,0 +1,544 @@
...
+        // the re-check below, closing the race where the owner's
+        // `notify_waiters` (always sent under the registry lock) fires
+        // between our check above and the await: either the transition
+        // happened before the re-check (we see it and retry immediately) or
+        // after (we are already registered and the notify wakes us).
+        let notified = notify.notified();
+        tokio::pin!(notified);
+        notified.as_mut().enable();
+        if parked_on_current_setup(trade_pubkey_hex, &notify).await {
+            notified.await;
```
grunch, line 243: 🟡 **Unbounded park turns a leaked `Setup` entry into a permanently hung API call.**

The cancellation caveat above says "no caller does that today", but the future does not have to be *dropped* to leak: a panic anywhere between `claim` and `mark_live`/`release` (`get_active_trade_keys`, `get_pool`, the pubkey parse, `client.subscribe`) unwinds `subscribe_daemon_messages`, FRB reports the error to Dart, and the entry stays in `Setup` forever. From then on every `claim` for that key — and therefore every `take_order`/`create_order` that awaits it — blocks here indefinitely, with nothing logged and no recovery short of a restart.

Two cheap ways to bound the blast radius, either is fine:

1. Bound the park and take over on expiry — setup is a handful of local awaits plus one channel push, so a generous ceiling is safe:
   ```rust
   const SETUP_PARK_CEILING: Duration = Duration::from_secs(15);
   if crate::rt::time::timeout(SETUP_PARK_CEILING, notified).await.is_err() {
       crate::api::logging::blog_warn("orders", format!(
           "claim for trade={} parked {}s on a Setup owner that never advanced — evicting it",
           short(trade_pubkey_hex), SETUP_PARK_CEILING.as_secs()));
       registry().lock().await.remove(trade_pubkey_hex); // loop retries and claims fresh
   }
   ```
2. An RAII guard returned by `claim` whose `Drop` releases a still-`Setup` entry (with `try_lock`, falling back to a spawned release) — covers both panic and future cancellation and removes the caveat entirely.

Not blocking on its own, but worth doing in this PR since the registry is what introduces the hang.

grunch (changes requested):
Round 2 review of d1ee715. I merged the branch onto current `main` locally and ran the full Rust gate: `cargo test` (468 passed), `cargo clippy -- -D warnings` (native) and `cargo clippy --target wasm32-unknown-unknown -- -D warnings` are all green. I also traced `Client::unsubscribe` in the locked nostr-sdk 0.45.2: `Relay::unsubscribe` → `_unsubscribe_long_lived_subscription` → `send_msg(Close, None)` → `try_send` on a bounded mpsc, so the module's "never a socket wait" claim holds and the lock discipline is sound.

Round-1 items are addressed: the `Setup | Live` state with parked claims closes the claim-to-subscribe window, the orders.rs test now asserts ownership retention, the spec has the lifecycle paragraph, and the trade-index mismatch is logged.

**Blocking (1)**
- `dismantle` purges the pending record for the key unconditionally, but both `create_order` and `take_order` insert their record (with a live `tx`) *before* calling `subscribe_daemon_messages`. A re-arm that parks on the registry lock while the owner is dismantling loses its record, subscribes from scratch, publishes, and its reply finds nothing to match — the exact `NoDaemonResponse` this PR sets out to prevent. Inline comment with a suggested fix.

**Non-blocking**
- `claim` parks with no upper bound. The cancellation caveat is documented, but a panic between `claim` and `mark_live`/`release` (FRB unwinds the future and reports the error to Dart) leaks a `Setup` entry, and every later claim for that key — and therefore every `take_order`/`create_order` on it — hangs forever. Inline comment.
- `specs/004-mostro-p2p-client/plan.md:125-127` draws the `rust/src/nostr/` module tree; `subscriptions.rs` should be listed there alongside `relay_pool.rs`.
- Branch is 31 commits behind `main`. The merge is clean and green (above), but please rebase before merge so CI validates what actually ships.

rust/src/api/orders.rs
```diff
@@ -1667,109 +1699,125 @@ pub(crate) async fn subscribe_daemon_messages(
...
+                        last_activity = crate::rt::time::Instant::now();
+                        continue 'watch;
+                    }
+                }
+                Exit::Shutdown => {
+                    crate::nostr::subscriptions::teardown(&unsub_client, &trade_pubkey_hex).await;
+                }
+            }
+            break 'watch;
         }
```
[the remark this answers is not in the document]
Forte11Cuba: Not applicable: the global mostro-dm feed covers every derived trade key regardless of trade status (seed_global_dm_coverage derives indices 1..=max) and dispatches through the same dispatch_mostro_message (handle_global_daemon_message). A late AddInvoice after the per-trade watcher's idle teardown arrives via that feed — the per-trade REQ is a redundant fast-path, and keeping one REQ per non-terminal trade indefinitely is exactly what the idle timeout avoids (relays cap concurrent REQs and answer CLOSED past the cap).

grunch (dismissed):
Round 3 review of `fa553ba`. **Approve** — my round-2 blocker is resolved and nothing below blocks the merge.

Ran the gate on the branch rather than reading it:

| check | result |
|---|---|
| `cargo test` | ✅ 483 passed, 0 failed, 24 ignored |
| `cargo clippy --all-targets -- -D warnings` | ✅ clean |
| `./scripts/frb-generate.sh --check` | ✅ no binding drift (the touched items are `pub(crate)`) |
| create/take/restore register the pending record **before** `subscribe_daemon_messages` | ✅ verified at all four call sites (`orders.rs:749, 995, 1414, 5305`) — the detached-only purge rests on this and it holds |
| `Relay::unsubscribe` frees the id even when the CLOSE send fails | ✅ `inner.rs:1819` removes from the map *before* `send_msg`, so a warned-and-ignored unsubscribe still leaves the id reusable |

Round-2 items all addressed: `dismantle` now purges detached records only (with a test that pins both halves), the RAII `SetupGuard` frees an abandoned setup, `mark_live`/`release` take the key only once they hold the lock (with a cancellation test that parks each on a held lock and aborts it there), `plan.md` lists the module, and the contract carries the lifecycle paragraph. The `Notify` handling is genuinely correct — `enable()` before the re-check, and `Arc::ptr_eq` to pin the *entry* rather than the state, which is the ABA hole most implementations of this leave open.

---

## MEDIUM

### 1. The justification for treating empty success as fatal is wrong for one of the two failure modes — and it is copied into the spec

`subscribe_accepted` (`orders.rs`) and `contracts/orders.md` both assert that a rejected REQ is *removed* from the relay's registry, so "the subscription exists nowhere and never will". In nostr-sdk 0.45.2 that holds for only one of the paths that can put a relay in `output.failed`:

- `relay/api/subscribe.rs:117` — `send_msg` fails → `remove_subscription(&id)`. Claim holds.
- `relay/inner.rs:340` — `add_long_lived_subscription` returns `Err("subscription ID already exists")` **without touching the existing entry**. The subscription very much still exists; it is the *old* one.

And the SDK's own comment at `subscribe.rs:104` says the pre-send insert is there "so an immediate CLOSED can still mark the subscription for retry" — a REQ the relay CLOSEs stays registered *and* lands in `output.success`, so empty-success is not the only coverage-less outcome either.

The code decision (fail on empty success, release the claim) is right regardless. It is the stated reason that overstates, and because the sentence was lifted verbatim into an authoritative contract it will be trusted later. Narrow it to the send-failure case, or drop the "never will" and justify it as "no relay confirmed, so nothing may be marked Live".

### 2. `SetupGuard::drop`'s contended path is fire-and-forget, and that is where the wedge comes back

`crate::rt::spawn` is `tokio::spawn` on native (`rt.rs:14-19`). Two gaps the uncontended `try_lock` path does not have:

- **It can panic where a panic is fatal.** `tokio::spawn` panics with no runtime in context, and this `Drop` runs during unwind — the very case the doc names ("a panic unwinding the setup"). A panic inside a destructor while unwinding aborts the process instead of surfacing to Dart through FRB. Today every drop happens on a runtime thread, so it is safe by circumstance, not by construction.
- **The release may never run.** A spawned task that is never polled (runtime tearing down, wasm executor gone) leaves the entry in `Setup` forever, and every later `claim` for that key — so every `take_order`/`create_order` on it — parks with no timeout. That is exactly the wedge the guard was added to prevent, reintroduced on the contended branch.

`blog_warn` itself is panic-free (`logging.rs:303` uses `.lock().ok()`), so the fix is only about the spawn. Cheapest robust shape: make the entry self-healing rather than relying on the release landing — stamp `Entry` with an `Instant` at claim time and have `claim` take over a `Setup` entry older than some bound instead of parking on it forever. Then `Drop` is best-effort by design and neither gap matters.

### 3. `mark_live` silently no-ops when the entry has vanished

```rust
if let Some(entry) = map.get_mut(&key) { ... }   // no else
```

If the entry is ever gone by then, `mark_live` returns quietly and the watcher runs with **no registry entry** — so the next `claim` for that key creates a second owner, which is precisely the state this module exists to make impossible. I could not find a reachable path (nothing removes a `Setup` entry but its own guard, and teardown only ever runs for a live watcher), so this is defensive. But it is the module's central invariant and the only one with no assertion behind it: `else { debug_assert!(false, ...); blog_warn(...) }` costs nothing and turns a silent duplicate-owner into a visible bug.

---

## LOW / nit

### 4. `&trade_pubkey_hex[..8]` in the new bounce log

`orders.rs`, the "already live … re-arm refreshed its lease" message. The module has `short()` for exactly this and the rest of the new code uses it — this is the round-1 nit reappearing in newly added code. Same panic-on-short-input shape, unreachable for a 64-hex pubkey, but the inconsistency is now inside the same feature.

### 5. Say which path collects a record whose setup failed after registration

`release`'s doc hands the pending record to "its caller's own rollback/timeout paths". Those cover a publish failure (`remove_pending_request`), but if the publish succeeded and only the subscribe failed, the record detaches at its 10 s timeout and no watcher will ever exist to purge it — nothing collects it for the life of the process. Not a regression (the unconditional purge never ran in that case either), and it is one small struct per trade key, but now that the purge is conditional the doc is the right place to name the gap.

### 6. Branch is 8 commits behind `main`

Down from 31, and the merge is clean, but please rebase before merge so CI validates what actually ships.

---

## On the three open CodeRabbit comments

Flagging these so nobody "fixes" them:

- **`orders.rs:1672`, gate `mark_live` on a non-empty success list** — already done in `fa553ba` via `subscribe_accepted`, with `a_per_trade_subscription_no_relay_accepts_is_an_error` covering it. Stale (it reviewed `ea54db6`).
- **`subscriptions.rs:278-279`, take the key only after the lock** — also already the code, and `cancelled_mark_live_and_release_still_free_the_key` pins both directions. Stale.
- **`orders.rs:1702-1820`, keep the watcher subscribed until `is_hard_terminal`, including `SettledHoldInvoice`** — that is the pre-existing 30-minute idle policy, untouched by this PR, and re-arm paths (#218/#291) are what restore coverage. Out of scope; changing it here would be a separate behavioral decision with its own contract update.
- **`orders.rs:3993-4005`, restore the filter when `replace_subscription` fails** — a real pre-existing gap, but this PR only moved that body into `subscribe_accepted`. Worth its own issue, not this PR.

Forte11Cuba:
@grunch conficts fixed!

PR #412, opened 2026-09-10
fix: classify the missing trade row instead of writing to nothing

### LINKED ISSUE #394: A daemon message for an order with no trade row has no defined behaviour

Forte11Cuba:
Closes #394
**Commit 1 — classify the absence.** Deliberate wipes (pre-active cancel, stale sweeper) leave a `trade_wiped:<order_id>` tombstone; replayed messages for those orders are dropped whole (no write, no TradeUpdate). Row creation lifts the tombstone (re-takes work). Status arms also skip value-identical rewrites — the startup replay no longer re-writes and re-emits what the row already holds — and the two 38383 sync paths stop writing `amount_sats` when `wire_status_applies` refused the status.

**Commit 2 — rebuild from the DM.** A never-written row (confirmation timeout) is reconstructed from the daemon message itself — ownership proven by decryption, role by the payload's trade pubkeys, never guessed. The late NewOrder confirmation now persists the maker row instead of delegating to the fingerprint.

**Commit 3 — remove the content fingerprint.** With DM-driven recovery in place, Kind 38383 goes back to being the order book: `order_content_key` and the ingest restore block are gone, and `is_mine` on cold start comes from the durable trade-key binding + row, keyed by daemon UUID like every reference client. Rationale and DM-coverage measurement (32/35, exceptions explained) in #394. Known trade-off: no `is_mine` restore once every relay dropped the DM (>30 days), where the fingerprint could also answer wrong (#326).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented replayed messages from restoring trades that were intentionally canceled or removed.
  * Preserved recovery of trades that were never successfully saved.
  * Improved handling when a previously removed order is legitimately taken again.
  * Prevented duplicate status updates and notifications when incoming information does not change an order.
  * Improved synchronization of trade status across order creation, payment, cancellation, and ingestion flows.
  * Improved late-response reconciliation without relying on content fingerprints.

* **Documentation**
  * Clarified order recovery and ownership behavior for late responses and replayed messages.

### DIFF
--- a/rust/src/api/orders.rs
+++ b/rust/src/api/orders.rs
@@ -1435,6 +1435,13 @@ fn trade_row_from_small_order(
 /// that key's trade index, which the daemon echoes from the release. It
 /// becomes a maker trade of this client's, listed and cancellable like the
 /// parent was. Returns whether an order was adopted.
+///
+/// With the content fingerprint gone (#394 step 3) this is also the ONLY
+/// path that restores a still-pending maker order on cold start: a create
+/// sends `trade_index`, mostrod echoes it in the Pending ack, and the
+/// replayed ack of a create whose confirmation timed out is adopted here as
+/// a maker row once no pending record remains to intercept it (review
+/// round 2).
 async fn adopt_range_remainder(
     order_id: &str,
     kind: &mostro_core::message::MessageKind,
@@ -1662,9 +1669,12 @@ async fn persist_late_create_confirmation(
     trade_index: u32,
 ) {
     let Some(mostro_core::message::Payload::Order(order)) = &kind.payload else {
-        log::warn!(
-            "[orders] late create confirmation for order={daemon_id} carries no order — \
-             nothing to persist"
+        crate::api::logging::blog_warn(
+            "orders",
+            format!(
+                "late create confirmation for order={daemon_id} carries no order — \
+                 nothing to persist"
+            ),
         );
         return;
     };
@@ -1672,7 +1682,10 @@ async fn persist_late_create_confirmation(
         Some(mostro_core::order::Kind::Sell) => TradeRole::Seller,
         Some(mostro_core::order::Kind::Buy) => TradeRole::Buyer,
         None => {
-            log::warn!("[orders] late create confirmation for order={daemon_id} names no kind");
+            crate::api::logging::blog_warn(
+                "orders",
+                format!("late create confirmation for order={daemon_id} names no kind"),
+            );
             return;
         }
     };
@@ -1696,7 +1709,10 @@ async fn persist_late_create_confirmation(
         return;
     };
     if let Err(e) = persist_trade_row(db, &trade).await {
-        log::warn!("[orders] late create confirmation not persisted for order={daemon_id}: {e}");
+        crate::api::logging::blog_warn(
+            "orders",
+            format!("late create confirmation not persisted for order={daemon_id}: {e}"),
+        );
         return;
     }
     crate::api::logging::blog_info(
@@ -2457,9 +2473,12 @@ async fn dispatch_mostro_message(
                 } else {
                     // Cold start / reconnect (no record — in-memory state is
                     // empty after a restart), or an uncorrelated event that
-                    // must not consume anything. Recovery is DM-driven: the
-                    // trade row either exists (nothing to do) or is rebuilt
-                    // from a message that proves it (#394).
+                    // must not consume anything. Recovery for a NewOrder is
+                    // NOT the prologue rebuild (it excludes NewOrder): a
+                    // replayed create ack is adopted by
+                    // `adopt_range_remainder` above, which just declined —
+                    // row already there, tombstone, older than the cursor,
+                    // or not a Pending order of this key's (#394).
                     crate::api::logging::blog_info("daemon-msg", format!(
                         "NewOrder: daemon order={daemon_id} with no matching \
                          pending create — leaving state untouched"
@@ -3322,9 +3341,12 @@ enum RowState {
     /// Deleted on purpose: status arms drop the message whole — no cursor
     /// advance, no write, no TradeUpdate.
     Wiped,
-    /// No row and no tombstone: never persisted. Kept on today's path (the
-    /// write warns "matched no row", the update still emits) until
-    /// DM-driven rebuild lands — issue #394 step 2.
+    /// No row and no tombstone covering the message's generation: never
+    /// persisted, or a later take of a wiped order. The dispatch prologue
+    /// rebuilds the row when the message proves one
+    /// (`rebuild_trade_from_dm`, #394 step 2); a message that proves
+    /// nothing falls through to the pre-#394 path — the write warns
+    /// "matched no row", the update still emits.
     NeverWritten,
     /// No store yet, or the store failed to answer: nothing to classify on,
     /// arms behave exactly as before this classification existed.
@@ -3371,12 +3393,18 @@ async fn trade_row_state(order_id: &str, trade_index: u32) -> RowState {
             Ok(Some(value)) if tombstone_covers(&value, trade_index) => RowState::Wiped,
             Ok(Some(_)) | Ok(None) => RowState::NeverWritten,
             Err(e) => {
-                log::warn!("[orders] tombstone lookup failed for order={order_id}: {e}");
+                crate::api::logging::blog_warn(
+                    "orders",
+                    format!("tombstone lookup failed for order={order_id}: {e}"),
+                );
                 RowState::Unknown
             }
         },
         Err(e) => {
-            log::warn!("[orders] trade lookup failed for order={order_id}: {e}");
+            crate::api::logging::blog_warn(
+                "orders",
+                format!("trade lookup failed for order={order_id}: {e}"),
+            );
             RowState::Unknown
         }
     }
@@ -3384,9 +3412,11 @@ async fn trade_row_state(order_id: &str, trade_index: u32) -> RowState {
 
 /// Gate for the status-writing dispatch arms (issue #394). `true` means drop
 /// the message whole. A wiped row is the normal drop on every restart replay,
-/// so it logs at debug; a never-written row falls through to today's behavior
-/// but logs the recovery candidate it is — that line is the field measurement
-/// step 3 of the issue is gated on.
+/// so it logs at debug. A never-written row reaching an arm means the
+/// prologue rebuild already declined — the message proved no trade — so it
+/// falls through to the pre-#394 behavior and logs the recovery candidate it
+/// still is: the line that measured DM coverage before the fingerprint went
+/// (#394 step 3), kept because a gap here is a lost trade.
 fn status_arm_gate(
     row_state: &RowState,
     action: &mostro_core::message::Action,
@@ -3444,7 +3474,10 @@ async fn wipe_trade_row(
     {
         // Classification degrades to pre-#394 behavior for this order:
         // replays write to nothing and emit, as they always did.
-        log::warn!("[orders] wipe tombstone not persisted for order={order_id}: {e}");
+        crate::api::logging::blog_warn(
+            "orders",
+            format!("wipe tombstone not persisted for order={order_id}: {e}"),
+        );
     }
     Ok(())
 }
@@ -3458,9 +3491,13 @@ async fn persist_trade_row(db: &impl Storage, trade: &crate::api::types::TradeIn
     if let Err(e) = db.delete_setting(&key).await {
         // Save anyway: a stale tombstone only mutes replays for this order,
         // and the next (re)creation retries the delete.
-        log::warn!(
-            "[orders] failed to lift wipe tombstone for order={}: {e}",
-            trade.order.id
+        crate::api::logging::blog_warn(
+            "orders",
+            format!(
+                "failed to lift wipe tombstone for order={}: {e} — replays for \
+                 this trade stay muted until a (re)creation retries the lift",
+                trade.order.id
+            ),
         );
     }
     db.save_trade(trade).await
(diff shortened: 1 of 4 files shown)


Catrya (changes requested):
**Changes requested.** The classification is the right shape and most of it is well tested, but the rebuild can resurrect trades the daemon already canceled, and I reproduced that through the real dispatcher.

## Blocking

### 1. The rebuild runs before the status cursor, so an older message resurrects a canceled trade

`rebuild_trade_from_dm` runs before the arms and never consults `status_write_blocked`. `Canceled` is excluded from the rebuild, but its arm still advances the cursor. So on a newest-first replay the cancel lands first, finds no row and rebuilds nothing, and the *older* take reply that follows rebuilds the row anyway. I drove this through `dispatch_mostro_message`, with no row and no tombstone:

| replay (newest first) | this PR | `main` |
|---|---|---|
| `Canceled@3000` → `AddInvoice@2000` | **row at `WaitingBuyerInvoice`**, updates `[Canceled, WaitingBuyerInvoice]` | no row, `[Canceled]` |
| `Canceled@3000` → `PayInvoice@2000` | **row at `WaitingPayment`**, updates `[Canceled, WaitingPayment]` | no row, `[Canceled]` |

Two populations hit this:

- **Every row wiped before this PR.** It has no tombstone, so after the upgrade it classifies as `NeverWritten`. On one real database I checked, 3 wiped orders still have daemon messages inside the 30-day window.
- **A take whose confirmation timed out and which the daemon then canceled on its waiting timeout.** That is the realistic ending of exactly the case step 2 is for.

The resulting row sits in a waiting state forever, because the daemon has nothing more to say about it. `the_newest_first_replay_rebuilds_once_and_blocks_the_tail` only covers the case where the newest message is itself one that rebuilds.

Suggested fix: gate the rebuild on the same cursor check the arms use, so a message older than the order's newest accepted event cannot create a row. The same guard on `adopt_range_remainder` would also close a pre-existing variant (see below).

### 2. `is_mine = false` on rebuilt rows rests on a false premise

The doc comment, the ingest comment and the test all say maker-ness "is not provable from a mid-trade message". It is: the role is proven, the payload carries the order kind, and in Mostro the maker is the seller of a sell order and the buyer of a buy order. The Dart side already relies on that equivalence in `_deriveIsBuyer` (`trade_state_header.dart:229`).

`a_never_written_trade_is_rebuilt_from_the_dm` pins the wrong answer: its row is a Sell order whose seller is our key, which makes us the maker, and it asserts `!is_mine`. User-visible effects on a recovered maker trade:

- "Taken by you" instead of "Created by you";
- `trades_list_item.dart` does not route a maker-seller at `WaitingPayment` to the pay-invoice screen;
- after a republish, the order shows in the book as a stranger's.

One line: `is_mine = matches!((order.kind, role), (Some(Kind::Sell), TradeRole::Seller) | (Some(Kind::Buy), TradeRole::Buyer))`.

### 3. The retake's tombstone lift is not held through its callers

Mutation: change `take_order` back to `db.save_trade(&trade)`, and all **449** tests stay green. `a_retake_after_a_wipe_lifts_the_tombstone` calls `persist_trade_row` directly, so it proves the helper works, not that the retake path uses it. If that call regressed, a re-taken order would silently swallow all of its daemon messages. A cheap guard is enough, for example a static test that no production `save_trade(` remains in `orders.rs` outside the helper.

## Non-blocking

- **The tombstone ignores the trade key.** Wipe on index 1, then a new take of the same order on index 5 whose confirmation times out: its first unconsumed message is classified `Wiped` and dropped, and the row never appears (probe below). The generation gate already reasons by `trade_index`. Storing the wiped index in the tombstone and classifying `Wiped` only for `trade_index <= that index` would close this. Rare, but it suppresses exactly the recovery step 2 exists for.
- **The two Kind 38383 paths stopped writing `amount_sats` when `applies` is false.** #394 doesn't motivate this and no test covers it: restoring the old write shape leaves everything green. It is probably harmless, since the amount still lands whenever the status applies, but please say why and pin it.
- **`adopt_range_remainder` is now load-bearing for cold start.** The create message sends `trade_index`, mostrod echoes it in the ack with status `Pending`, so after a restart the replayed ack of a timed-out create is adopted there as a maker row (`is_mine = true`, measured, same on `main`). With the fingerprint gone, this is the only path that restores a still-pending maker order. It is documented as range-remainder only, though, and its new tombstone check is untested (removing it leaves everything green). Worth a line in its doc and a test.
- **Doc rot inside the PR.**
  - `RowState::NeverWritten` still says "until DM-driven rebuild lands — issue #394 step 2", and step 2 is commit 2 of this PR.
  - `status_arm_gate` refers to "the field measurement step 3 of the issue is gated on", and step 3 is commit 3.
  - The `NewOrder` else-branch says the row "is rebuilt from a message that proves it", but the rebuild excludes `NewOrder`; the path that does that work is `adopt_range_remainder`.
- **The new failure paths log with `log::warn!`, which never reaches anyone in the app.** `install_log_bridge` only runs from `init_app`, which FRB never wires (`init_app` is absent from `frb_generated.rs`), and Dart never calls `installLogBridge`. The sharp one is `persist_trade_row`: a failed tombstone lift silently mutes a live trade. `blog_warn` would make these visible.
- **Pre-existing, not introduced here:** a maker who canceled their own pending order before this PR gets the order resurrected as `Pending` on the next start, because the replayed create ack is adopted after the cancel. `main` does the same. The cursor gate from blocker 1, applied to `adopt_range_remainder`, would close this too.

## Probes

Each probe was a scratch test appended to `orders.rs` that drives `dispatch_mostro_message` directly, then removed:

```
P1 Canceled@3000 → AddInvoice@2000        PR: ROW WaitingBuyerInvoice role=Buyer      main: no row
P2 Canceled@3000 → PayInvoice@2000        PR: ROW WaitingPayment role=Seller          main: no row
P3 Canceled@3000 → create ack@2000        PR: ROW Pending is_mine=true                main: same (pre-existing)
P4 cold-start create ack, no pending rec  PR: ROW Pending is_mine=true (via adopt)    main: same
P5 tombstone(idx 1) → AddInvoice on idx 5 PR: no row, no update
```

Catrya (approved):
**Approved.** All three blockers from round 2 are fixed, and each fix is held by a test: removing it makes that test fail.

## Round 2 findings

| Finding | Fix | Verified |
|---|---|---|
| The rebuild ran before the status cursor, so an older take reply resurrected a canceled trade (P1/P2) | `1415297` gates both the rebuild and `adopt_range_remainder` on `status_write_blocked` | Removing either gate fails `a_rebuild_older_than_an_accepted_cancel_is_refused` / `a_create_ack_older_than_an_accepted_cancel_is_not_adopted` |
| `is_mine = false` on rebuilt rows | `749eb9d` derives it from kind and role | Forcing it to `false` fails `a_never_written_trade_is_rebuilt_from_the_dm` |
| Reverting `take_order` to `save_trade` stayed green | `277d17a` | The mutation now fails `production_code_saves_trades_only_through_
| Tombstone ignored the trade key (P5) | `9191540` | Making `tombstone_covers` always true fails three tests |
| `adopt_range_remainder`'s tombstone checks_the_wipe_tombstone_without_a_cursor` |Removing the check fails it |
| `amount_sats` gate unmotivated and unpinnlf pinned (see below for the d-tag half) |
| Doc rot, silent `log::warn!` | `f196d84` | None of the three stale passages remain; the new failure paths use
`blog_warn` |

I also re-ran the resurrection scenario tha `dispatch_mostro_message` with anewest-first replay:

- **No row, no tombstone, book entry at `Canceled`, no cursor.** The replayed `Canceled` is skipped as terminal and
records no cursor, but the older create acksame terminal check. It is not adopted.
- **An already-resurrected `Pending` row.** On the next start the replayed `Canceled` wipes it and leaves `3000:21`.
The ack behind it no longer brings it back.
- **Oldest-first delivery, then a restart.** It is adopted briefly and wiped by the `Canceled`. After the restart it
stays gone.

So this also closes the zombie maker rows #ding the ones already in a user's DB.

## Non-blocking

1. **The static save-path test cuts the fil.** Any `#[cfg(test)]` helper earlier in
`orders.rs` makes it count too little and fBook::has_wire_note`, around line 549).
Splitting on `"#[cfg(test)]\nmod tests"` woere test-only helpers live.
2. **The d-tag half of the `amount_sats` ga the old write shape there leaves everything
green. Only the book-feed path is covered b_not_sneak_its_amount_into_the_row`. Thatloop is inline in the spawned task, so it is hard to reach here. #419 moves it into `apply_single_order_update`,
where the test can go when the two meet.

## What I checked

- The full branch diff locally. It touches , `pending.rs` and the contract: no buildfiles, scripts or dependency changes.
- Merged with today's `main`, 72 commits behind, and it merges cleanly. `cargo test --locked` (484), `cargo clippy --locked -- -D warnings`, `cargo check --taand `flutter test` all pass.
- CI on `f196d84`: Rust and Flutter green.  when I looked.

Catrya (approved):
tACK

PR #419, opened 2026-09-10
fix(#417): a lost take returns to the ex-taker's book, one trade row per order

### LINKED ISSUE #417: Ex-taker cannot retake an order: it disappears from their order book, and a retake leaves duplicate trade rows
A taker whose take ends before the trade goes active (they cancel it, or the daemon times them out) can never take that order again from the same device. Mostrod republishes the order as `pending` and every other client sees it, but for the ex-taker it **disappears from the order book**: the local entry stays at the status it had (`WaitingBuyerInvoice` after a timeout, `Canceled` after their own cancel), and the book screen only lists `pending` orders (`home_order_providers.dart:257`). There is nothing to tap. Even called directly, `take_order` would refuse it with `OrderAlreadyTaken` (`orders.rs:864`).

Fixing only that would make the retake possible but broken: a retake leaves **two trade rows for the same order**, and every lookup by order id can return the old one. The two problems have to be fixed together.

This also matters for #335. The retake is exactly what #375 (merged) protects: a taker cancel does leave the previous session behind, and once the retake is reachable, `install_session` is what replaces it. Today that code never runs in a normal flow, because this issue blocks the retake first.

## 1. The order disappears from the ex-taker's order book

Measured against a live `mostrod` (v0.18.7) with a scratch probe that drives the real `take_order` / `cancel_order`:

- **Taker cancels (`cancel_order`).** The optimistic update marks the row `Canceled` (`orders.rs:1520`), so the daemon's `Canceled` is skipped (`skip replayed Canceled: already Canceled`), and the republished `38383 pending` is refused:
  ```
  status order=5de3ace7 wire=Pending local=Canceled applies=false src=38383/d-tag
  ```
  `wire_status_applies` (`status.rs:116`) accepts only terminal wire statuses once a non-`Pending` row exists. The row stays `Canceled` and survives restarts, so the ex-taker never sees this order as `pending` again.

- **Daemon-side cancel of a take that never went active** (measured as a taker cancel published without the optimistic update, which is equivalent to a waiting timeout from the client's point of view). Mostrod publishes the `38383 pending` **before** the `Canceled`. When it arrives the row is still `WaitingBuyerInvoice`, so it is refused. The `Canceled` then wipes the row and the session (`Canceled before active — removed trade`, `orders.rs:2276`), but no further `38383` arrives, and the book entry stays `WaitingBuyerInvoice` until restart.
  This contradicts the comment at `orders.rs:2335` (*"its wiped row leaves the book's `pending` as the local status"*): the book holds whatever `take_order` wrote into it, not `pending`.

In both cases the order is `pending` in the daemon's database, yet it vanishes from the ex-taker's order book, because the local entry is no longer `pending` and the book screen filters everything else out (`home_order_providers.dart:257`). This matches what shows up when testing by hand: after losing an order to a timeout, the taker never sees it in the book again, even though it has been republished.

## 2. A retake leaves duplicate trade rows

Forcing the book back to `Pending` in the probe (the daemon accepted the retake with `trade_key_index=2`) leaves the local database with two rows for the same order:

```
5b5de7df  trade_key_index=1  order.id=5de3ace7
a2fb425f  trade_key_index=2  order.id=5de3ace7
```

`take_order` saves each take under a fresh `TradeInfo.id` (`orders.rs:1058`, `:1118`), and `save_trade` is `INSERT OR REPLACE` keyed by that id (`sqlite.rs:239`), so nothing ever replaces the first take's row.

What this breaks. The first bullet was observed; the other three follow from reading the code and were **not** exercised end to end:

- `get_trade_by_order_id` is `LIMIT 1` with no `ORDER BY` (`sqlite.rs:541-548`). In the probe it returned the **first** take's row.
- `current_local_status` / `local_trade_status` (`orders.rs:2814`, `:3010`) read through it. After a taker cancel that old row is `Canceled`, which is hard-terminal. So `status_sync_blocked_by_terminal` (`orders.rs:2828`) would skip every daemon message of the retaken trade as *"already Canceled"*, including the peer reveal, which carries the same guard. The retaken trade would freeze with no chat.
- `update_trade_fields` updates **every** row with that order id (`WHERE json_extract(data, '$.ord- The #381 session rebuild reads the row's `trade_key_index`, so after a restart the retaken trade's chat would be rebuilt from the first take's key. That is #335's symptom again, this time from the durable copy.

## Why together

Fixing 1 alone makes the retake reachable and walks straight into 2. Fixing 2 alone changes nothing a user can see, because the retake stays blocked.
er.id') = ?`, `sqlite.rs:587`).

## Suggested direction (not prescriptive)

- **Taker cancel of a never-active trade:** wipe the row and remove the session, as the daemon `Canceled` arm already does for `pending` / `waiting-*` (`cancellation_wipes_history`), instead of marking the row `Canceled`. The republished `pending` then applies, and there is no stale session or duplicate row left to trip over.
- **Timeout path:** after the `Canceled` arm wipes the row, the book entry needs to return to `pending`, since the `38383` that said so arrived first and was refused.
- **Defensively:** a confirmed take could replace any existing row for the same order id, so a retake never leaves two rows even on paths not covered above.

## Acceptance criteria

- [ ] After a taker cancel of a never-active trade, the order reappears as `pending` in the ex-taker's order book and can be taken again from the same session.
- [ ] Same after a waiting timeout.
- [ ] After a retake there is exactly one trade row for the order, carrying the new `trade_key_index`.
- [ ] The retaken trade receives its daemon messages (status progresses, peer reveal lands, chat works), and its session carries the new index (the #375 path, now reachable by hand).
- [ ] Tests pin both halves, going through the real functions rather than copies of the logic.

## Docs to correct in the same PR

Left over from #375, whose text describes a trigger that does not create a stale session. A take the daemon rejects, or one that hits the 10 s client timeout, returns from `take_order` before any session is installed. What leaves a stale session is an earlier take that **was confirmed** and never cleaned up, which is exactly the taker-cancel path above.

- `contracts/orders.md:124`: *"a prior failed or timed-out attempt left a stale one"* → describe the real source, and add that a rejected or timed-out take installs nothing.
- `contracts/orders.md:495`: the binding is *"written by `take_order` on every attempt"* → *"on every confirmed take"* (it is written at `orders.rs:1100`, after the reply is accepted).
- `session.rs:167` (`install_session` docstring), the comments at `orders.rs:1129-1135`, and the test comment at `orders.rs:7594`: same false trigger.
- The docstring of `retake_replaces_stale_session_trade_key_index` claims the `take_order → install_session` seam *"is covered by the manual regtest run"*. That run never retook an order. Reword it, or better, replace it with the new retake test.
- `orders.rs:2335`: the *"wiped row leaves the book's `pending`"* comment, which should be true once this is fixed.
- `orders.rs:1148`: `log::warn!` → `blog_warn`. `log::` records are discarded in the app because `install_log_bridge` never runs.
- Update the contract for the new cancel / retake behaviour itself.

Catrya:
Closes #417

## Problem

A taker who lost a take before the trade went active (their own cancel, or the daemon's waiting timeout) never saw that order in their book again, although mostrod republishes it as `pending` and every other client sees it. Had they been able to retake it, the retake would have left a second trade row for the same order, and lookups by order id could return the dead one.

Two local mechanisms caused this. None of the reference clients (mobile, mostrix, mostro-cli) has either:

- **`cancel_order` marked the row `Canceled` before the daemon replied.** The daemon's `Canceled` then skipped the row as *"already Canceled"*, so the row and the session outlived the trade, and the terminal row refused the `pending` republish.
- **The book entry of an order of ours carries the local trade status** wherever the wire's is refused. mostrod publishes the `pending` republish **before** the `Canceled`, so the republish arrived while the `waiting-*` row still stood and was refused. Nothing arrives after the `Canceled` to correct it. The book screen lists only `pending` orders, so the order was gone for the ex-taker alone.

### Losing and retaking a take
- **`cancel_order` no longer writes `Canceled` for a trade that never went active** (`pending` / `waiting-*`), maker or taker. Active trades are still marked `Canceled` right away.
- **A never-active trade is wiped by whichever of the daemon's two reports lands first:** the kind-14 `Canceled`, or the Kind 38383 `canceled` (`wipe_on_public_cancel`). mostrod publishes the event and then enqueues the message, and the two reach separate subscriptions. Letting the event write `Canceled` made the outcome depend on arrival order, for a maker's own cancel and for a take whose maker cancelled alike. The event is also the only report of an expired pending order: mostrod publishes `Expired` as `canceled` and sends no message.
- **A wiped take hands its order back to the public book.** `OrderBook` notes the daemon's latest public view of each order the d-tag subscription watches, and the wipe settles the entry from it:
  - the view is `pending` → the entry is restored to it;
  - any other view, or none while the entry holds a local status → the entry is dropped, so the next 38383 applies as is;
  - already `pending` → left alone.
  A note is forgotten once nothing can read it (hard-terminal view, a trade ended without a wipe, any wipe), and the notes tolerate a poisoned lock.
- **Every wipe goes through `wipe_trade_row`** (#394) and leaves the tombstone. That covers the daemon's `Canceled`, the public `canceled` and the stale sweep. So on the next start neither the DM rebuild nor `adopt_range_remainder` brings a canceled order back as a `Pending` row.
- **A confirmed take is its order's only row.** `take_order` deletes every earlier row for the order and saves through `persist_trade_row`, which lifts the tombstone so the retake's own messages are not dropped. On web, `delete_trade_by_order_id` now removes every matching document, as SQLite does.
- **One d-tag task per order.** A retake replaces the earlier take's `subscribe_single_order` task. The old task stops at its next wake without touching the subscription, and the new one re-opens it with a fresh idle window and is the only one that may drop it. Before, both tasks applied every event, and whichever exited first unsubscribed the REQ the other relied on.

### Only the daemon, and only the active node, moves local state
- The d-tag loop checks the event's author, and the refetch drops other authors. A `canceled` now deletes a row, and a d-tag is public.
- A d-tag task stops as soon as an event of its order arrives while its node is no longer the active one, as `dispatch_mostro_message` already rejects any other sender.

### The trade screen
- **It leaves for home once the trade is no longer the user's** (no trade row, and not the maker), with "You're no longer part of this trade". Otherwise a lost take's order, back in the book as `pending`, rendered as the user's own published order, cancel button included. It waits for settled reads of both the trades list and the order book.
- **Cancelling a never-active trade goes home right away**, as the invoice screens do. The decision uses the live status after confirmation, because the seller's payment can land while the dialog is open.
- **The cancel dialog says what the cancel does:** immediate before `active`, cooperative from `active` on, and both outcomes for `in-progress` (#203). The copy follows the live status while the dialog is open. The invoice screens always say it is immediate.
- The Dart copy of `cancellation_wipes_history` is pinned to the Rust predicate by a test that reads the Dart source.
- Both cancel call sites go through `cancelOrderActionProvider`.

### Docs
`contracts/orders.md` describes the cancel, the retake, the tombstone, the single-task rule, and what the user sees between a cancel and the daemon's answer. Its `cancel_order` section no longer describes ownership checks and error codes that do not exist. It also corrects what #375 left behind: a failed or timed-out take leaves no session behind, because the stale one comes from an earlier confirmed take whose `Canceled` never arrived.

## User-visible behaviour

- After losing a take, by cancelling it or by timeout, the order **reappears** in the ex-taker's book within a second of the daemon's `Canceled` and can be taken again.
- A cancelled trade that never went active **leaves My Trades**, for a maker and for a taker, whichever of the daemon's reports lands first. That matches v1, and the timeout path already did it. An expired pending order leaves too, once its public `canceled` is seen.
- For the second or so before the daemon answers, the trade keeps its previous status instead of showing "Canceled". The screen the user cancelled from goes home at once.
- A cancelled order no longer comes back as `Pending` in My Trades after a restart.
- The cancel dialog no longer promises a cooperative request for a trade that hasn't started. The trade screen closes with "You're no longer part of this trade" instead of showing someone else's order as yours.

## Tests

Every test goes through the real functions (`dispatch_mostro_message`, `apply_single_order_update`, `ingest_order_event_with`, `take_order`'s persistence, the screens), and each was **mutation-checked**: it fails when the piece it pins is removed.

| Test | Fails when |
|---|---|
| `cancel_of_a_never_active_take_is_left_for_the_daemons_canceled` | the optimistic `Canceled` write is restored |
| `a_makers_never_active_cancel_is_wiped_whichever_signal_lands_first` | the public `canceled` stops wiping (also covers an expiry, which gets no message) |
| `a_take_whose_maker_cancelled_is_wiped_by_the_public_canceled` | the d-tag path stops wiping |
| `a_public_canceled_keeps_an_active_trade_as_history` | the wipe ignores how far the trade got |
| `a_lost_take_returns_to_the_book_when_the_republish_came_first` / `…_canceled_came_first` | no settle after the wipe, or the stale entry is kept |
| `a_republish_seen_only_by_the_book_feed_survives_the_wipe` | the book feed stops refreshing the note |
| `a_cancelled_take_settles_before_its_note_is_forgotten` / `a_finished_take_leaves_no_wire_note_behind` / `a_wiped_makers_order_leaves_no_wire_note_behind` / `a_poisoned_note_lock_still_notes_and_forgets` | a note is forgotten too early, never, or not under a poisoned lock |
| `a_public_canceled_wipe_stops_the_replayed_create_ack` | a wipe stops leaving the tombstone |
| `a_confirmed_take_replaces_the_orders_earlier_row` | no delete before save |
| `a_retake_replaces_the_earlier_single_order_task` / `subscribe_single_order_claims_before_it_spawns` | the registry stops replacing, or a superseded task can drop the subscription |
| `the_d_tag_task_stops_once_its_node_is_no_longer_active` / `the_d_tag_task_applies_only_its_active_nodes_events` | the node, author or order-id check goes |
| `a_refused_d_tag_status_does_not_sneak_its_amount_into_the_row` | the d-tag half of #394's amount gate goes |
| `the_trade_screen_copy_of_cancellation_wipes_history_matches` | the Dart set and the Rust predicate diverge |
| Trade screen widget tests (leave on a lost take, maker stays, not judged before the trades list and the book load, status turning active while the dialog is open, dialog copy per status) | each rule is reverted |
| `invoice_cancel_dialog_test`, `my_order_screen_test` (cancel through the provider) | the copy or the seam is reverted |

A two-phase `#[ignore]` E2E against a live daemon, `retake_e2e_maker_creates_order` followed by `retake_e2e_taker_cancels_and_retakes`, exercises the whole flow: take → cancel → order back as `pending`, row wiped and session removed → retake. Before the retake it plants the first take's row and session again, so the live `take_order` also drives the one-row rule and `install_session`'s replacement of a stale session (#335).

## Test plan

- [x] `cargo test --locked`: 505 passed, 27 ignored
- [x] `cargo clippy --locked -- -D warnings`: clean
- [x] `cargo check --locked --target wasm32-unknown-unknown`: clean
- [x] `flutter analyze` and `flutter test`
- [x] E2E against a local `mostrod` (v0.18.7) over `relay.mostro.network`, and the timeout path with `expiration_seconds = 60` (on the first version of this PR)
- [x] Manual runs in the Linux app during the review:
  - maker cancel with the public `canceled` first, row wiped;
  - take → cancel → back in the book → retake → cancel, twice;
  - maker cancel from `waiting-payment`;
  - first start of #412 on a real database: no rebuild and no resurrection
- [x] Manual retest of the final branch: retake (single d-tag task, tombstone lifted), restart after a maker cancel, trade-screen exits and copy

## Related

- **#412** (merged): its tombstone is what every wipe here now leaves, and `persist_confirmed_take` saves through its `persist_trade_row`. Together they close the replay resurrection of wiped maker rows.
- **#407**: its per-trade subscription registry leaves `subscribe_single_order` out of scope. The single-order registry here covers the second caller this PR creates.
- **#303**: it emits `Canceled` right after `cancel_order`'s optimistic write, which this PR removes for never-active trades. Whichever lands second has to reconcile the two.
- **#395**: one row per order id is a step toward a single row identity, but the primary key is unchanged.

## Out of scope, follow-ups

- **`cancel_order` is fire-and-forget:** a `CantDo` is `Canceled`/`CantDo` the way take and add-invoice do since #170.
- **The stale sweep ignores maker `Pending` rows**, so one whose public `canceled` was missed (the app was offline for more than 48 h) stays listed.
- **The book keeps the last write, not the newest event** (`apply_upsert` has no timestamp check, unlike every reference client). Local state used to mask it.
- **The peer-reveal path does not compare `trade_key_index`** (#417). It is reachable only through a Cashu take accepted after the client timeout.
- **`contracts/orders.md` still lists `request_cooperative_cancel` / `accept_cooperative_cancel`**, which do not exist as functions.
- **Relay subscription limits** (`relay.mostro.network`: *"Number of subscriptions exceeds limit"*; `nos.lol`: *"too many concurrent REQs"*) show up on every take. They predate this PR.

## Summary by CodeRabbit

* **New Features**
  * Added clearer cancellation messaging based on trade status and counterpart approval requirements.
  * Improved order and trade screens with updated status displays, cancellation controls, feedback, and navigation.
* **Bug Fixes**
  * Improved handling of canceled, expired, and never-activated orders.
  * Restored orders to the public book when a take is lost or wiped.
  * Prevented duplicate or stale trade records from affecting order status.
  * Improved handling of repeated, unauthorized, and outdated order updates.
* **Localization**
  * Added cancellation messaging in English, German, Spanish, French, and Italian.


Catrya:
@coderabbitai review

rust/src/api/orders.rs
```diff
@@ -1170,6 +1251,41 @@ pub async fn take_order(
...
+        crate::api::logging::blog_warn(
+            "orders",
+            format!(
+                "take_order: trade not persisted for order={}: {e}",
+                crate::api::logging::short_id(&trade.order.id),
+            ),
+        );
+    }
+}
+
```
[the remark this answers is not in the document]
Catrya: Declining this one. The failure it describes — `save_trade` failing leaves the confirmed take without a row while the binding and session proceed — predates this change: `take_order` on `main` already only logged a failed save and carried on. The only difference here is that the *earlier* row is gone too, and that row belongs to a take that is over by construction: the daemon only confirms a new take of an order once the previous one ended. Rolling back to it would reinstate exactly what this commit removes — `get_trade_by_order_id` returning the dead take, whose terminal status makes the guards skip the new trade's messages and whose `trade_key_index` rebuilds the chat with a superseded key. In the normal flow the delete is a no-op anyway, since the daemon's `Canceled` already wiped that row. A trade with no row at all is the case #394 / #412 address.

rust/src/api/orders.rs
```diff
@@ -1551,6 +1645,101 @@ pub async fn cancel_order(order_id: String) -> Result<()> {
...
+                "orders",
+                format!(
+                    "cancel: trade lookup failed for order={}: {e}",
+                    crate::api::logging::short_id(order_id),
+                ),
+            );
+            return;
+        }
+    };
+    if cancellation_wipes_history(&status) {
```
grunch, line 2010: _⚠️ Potential issue_ | _🟠 Major_

**A maker's own pending-order cancel is now ordering-dependent.**

`apply_local_cancel` defers the `Canceled` write for every never-active row, not only for takes. For a maker's own `Pending` order that makes the My Trades outcome depend on which of two independent subscriptions the client processes first:

- Kind 38383 `canceled` first (book feed) → `ingest_order_event_with` syncs the row to `Canceled`, then the kind-14 `Canceled` hits the "history kept" arm → the order stays in My Trades as Canceled (pre-PR behaviour).
- Kind-14 `Canceled` first → the row is still `Pending`, `cancellation_wipes_history` is true, `wipe_never_active_trade(oid, false)` deletes the row → the order vanishes from My Trades.

mostrod does publish the 38383 before enqueueing the DM (`cancel.rs`: `update_order_event(Status::Canceled)` then `enqueue_order_msg(Action::Canceled)`), but delivery goes through different relays/subscriptions and is handled by different tasks here, so "arrives first" is not guaranteed. Before this PR the optimistic write made the maker path deterministic. The PR description acknowledges relying on that ordering ("unaffected in practice").

Since the whole fix is about takes, keep the optimistic write for maker rows and defer only for takes:

```diff
-    let status = match db.get_trade_by_order_id(order_id).await {
-        Ok(Some(trade)) => trade.order.status,
+    let (status, is_take) = match db.get_trade_by_order_id(order_id).await {
+        Ok(Some(trade)) => (trade.order.status, !trade.order.is_mine),
         // ...
     };
-    if cancellation_wipes_history(&status) {
+    if is_take && cancellation_wipes_history(&status) {
```

If that is adopted, the comment in `lib/features/order/screens/my_order_screen.dart` (maker screen) goes back to its previous wording, and the `cancel_order` section of `contracts/orders.md` should say the deferral applies to takes only. A test pinning "maker's own pending cancel keeps a Canceled history row regardless of which message lands first" would close this.

rust/src/api/orders.rs
```diff
@@ -470,6 +475,81 @@ impl OrderBook {
...
+    ///
+    /// While a trade of ours lives, its book entry carries the local trade
+    /// status wherever the wire's is refused (`wire_status_applies`) — the
+    /// public `pending`/`in-progress` buckets are not a trade's status. That
+    /// stops being right the moment the trade is wiped: the daemon published
+    /// its `pending` republish *before* telling us the take was canceled, so
+    /// nothing arrives afterwards to correct the entry, and the ex-taker never
+    /// sees the order again. The wipe restores from this note
+    /// ([`Self::settle_after_lost_take`]).
+    pub(crate) fn note_wire_order(&self, order: &OrderInfo) {
```
grunch, line 515: _🧹 Nitpick_ | _🟡 Minor_

**`wire_orders` grows for the life of the process.**

`note_wire_order` inserts on every d-tag update of every order we created or took, in any status, but the only removal is `settle_after_lost_take`, which runs for wiped takes alone. Maker orders, takes that go active, and trades that end in `success`/`canceled` leave their `OrderInfo` clone in the map forever, and the map is never pruned by the stale sweep either. Bounded by the user's activity, so not a leak in the alarming sense, but it is unbounded by design and each entry is a full `OrderInfo`.

Cheapest fix: drop the note when a hard-terminal public status arrives (`is_hard_terminal(&order.status)` in `apply_single_order_update` / `refresh_wire_order`), and in the non-wipe branches of the `Canceled`/`Success` arms.

Also, both `lock()` calls use `if let Ok(...)`: a poisoned mutex silently disables the restore for the rest of the session, and nothing else would tell you. No invariant spans the guard, so `lock().unwrap_or_else(PoisonError::into_inner)` is the safer default.

rust/src/api/orders.rs
```diff
@@ -8598,4 +9151,149 @@ mod restore_e2e_tests {
...
+
+        let first = take_order(order_id.clone(), TradeRole::Buyer, None)
+            .await
+            .expect("first take");
+        println!("[test] first take idx={}", first.trade_key_index);
+        cancel_order(order_id.clone()).await.expect("cancel the first take");
+
+        assert!(
+            wait_for_book_status(&order_id, OrderStatus::Pending, 40).await,
+            "a lost take's order must come back to the ex-taker's book"
```
grunch: _⚠️ Potential issue_ | _🟡 Minor_

**Flaky assertion order in the live E2E.**

`wait_for_book_status(Pending)` can be satisfied before the daemon's `Canceled` is processed: the book feed writes a `pending` republish straight into the entry for an order with `ours == false` (`apply_ingested_order` → `upsert_order_*`, no `wire_status_applies` gate on `pending`), and mostrod publishes that republish before enqueueing the DM. When that happens the next two assertions (`get_trade_by_order_id(..).is_none()` and `get_session(..).is_none()`) run before `wipe_never_active_trade` and fail intermittently.

Poll for the row (or the session) to disappear with the same 40 s budget instead of asserting it right after the book check, e.g. a `wait_for_row_gone(&order_id, 40)` helper next to `wait_for_book_status`.

grunch (changes requested):
## Review summary

Solid PR: the root-cause analysis is right, the split into one commit per concern makes it reviewable, and the tests go through the real dispatch/ingest paths rather than helpers. CI is green (Rust, Flutter, wasm smoke). I verified the claims against `mostrod`'s `cancel.rs` and against `apply_ingested_order` / `ingest_order_event_with` on this branch.

**Requesting changes** for one Major finding; the rest are Minor/nitpicks.

### Findings

| Severity | File | Finding |
|---|---|---|
| 🟠 Major | `rust/src/api/orders.rs` (`apply_local_cancel`) | A maker's own pending-order cancel now depends on whether the 38383 `canceled` or the kind-14 `Canceled` is processed first: Canceled history row vs. row wiped. Pre-PR this was deterministic. Defer the optimistic write for takes only. |
| 🟡 Minor | `rust/src/api/orders.rs` (`OrderBook::note_wire_order`) | `wire_orders` is only ever pruned on a lost-take wipe; every other order we created/took stays in it for the process lifetime. Poisoned-lock branch silently disables the feature. |
| 🟡 Minor | `rust/src/api/orders.rs` (`retake_e2e_taker_cancels_and_retakes`) | The row/session assertions after `cancel_order` can run before the daemon's `Canceled` is processed, because the book feed applies the `pending` republish directly. Poll for the wipe instead. |

### Nitpicks (no inline comment)

- **Existing duplicate rows are not migrated.** `persist_confirmed_take` enforces one row per order only on the next confirmed take. DBs that already hold two rows for an order (the exact situation the PR fixes) keep them, and `get_trade_by_order_id`'s unordered `LIMIT 1` keeps picking arbitrarily until that retake. A one-shot cleanup in the stale sweep (keep the row with the highest `trade_key_index`) would close the gap.
- **Two feeds disagree on `pending` for a taken order** (pre-existing, not introduced here): the d-tag path refuses it and keeps the local status, the book feed writes it straight into the entry. Last write wins, as the "Out of scope" note says. Worth the issue you mention, since `settle_after_lost_take` now reads that entry.

### Verified OK

- `refresh_wire_order` runs before `info.status` is overwritten with the local status in `ingest_order_event_with`, so the note carries the wire's view.
- `is_mine` is reliably `false` on a take's row (`parse_order_event` hardcodes it; fingerprint restore only recovers maker orders), so `!local.is_mine` is a sound "was take" signal in both wipe call sites.
- The sweep's `Wipe` decision requires the entry/wire to say `pending` or a terminal status, so `settle_after_lost_take` on that path lands in the "already public pending" or "dropped (no-op)" arms. No regression there.
- SQLite `delete_trade_by_order_id` was already a plain `DELETE … WHERE`; the IndexedDB change brings it in line.
- Contract doc updates match the code, including the corrected `cancel_order` errors.

Catrya:
@coderabbitai review

rust/src/api/orders.rs
```diff
@@ -3321,7 +3677,14 @@ async fn subscribe_single_order(order_id: &str) {
...
             match timeout(remaining, rx.next()).await {
                 Ok(Some(ClientNotification::Event { event, .. })) => {
-                    if let Some(mut order) =
+                    // The notification stream carries every subscription's
+                    // events, and a d-tag is public: only the daemon may move
+                    // a trade of ours — a `canceled` now wipes a never-active
+                    // one (`wipe_on_public_cancel`).
+                    if event.pubkey != mostro_pubkey {
+                        continue;
+                    }
```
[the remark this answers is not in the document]
Catrya: Valid in substance, not as a security issue: the event must be signed by the previous node, which is the authority on its own order, and before this PR the task accepted any author at all. What is inconsistent is that this task is the only path still acting on a non-active node after a switch. dispatch_mostro_message rejects any sender that is not the active node, and the book loop does the same. The task now stops as soon as it sees an event of its order while the node has changed, so it can no longer wipe rows or write into the new node's book. Recreating it for the new node doesn't apply, since the order doesn't exist there..

Catrya:
@coderabbitai review

Catrya:
@coderabbitai review

rust/src/api/orders.rs
```diff
@@ -3296,6 +3595,121 @@ async fn apply_peer_reveal(
...
+            ),
+        );
+        return SingleOrderEvent::NodeChanged;
+    }
+    log::info!(
+        "[orders] d-tag update: order={} status={:?}",
+        order_id,
+        order.status
+    );
+    apply_single_order_update(order).await;
```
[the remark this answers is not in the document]
Catrya: Not reachable, so I'm leaving this one as is.

Every public status that wipes a trade is absorbing in mostrod. `wipe_on_public_cancel` only acts on `canceled`, `expired` and `canceled-by-admin`, and mostrod publishes all three as `canceled` (`nip33.rs`, `create_status_tags`). Nothing is published for an order after that, and it can't be taken again:

- A take requires `Pending` (`take_buy.rs:43`, `take_sell.rs:81`). A canceled order answers `CantDo`.
- The only transitions back to `Pending` start from waiting states: the taker's own cancel (`cancel.rs`, `cancel_order_by_taker_inner`), a taker-responsible waiting timeout (scheduler), and a dropped `WaitingTakerBond` (`bond/flow.rs`). None starts from `Canceled`.

So a `canceled` event can't be older than a confirmed retake of the same order. If the order had been canceled, the retake would have been refused. If the retake was confirmed, any later `canceled` is newer and legitimately ends it: the maker cancelled, or the maker's waiting window ran out.

The ordering gap that does exist is a late `pending` / `in-progress` event, because the book keeps the last write rather than the newest event. That predates this PR and is listed under Out of scope. Neither status triggers a wipe. A per-order public-event cursor would change book ingestion for every order, which I'd rather track separately than add here without a reachable case.

rust/src/api/orders.rs
```diff
@@ -1138,23 +1238,21 @@ pub async fn take_order(
...
-        if let Err(e) = db.save_trade(&trade).await {
-            log::warn!("[orders] failed to persist trade: {e}");
-        }
-    }
+    persist_confirmed_take(&trade).await;
     // Subscribe to d-tag K38383 updates for this specific order so we still
     // see the public buckets the daemon does publish (in-progress once taken,
     // success / canceled at the end); the fine-grained states only ever arrive
     // as daemon messages.
     subscribe_single_order(&order_id).await;
```
grunch, line 1239: **HIGH — a retake leaves two live d-tag tasks fighting over one subscription id.**

This PR makes retaking an order reachable for the first time. When it happens, `take_order` calls `subscribe_single_order(&order_id)` again while the task from the first take is still running: that task only breaks on a 30-minute idle timeout, a shutdown, or a node switch — a wipe does not stop it.

Both tasks use the same stable id, `mostro-order-<order_id>`. Two consequences, and they come from this file's own findings:

1. The second `client.subscribe(...).with_id(sub_id)` is refused because the id already exists (`a_node_switch_replaces_the_live_subscriptions` documents exactly this nostr-sdk 0.45 behaviour, "reporting it per relay, not as an error"). The new task then rides the *old* task's REQ. Same filter, so it works — by accident.
2. Every event is now handled twice: two `update_trade_fields`, two `note_wire_order`, two `upsert_order` per 38383 update.
3. Whichever task breaks first runs the `client.unsubscribe(&sub_id)` after the loop and drops the REQ **the survivor depends on**. The surviving task then receives nothing until its own idle timeout retires it.

Suggested fix: keep a registry of live single-order tasks keyed by order id (a cancellation token / `AtomicBool` the old task checks, or a `HashSet` guard so the second spawn is a no-op). A no-op second spawn is the smaller change and is correct here: the filter is identical, so the first task's subscription already covers the retake.

Worth a regression test in the same style as the others: two `subscribe_single_order` calls for one order, then assert only one task is live.

rust/src/api/orders.rs
```diff
@@ -1194,6 +1295,41 @@ pub async fn take_order(
...
+/// (`get_trade_by_order_id` is `LIMIT 1`, unordered), and the dead one's
+/// status feeds the guards that gate the new trade's daemon messages while its
+/// `trade_key_index` is what the chat-session rebuild derives keys from (the
+/// durable twin of #335). One row per order id, the way every reference client
+/// keys its trades.
+async fn persist_confirmed_take(trade: &crate::api::types::TradeInfo) {
+    let Some(db) = crate::db::app_db::db() else {
+        return;
+    };
+    if let Err(e) = db.delete_trade_by_order_id(&trade.order.id).await {
```
grunch, line 1319: **MEDIUM — the delete is not scoped to the rows the invariant is about.**

`delete_trade_by_order_id` removes *every* row for the order, but the justification in the doc comment is only about never-active leftovers ("its `Canceled` never reached this client, or it predates the wipe"). If a terminal row for the same order id can exist — an earlier take that ended `Success`, `CooperativelyCanceled`, or in a dispute — this silently erases it from My Trades, and the user has no way to tell.

Two ways out, either is fine:

- Scope the delete to rows `cancellation_wipes_history` accepts, so history is never collateral; or
- State in the doc comment why a terminal row cannot be here (mostrod does not republish a `success`/`cooperatively-canceled` order, so it can never be retaken), which makes the broad delete provably safe.

Related, smaller: when the delete fails you log a warning and save anyway, which is the right call for not losing the new trade, but it leaves exactly the two-row state the function exists to prevent — and `get_trade_by_order_id` is `LIMIT 1`, unordered. Worth saying so in the warning text so a log reader knows what state the app is in.

grunch (commented):
## Review — strict pass

Very strong PR: the failure is explained from the daemon's ordering outwards, each commit is one concern, the tests go through the real dispatch paths and are mutation-checked, and the contract doc is corrected rather than appended to. The `wire_orders` note is the right shape for the problem — the entry has to be settled from something, and the wire view is the only honest source.

What follows is what I would block on, what I would fix before merge, and what is worth an issue instead.

| # | Severity | Where | Issue |
|---|---|---|---|
| 1 | **HIGH** | `rust/src/api/orders.rs:1246` | A retake spawns a second `subscribe_single_order` task on the same subscription id while the first is still alive |
| 2 | MEDIUM | `rust/src/api/orders.rs:1313` | `persist_confirmed_take` deletes *every* earlier row for the order, terminal history included |
| 3 | MEDIUM | `lib/features/trades/screens/trade_detail_screen.dart:493` | `orderNoLongerActive` says the opposite of what happened to a lost take |
| 4 | MEDIUM | `lib/features/trades/screens/trade_detail_screen.dart:160` | `_cancelEndsTrade` duplicates Rust's `cancellation_wipes_history` with nothing pinning the two together |
| 5 | LOW | `rust/src/api/orders.rs:1785` | A maker wipe leaves its `wire_orders` note behind for the life of the process |
| 6 | LOW | `rust/src/api/orders.rs:567` | `settle_after_lost_take` logs "book entry dropped" when there was no entry |
| 7 | NIT | `lib/features/order/screens/my_order_screen.dart:57` | `cancelOrderActionProvider` is introduced but this call site still bypasses it |
| 8 | NIT | `rust/src/db/indexeddb.rs:517` | The delete now scans every trade document, on every confirmed take |

Nothing here disputes the design. #1 is the only one I would hold the merge for, and it is pre-existing code that *this* PR makes reachable for the first time — which is exactly the kind of thing a PR that unlocks a path owns.

### Verified while reviewing

- `wipe_on_public_cancel` correctly excludes `CooperativelyCanceled`: a never-active trade cannot reach it.
- The `refresh_wire_order` / `wipe_on_public_cancel` / `forget_wire_order` ordering inside `ingest_order_event_with` is what the doc comments claim, including the "the wipe settles from this very view" case.
- `handle_single_order_event` reads `active_mostro_pubkey()` only for an event of its own order, and the `NodeChanged` break still reaches the `unsubscribe` after the loop.
- The `refetch_active_node_orders` author filter closes a real hole now that a public `canceled` can wipe a row.
- `std::sync::MutexGuard` is never held across an `await`: every `wire_notes()` guard is a statement temporary.
- The five `.arb` files all carry both new keys.

### Out of scope, agreed

Your own "out of scope" list is accurate. `apply_upsert` keeping the last write rather than the newest event is the one I would open an issue for now that local state no longer masks it.

rust/src/api/orders.rs
```diff
@@ -1575,6 +1689,173 @@ pub async fn cancel_order(order_id: String) -> Result<()> {
...
+/// and the entry still describe the same trade.
+async fn wipe_never_active_trade(order_id: &str, was_take: bool) -> Result<()> {
+    if let Some(db) = crate::db::app_db::db() {
+        db.delete_trade_by_order_id(order_id).await?;
+    }
+    crate::mostro::session::session_manager()
+        .remove_session(order_id)
+        .await;
+    if was_take {
+        order_book().settle_after_lost_take(order_id).await;
```
grunch, line 2076: **LOW — the maker path leaves a `wire_orders` note with no reader and no owner.**

`settle_after_lost_take` is what consumes a note, and it only runs for `was_take`. The `ingest_order_event_with` path forgets the note separately because the event is hard-terminal, so it is covered — but `run_stale_sweep_once` reaches `wipe_never_active_trade(&oid, !trade.order.is_mine)` with `was_take == false` and nothing forgets the note there.

It is bounded by the number of maker orders swept in one process, so it is small. But `wire_orders` has no eviction of any kind, so "small and permanent" is worth closing now:

```rust
if was_take {
    order_book().settle_after_lost_take(order_id).await;
} else {
    order_book().forget_wire_order(order_id);
}
```

That also makes the invariant statable in one line: after a wipe, the order has no note.

rust/src/api/orders.rs
```diff
@@ -494,6 +500,100 @@ impl OrderBook {
...
+    /// * The latest public view is `pending` — the daemon republished the
+    ///   order: the entry becomes that view, and the ex-taker can see and take
+    ///   it again.
+    /// * Anything else, or no view at all while the entry still carries a
+    ///   non-`pending` local status: the entry is dropped, so the next Kind
+    ///   38383 event lands on nothing local and applies as is. That covers a
+    ///   `Canceled` that overtook the republish on the way here.
+    /// * No view, entry already `pending`: already public, left alone.
+    pub(crate) async fn settle_after_lost_take(&self, order_id: &str) {
+        let noted = self.wire_notes().remove(order_id);
```
grunch, line 568: **LOW — the log claims an action in the case where nothing happened.**

When there is no note *and* no book entry, this falls to the last arm and logs `book entry dropped (latest public view None)` after a `remove_order` that removed nothing. During a manual run — which is what commit 5 added this logging for — that reads as "the entry was there and I dropped it".

Splitting the `(None, None)` case out keeps the log honest about which of the three outcomes actually occurred.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -394,14 +464,36 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
+    // neither an absent row nor a stranger's order (a cold start can resolve
+    // the trades before the book's first emission).
+    if (!_leaving &&
+        ref.watch(orderBookProvider).hasValue &&
+        !tradeAsync.isLoading &&
+        tradeAsync.hasValue &&
+        trade == null &&
+        order?.isMine != true) {
+      WidgetsBinding.instance.addPostFrameCallback(
+        (_) => _leave(l10n.orderNoLongerActive),
```
grunch: **MEDIUM — this message tells the user the opposite of what just happened.**

`orderNoLongerActive` is *"This order is no longer active"*. In the case this PR creates — a take lost before going active — the order is emphatically still active: the daemon republished it as `pending`, it is back in the book, and the whole point is that the user can take it again. The screen sends them home with a message that says not to bother.

Its `@description` in `app_en.arb` also still reads *"...while the user is on the pay invoice screen"*, so reusing it here leaves the ARB metadata describing a place the string is no longer only used in.

Suggest a dedicated key for this exit, along the lines of "This trade ended — the order is back in the book", with the five translations. It is also the one piece of user-facing copy in this PR that the tests cannot catch being wrong.

lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -138,14 +138,76 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
...
+    ).showSnackBar(SnackBar(content: Text(message)));
+    context.go(AppRoute.home);
+  }
+
+  /// Whether a cancel in [status] ends the trade outright. Mirrors Rust's
+  /// `cancellation_wipes_history`: before `active` mostrod cancels at once —
+  /// a take hands the order back to the book, a maker's order dies — and the
+  /// trade row is wiped. From `active` on it is a cooperative request, and
+  /// `inProgress` may be either.
+  static bool _cancelEndsTrade(TradeStatus status) => const {
```
grunch, line 163: **MEDIUM — a cross-language invariant with nothing holding the two halves together.**

`_cancelEndsTrade` is a hand-copy of Rust's `cancellation_wipes_history` (`rust/src/mostro/status.rs:137`), and the doc comment says so. The comment is the only link: add a status to the Rust predicate and this set silently stops matching, at which point the dialog tells the user their cancel is immediate when it is a cooperative request, or the screen stays open on a trade that was wiped. Both are user-visible and neither fails a test.

`CLAUDE.md`'s golden rule points the same way — this is protocol logic, not UI state.

Options, cheapest first:

- Export the predicate over the bridge (`rust/src/api/`) and have the UI ask, which removes the duplicate outright;
- or, if a bridge call per dialog is unwanted, add a test that pins the Dart set against the statuses the Rust side treats as never-active, so a divergence fails CI rather than a trade.

lib/features/order/screens/my_order_screen.dart
```diff
@@ -55,7 +55,10 @@ class _MyOrderScreenState extends ConsumerState<MyOrderScreen> {
     setState(() => _cancelling = true);
     try {
       await orders_api.cancelOrder(orderId: widget.orderId);
```
grunch: **NIT — one of the two cancel call sites still bypasses the new provider.**

`cancelOrderActionProvider` was added in this PR so the trade screen's cancel can be overridden in a widget test. This call site keeps going straight to `orders_api.cancelOrder`, so there are now two paths to the same action and only one of them is injectable — which is how the *other* one ends up untested later.

Routing this through the provider too is a two-line change and makes the provider the single seam.

rust/src/db/indexeddb.rs
```diff
@@ -508,15 +508,19 @@ impl Storage for IndexedDbStorage {
...
+        // SQLite's `DELETE … WHERE` does: `take_order` relies on this to leave
+        // one row per order after a retake. Messages stay untouched.
         // Serialised with the patches so a delete never races one.
         let _section = self.exclusive(TRADES_LOCK).await;
-        let Some(doc) = self.trade_document_by_order_id(order_id).await? else {
-            return Ok(());
-        };
-        if let Some(id) = doc.get("id").and_then(serde_json::Value::as_str) {
-            self.delete_key(TRADES_STORE, id).await?;
+        for doc in self.trade_documents().await? {
```
grunch, line 517: **NIT — correct, but it now reads the whole store on a hot path.**

Matching SQLite's `DELETE … WHERE` semantics is the right fix. Note what it costs on web: `trade_documents()` materialises every trade document, and `persist_confirmed_take` calls this on **every** confirmed take, not just retakes — so each take is now O(all trades) in deserialisation.

Fine at today's volumes, and web's trade store is a stub anyway (#233). Worth a comment saying the scan is deliberate, so nobody "optimises" it back to the first-match version this commit just fixed.

rust/src/api/orders.rs
```diff
@@ -1575,6 +1689,173 @@ pub async fn cancel_order(order_id: String) -> Result<()> {
...
+                "orders",
+                format!(
+                    "cancel: trade lookup failed for order={}: {e}",
+                    crate::api::logging::short_id(order_id),
+                ),
+            );
+            return;
+        }
+    };
+    if cancellation_wipes_history(&status) {
```
grunch, line 2010: **Question, not a blocker — the in-flight cancel is now invisible.**

Dropping the optimistic `Canceled` is right, and the PR body is honest that for a second or so the trade keeps its previous status. The case I would like stated in `contracts/orders.md` is the one where the daemon's answer never arrives: relay down, app backgrounded, daemon refusing the cancel. The row then sits at `waiting-*` with nothing recording that the user asked to cancel, until the stale sweep gets to it — and the user's own reading of the screen is that their cancel did not happen.

The old behaviour lied in the other direction (a refused cancel looked canceled), so this is still an improvement. But "what the user sees between the request and the daemon's answer, and how long that can last" is exactly what the contract section you rewrote should pin down.

grunch (changes requested):
**Verdict for the review above** ([#pullrequestreview-5186562968](https://github.com/MostroP2P/app/pull/419#pullrequestreview-5186562968)), which I submitted without one.

Requesting changes on finding **#1** only: a retake spawns a second `subscribe_single_order` task for the same order while the first is still alive, both on the `mostro-order-<id>` subscription id — duplicate handling of every d-tag event, and the first task to exit unsubscribes the REQ the survivor is riding. This PR is what makes retakes reachable, so the path it unlocks should not land already broken.

Everything else in that review is a fix-or-answer, not a block:

- **#2** (`persist_confirmed_take` deletes terminal history too) and **#4** (the Dart copy of `cancellation_wipes_history`) — either fix, or an answer in the doc comment saying why the case cannot arise.
- **#3** (`orderNoLongerActive` tells the user the opposite of what happened) — a new key plus five translations; worth doing here because no test can catch the copy being wrong.
- **#5**–**#8** and the contract question — happy to see them as follow-ups.

The design itself I would approve as it stands.

Catrya:
@grunch  Thanks for the strict pass. Everything is addressed below. Where I went a different way from the suggestion, the reason is stated. The branch also merges today's `main`, #412 included: the wipe paths of both PRs are unified, so every wipe (the daemon's `Canceled`, the public `canceled`, the sweep) now goes through `wipe_trade_row` and leaves the tombstone.

**#1 (HIGH), duplicate d-tag task on a retake.** Fixed in `774ce2a`. A registry of single-order tasks keyed by order id, claimed before the spawn. I made the newest claim *replace* the older one rather than making the second spawn a no-op. With a no-op, the retake would ride the first take's task, whose 30-minute idle window counts from its own last activity. A retake 25 minutes after the first take would then lose its subscription about 6 minutes in. Instead, the superseded task stops at its next wake, before handling the event, and leaves the subscription alone. The new task drops the REQ opened under the same id (nostr-sdk refuses an existing id), re-opens it with a fresh idle window, and is the only one allowed to unsubscribe. The relay replays the order's latest event on the new REQ, so nothing is lost in between. Pinned by `a_retake_replaces_the_earlier_single_order_task` and `subscribe_single_order_claims_before_it_spawns`. The in-loop check and the "superseded task does not unsubscribe" path need live relays; the retake E2E drive

**#2, `persist_confirmed_take` deletes termdoc comment (`4a13aa8`), not by scoping thedelete, because scoping would bring the bug back. A trade that truly ended leaves its order in a status mostrod
never takes it out of. A take needs `Pendin.rs`), and the only transitions back to`Pending` start from waiting states. So no confirmed take can follow real history. What can sit next to a new take
is a leftover of a never-active take, and oPR some of those read `Canceled`, because
the taker's own cancel wrote it up front. Aion_wipes_history` statuses would keepexactly those rows and leave two rows per order. The warning on a failed delete now says the order may have two rows
and that a lookup by order id can return th

**#3, `orderNoLongerActive` says the opposite.** Fixed in `4a13aa8` with a new key, `tradeNoLongerYours` ("You'longer part of this trade"), translated intone's register. I kept it neutral ratherthan "the order is back in the book", because this exit also fires when the maker cancelled (the order is gone)when an old trade is opened from a notificared). `orderNoLongerActive` stays on theinvoice screens, and its `@description` now names both of them.                                                **#4, the Dart copy of `cancellation_wipes_n bridged (`4a13aa8`). Asking Rust atruntime would be the bridge's first `#[frb( verified on web, and it would force a seaminto every widget test of the screen. Inste_cancellation_wipes_history_matches`(`status.rs`) reads the `OrderStatus → Tradatus.dart` and the `_cancelEndsTrade` setfrom source, and fails when they diverge from the Rust predicate. It also refuses to compile when a new `OrderStatus` variant is added, until it is placed. Mutation-checked: adding `inProgress` to the Dart set, drop`waitingPayment`, or widening the Rust preddoc comment points to the test.
                                                                                                               **#5, maker wipe leaves a note.** Done in `_order }`), so after any wipe the order hasno note. For the record, it isn't reachable today: notes are only written by the d-tag task, which only a take spawns. The invariant just shouldn't rest okers_order_leaves_no_wire_note_behind`,which plants the note.                                                                                         **#6, "book entry dropped" with nothing theone, None)` now logs "no book entry tosettle". The behaviour is unchanged, since othing.

**#7, `MyOrderScreen` bypasses the providerncels through `cancelOrderActionProvider`,
and a widget test drives it.

**#8, IndexedDB scan.** No change, for two reasons. The whole-store read predates this PR: the first-match version
went through `trade_document_by_order_id`, ents()` and then `find`s. The only new costis that the delete now runs on every confirmed take. And the guard comment is already there: "every one of them, as
SQLite's `DELETE … WHERE` does: `take_order row per order after a retake".

**Contract question, the in-flight cancel.** Written down in `06e1be1` under `cancel_order`, "Between the request and the daemon's answer":
- The screen goes home at once, and My Traduntil the `Canceled` or the public
`canceled` wipes the trade, normally within
- With no answer, the stale sweep settles as past its window. A maker's pending row is
left to the public `canceled`, since the swows.
- A refused cancel changes nothing locally,s told: `cancel_order` doesn't wait for the
reply.
- An active trade reads `Canceled` at once,s a cooperative request.

The last two are the fire-and-forget gap, w`cancel_order` should await`Canceled`/`CantDo` the way take and add-invoice do since #170.

Verification: `cargo test` (505), clippy, w `flutter test`. Each new test wasmutation-checked against the piece it pins.

rust/src/api/orders.rs
```diff
@@ -3890,14 +4362,25 @@ async fn subscribe_single_order(order_id: &str) {
...
+        if replaced {
+            // The earlier take's REQ is still open under this same id, and
+            // nostr-sdk refuses a subscribe whose id exists (it keeps the old
+            // filter and reports that per relay, not as an error). Drop it so
+            // this subscribe is accepted and owned by this task; the relay
+            // replays the order's latest event on the new REQ, so nothing is
+            // missed in between.
+            let _ = client.unsubscribe(&sub_id).await;
+        }
         if let Err(e) = client.subscribe(filter).with_id(sub_id.clone()).await {
```
grunch, line 4382: **LOW — a replaced task that fails to subscribe leaves the order with no REQ.**

On the `replaced` path the earlier take's REQ was just unsubscribed a few lines above. If this `subscribe` then fails, the claim is released and the task returns, and the earlier task — no longer current — will break at its next event without re-opening anything. Only reachable when there is no relay to subscribe on, so not blocking; the warning could say the order lost its d-tag subscription so a manual run reads it right.

rust/src/api/orders.rs
```diff
@@ -3921,54 +4404,24 @@ async fn subscribe_single_order(order_id: &str) {
 
             match timeout(remaining, rx.next()).await {
                 Ok(Some(ClientNotification::Event { event, .. })) => {
-                    if let Some(mut order) =
-                        crate::nostr::order_events::parse_order_event(&event, None)
+                    // A retake replaced this task while it waited: stop
```
grunch, line 4407: **Note, not blocking — "no event is applied twice" holds for events received after the claim.**

If the claim moves while this task is already inside `handle_single_order_event`, that event is applied here and again when the relay replays the order's latest event on the retake's fresh REQ. The upsert is idempotent, so nothing is wrong; just keeping the comment's claim precise.

lib/l10n/app_en.arb
```diff
@@ -836,7 +840,9 @@
   "noLightningWalletFound": "No Lightning wallet found on this device",
   "@noLightningWalletFound": {"description": "Snackbar shown when no app can handle the lightning: URI"},
   "orderNoLongerActive": "This order is no longer active",
-  "@orderNoLongerActive": {"description": "Neutral notice shown when the order reaches a terminal state (canceled, cooperatively canceled, canceled by admin, or expired) while the user is on the pay invoice screen"},
+  "@orderNoLongerActive": {"description": "Neutral notice shown when the order reaches a terminal state (canceled, cooperatively canceled, canceled by admin, or expired) while the user is on the add-invoice or pay-invoice screen"},
+  "tradeNoLongerYours": "You're no longer part of this trade",
```
grunch, line 1633: **Must before merge — this region conflicts with `main`.**

`main` reformatted `@orderNoLongerActive` (and `@cancelTradeDialogContent`, further up) into multi-line objects, so both hunks this branch edits now conflict; GitHub reports the PR as not mergeable. One more merge of `origin/main` resolves it — keep your one-line style for the new `@tradeNoLongerYours` or match `main`'s, either is fine.

grunch (dismissed):
## Review — round 2 (strict pass on `774ce2a`, `4a13aa8`, `06e1be1`)

I went through the three response commits against the eight findings and the contract question of the first pass, ran the branch's CI state, and mutation-checked the new source-reading test myself.

**Verdict: approve — the code answers every finding.** Two things still stand between this and the merge button, neither of them code.

### Must do before merge

1. **The branch conflicts with `main`** (`mergeable: CONFLICTING`). `lib/l10n/app_en.arb` conflicts in two hunks: `main` reformatted the `@cancelTradeDialogContent` and `@orderNoLongerActive` descriptions into multi-line objects, and this branch edits both lines (and adds `tradeNoLongerYours` right after the second). One more `Merge origin/main` resolves it; nothing else in the branch collides.
2. **The response comment is unreadable for #2, #4, #5, #6, #8 and the contract answer.** The comment body on GitHub itself carries overlapping text (e.g. `deletes termdoc comment`, `A take needs `Pendin.rs``, `Aion_wipes_history` statuses would keepexactly`) — it looks like a paste from a terminal that wrapped over itself. I reconstructed what was done from the diff, but the comment is the record a future reader will find from the finding thread. Please repost it.

### Findings, one by one

| # | First pass | Now | Verified |
|---|---|---|---|
| 1 | HIGH — two live d-tag tasks on one subscription id | **Fixed** (`774ce2a`) | Registry `order id → generation`, claimed *before* the spawn; the superseded task breaks at its next event without handling it and `release_single_order_task` returns `false` so it never unsubscribes; the new task drops the old REQ under the same id before re-subscribing. Replace-not-no-op is the right call for the idle window reason you give. `contracts/orders.md` states the one-task rule. |
| 2 | MEDIUM — delete not scoped to never-active rows | **Answered in the doc comment** (`4a13aa8`) | The argument holds: mostrod only takes a `Pending` order, and the only way back to `Pending` is from a waiting state, so a confirmed take can never follow a terminal row of the same order — while scoping *would* keep the pre-fix `Canceled` leftovers. The failed-delete warning now states the consequence (two rows, `LIMIT 1` lookup). Accepted. |
| 3 | MEDIUM — `orderNoLongerActive` says the opposite | **Fixed** (`4a13aa8`) | New `tradeNoLongerYours` in all five `.arb` files, neutral copy (correct: this exit also fires when the maker cancelled), `@description` of both keys names their screens, the widget test asserts the new key. `orderNoLongerActive` stays on the two invoice screens only. |
| 4 | MEDIUM — `_cancelEndsTrade` unpinned from Rust | **Fixed** (`4a13aa8`) | `the_trade_screen_copy_of_cancellation_wipes_history_matches` reads both `trade_status.dart` and the screen's set via `include_str!`, maps through `tradeStatusFromOrderStatus`, and the exhaustive `match` forces a new `OrderStatus` variant to be placed. I checked the parser against the actual shape of both Dart functions, and reproduced the mutation myself: with `TradeStatus.inProgress` added to the Dart set the test fails with `InProgress keeps its history, yet the screen treats TradeStatus.inProgress as ending the trade`; restored, it passes. |
| 5 | LOW — maker wipe leaves a `wire_orders` note | **Fixed** (`06e1be1`) | `wipe_never_active_trade` forgets the note on the maker branch; `a_wiped_makers_order_leaves_no_wire_note_behind` plants one and asserts it is gone. |
| 6 | LOW — "book entry dropped" with nothing there | **Fixed** (`4a13aa8`) | `(None, None)` arm before the catch-all, own log line. |
| 7 | NIT — `MyOrderScreen` bypasses the provider | **Fixed** (`06e1be1`) | Goes through `cancelOrderActionProvider`; widget test drives the seam. |
| 8 | NIT — IndexedDB full scan per take | **No change, accepted** | Pre-existing whole-store read; the guard comment is already there. |
| — | Contract: the in-flight cancel | **Written** (`06e1be1`) | The four cases are in `cancel_order` → *Between the request and the daemon's answer*, including the honest gap that a refused cancel is never surfaced. Worth its own issue: `cancel_order` awaiting `Canceled`/`CantDo` like take and add-invoice do. |

### Two small notes on `774ce2a` (not blocking)

- **Replaced task, then `subscribe` fails.** If the retake's task has already unsubscribed the earlier REQ and its own `client.subscribe` then fails, it releases its claim and returns — at which point the order has *no* REQ and the earlier task, no longer current, will break at its next event without re-opening anything. It only happens when there is no relay to subscribe on, so the trade is in trouble anyway, but the log line for that branch could say the order lost its d-tag subscription.
- **One event can still be applied twice**, in the window where the old task is inside `handle_single_order_event` when the claim moves and the relay then replays the same latest event on the new REQ. Harmless — the upsert is idempotent — just noting it against the "no event is applied twice" comment, which is true for events *received* after the claim.

### Checked

- CI is green on the head commit `06e1be1` (Rust build/test/clippy/wasm, Flutter analyze/test, web build + smoke).
- All four tests named in the response exist at the stated paths.
- No `MutexGuard` from `single_order_tasks()` is held across an `await`.

Catrya:
@grunch conflicto resuelto

PR #448, opened 2026-09-13
fix(trades): handle takes left Canceled by the old optimistic cancel

### LINKED ISSUE #434: A canceled take shows as the user's pending order once the order is back in the book
A taker who cancelled before the trade went active can end up with a trade row in `Canceled`. The order itself is back in the public book as `pending`, open to anyone. My Trades and the trade screen show that public `pending` as the status of the user's trade:

- My Trades lists it as **Pending**.
- The trade screen shows it as a pending order with a **Cancel** button.
- Tapping Cancel signs with the old trade key. The daemon answers `CantDo(IsNotYourOrder)`, and the local cancel also removes the order from the in-memory book, so a live order vanishes from the user's book until the next event or restart.

The order book itself is right: the order is not the user's there.

## Steps to reproduce (on `main`)

1. Take someone else's order.
2. Cancel it while it is still `waiting-buyer-invoice` or `waiting-payment`.
3. The daemon republishes the order as `pending`. Leave it untaken.
4. Restart the app and open My Trades.

Log from a reproduction:

```
orders: skip replayed Canceled order=807983cb: already Canceled
orders: cancel published for order=807983cb trade_index=71
daemon-msg: action=CantDo order_id=Some(807983cb-…) … payload=CantDo(Some(IsNotYourOrder))
```

## Cause

Three things combine:

1. **The row is left `Canceled`.** On `main`, `cancel_order` marks the trade row `Canceled` as soon as the cancel is published, including for trades that never went active. When the daemon's `Canceled` arrives it finds the row "already Canceled" and skips the wipe it would otherwise do. #419 removes that write for never-active trades, so no new rows like this are created, but the ones already on disk stay.
2. **A public `pending` enters the book unchecked.** `ingest_order_event_with` compares only non-`pending` statuses against the local row (`wire_status_applies`). A `pending` event goes into the book as is.
3. **The UI reads the book before the row.**
   - `tradeStatusLookupProvider` calls `getOrder` (the book) before `listTrades` (the row).
   - `trades_list_item.dart` uses `effectiveStatus = live ?? trade.status`.
   - `TradeDetailScreen._status()` reads `tradeStatusProvider`.

## What the user sees, by what the order does next

| Public order | Shown for the canceled take | Correct? |
|---|---|---|
| `pending` | Pending + Cancel | ❌ |
| taken by someone else (`in-progress`) | Canceled (the local status wins in the book) | ✅ |
| completed by someone else (`success`) | **Completed**, and it stays | ❌ |
| `canceled` / expired | Canceled | ✅ |

## Who is affected

Native users with a take they cancelled before it went active, on a build without #419, while that order is still published. The `pending` case lasts at most until the order expires or someone takes it. The `success` case is rarer, but it does not go away on its own.

## Proposed fix (Dart only, no Rust change)

In the two places that pick the status to show, `TradeDetailScreen._status()` and `effectiveStatus` in `trades_list_item.dart`, **the row's status wins only when all three hold**:

- the trade is a take (`!order.isMine`),
- the row is hard-terminal (the success or canceled family),
- the book says `pending`.

A public `pending` can never be the status of a taker's trade: it means nobody holds the order.

Why it has to be this narrow:

- **A plain "a terminal row wins" would regress one case.** A taker who requests a cooperative cancel on an active trade gets an optimistic `Canceled` row. If the peer does not agree and the trade completes, the book's `success` is what shows Completed today.
- **`tradeStatusLookupProvider` itself must not change.** The Kind 38383 ingest only syncs the row for makers, so for a taker a `success` can arrive through the book alone. `tradeStatusProvider` also stops polling at a terminal status, which would freeze a stale value across a retake.

Tests (widget, for both the list item and the trade screen):

- take + `Canceled` row + book `pending` → Canceled, no Cancel button;
- take + `Canceled` row + book `success` → Completed (unchanged);
- maker + `Canceled` row + book `pending` → pending (unchanged);
- take + non-terminal row + book `pending` → unchanged.

## Out of scope

The `success` case (Completed shown for a trade someone else finished) needs the legacy rows themselves cleaned up: the old optimistic-`Canceled` rows of never-active takes. That goes in the separate cleanup issue.

Related: #417, #419.

The issue's "Out of scope" section points to the cleanup issue for old Canceled rows, which isn't open yet. If you open that one first, swap the reference for its number.

Catrya:
Closes #434.

Builds before #419 marked a take `Canceled` as soon as its cancel was published, even when the trade never went active. The daemon then put the order back in the book as `pending`, but the row stayed. #419 stops writing those rows; this PR fixes how the ones already on disk are shown and handled:

- **The trade screen showed the book's public `pending` as the trade's status.** The user saw their own pending order, with a Cancel the daemon refuses (`IsNotYourOrder`). Since #436 the My Trades list keeps an ended row's status, so the list said *cancelled* while the screen said *pending*.
- **The take screen treated any row on the order as participation.** Opening the order from the book, or tapping *Take*, redirected to that dead trade, so the user could never take the order again. That is #417 again, for rows written by the old cancel.

No Rust change: `take_order` already accepts this case (it only needs the order to be `pending` in the book), and `persist_confirmed_take` replaces every earlier row of the order.

### 1. `fix(trades): a public pending is never a take's status`

One rule, `shownTradeStatus` (`trade_state_provider.dart`), now decides what the My Trades list and `TradeDetailScreen` show. The trade row wins over the live status (`tradeStatusProvider`, which reads the order book first) in two cases:

- **The row has ended** (the success or cancelled family). Whatever the book says about the order later is no longer this trade. The list already did this; the screen now matches it.
- **It is a take and the book says `pending`.** A public `pending` means nobody holds the order, so it is never a take's status. This also covers a take parked at `WaitingTakerBond`, whose order is still `pending` in public.

Every other live status still wins over an open row. In the trade screen the rule applies everywhere the live status was read: the status itself, the post-cancel decision, the cancel dialog's copy, and the haptic nudge on a step change (so a book change the row outranks does not vibrate).

### 2. `fix(order): a take that already ended no longer blocks taking the order`

`tradeRoleLookupProvider`, which only the take screen uses, now answers through `participatingRole`:

- A take whose row has ended **does not count** as participation. An order whose trade truly ended never goes back to `Pending` in mostrod, so once the order can be taken again, such a row can only be a leftover of a take that never went active.
- A maker's row always counts, and so does any row still open.
- It reads every row of the order, not one: a database from before #419 can hold two rows for the same order.

`take_order_screen.dart` itself is untouched.

## Known trade-off

With the first rule, a cooperative cancel requested on an active trade still marks the row `Canceled` right away (the remaining optimistic write in `apply_local_cancel`). If the peer never agrees and the trade completes, both the list and the screen show it as cancelled. The list already behaved this way since #436. The fix belongs to that optimistic write, and it will be tracked separately.

## Tests

- **`trade_state_provider_test.dart`**:
  - `shownTradeStatus` over every combination of row status, live status and take/maker;
  - `participatingRole` and `isEndedTake`: no row, open takes, ended takes, maker rows, and two rows on the same order.
- **`trade_rows_provider_test.dart`**: a take keeps its own status over a public `pending`; a maker's order still follows the book.
- **`trade_detail_screen_test.dart`**:
  - a leftover `Canceled` take reads cancelled, without Cancel, over a public `pending`;
  - it stays cancelled when someone else completes the order;
  - an open take keeps its own step over a public `pending`;
  - a maker's order still reads `pending`;
  - a book change the row outranks does not trigger the haptic nudge.
- **`take_order_screen_test.dart`**: an open take lands on its trade; an ended take leaves the order takeable. The test harness now runs under a router, so the redirect is observable.

With each rule disabled, the tests that cover it fail. `flutter analyze` is clean and `flutter test` passes (1042 tests).

## Manual test

Against a local mostrod, with the same database for both builds:

1. **Pre-#419 build** (`40ccde9`, in a separate worktree): took and cancelled two orders while they were still waiting. Both rows stayed `Canceled` (`skip replayed Canceled … already Canceled`) while the daemon republished the orders as `pending`.
2. **This branch**:
   - after a restart, both rows still read `Canceled` in the database;
   - opening each order from the book showed the take screen, with no redirect to the old trade;
   - both were taken again (fresh trade indexes), cancelled, and wiped as #419 does;
   - one order was taken a third time, and the earlier take's d-tag task was replaced (`superseded by a retake`);
   - the old rows were replaced by the new takes, and no `IsNotYourOrder` was sent from an old key.

## Notes

- `feat/bond-1b-pay-screen` and `feat/bond-1c-gate-flip` touch `take_order_screen.dart` and `trade_detail_screen.dart` in other sections. Whichever lands second may need a small merge.
- Still open, outside this PR:
  - the optimistic `Canceled` on an active cooperative cancel;
  - cleaning up leftover `Canceled` rows of takes that never went active.

### DIFF
--- a/lib/features/order/providers/trade_state_provider.dart
+++ b/lib/features/order/providers/trade_state_provider.dart
@@ -41,12 +41,46 @@ final tradeStatusLookupProvider =
       },
     );
 
-/// Reads the local user's role in a trade through the bridge; injectable so
-/// screens that must know whether they already participate can be tested
-/// without the Rust side.
+/// The local user's role in an order they still take part in
+/// ([participatingRole]), read through the bridge; injectable so screens that
+/// must know whether they already participate can be tested without the Rust
+/// side.
 final tradeRoleLookupProvider = Provider<Future<TradeRole?> Function(String)>(
-  (ref) => (orderId) => orders_api.getTradeRole(orderId: orderId),
+  (ref) =>
+      (orderId) async =>
+          participatingRole(await orders_api.listTrades(), orderId),
 );
+
+/// The role of the user's trade on [orderId] among [trades], or null when
+/// they no longer take part in it: no row at all, or only a take that has
+/// ended ([isEndedTake]).
+///
+/// Every row for the order is read, not just one: a database from before
+/// takes replaced their order's earlier row can hold two, and a live one
+/// among them still makes the user a participant.
+TradeRole? participatingRole(Iterable<TradeInfo> trades, String orderId) {
+  for (final trade in trades) {
+    if (trade.order.id == orderId && !isEndedTake(trade)) return trade.role;
+  }
+  return null;
+}
+
+/// Whether [trade] is a take whose row has ended, which leaves the user
+/// nothing to follow on its order.
+///
+/// A trade that truly ended leaves its order in a status mostrod never takes
+/// it out of: a take needs `Pending`, and only a waiting state goes back to
+/// it. So once the order can be taken again, such a row is what a take that
+/// never went active left behind: older builds marked it `Canceled` as soon
+/// as its cancel went out. Holding on to it sent the user to that dead trade
+/// instead of letting them take the order again (#434). Rust already takes
+/// over such a row: the confirmed take replaces every earlier row of its
+/// order.
+///
+/// Never a maker's row: its order is theirs, and the take screen is not
+/// where they manage it.
+bool isEndedTake(TradeInfo trade) =>
+    !trade.order.isMine && isTerminalTradeStatus(trade.order.status);
 
 /// Takes an order through the bridge; injectable so the take screen's
 /// outcomes (loading, already taken, rejected) can be tested without Rust.
--- a/lib/features/order/providers/trade_state_provider.dart
+++ b/lib/features/order/providers/trade_state_provider.dart
@@ -88,7 +88,7 @@ final tradeStatusProvider = StreamProvider.family
         final status = await lookup(orderId);
         if (status != null) {
           yield status;
-          if (_isTerminal(status)) return;
+          if (isTerminalTradeStatus(status)) return;
         }
         await Future.delayed(const Duration(seconds: 2));
       }
@@ -111,8 +111,9 @@ final tradeUpdatesProvider = StreamProvider.autoDispose<TradeUpdate>((
   }
 });
 
-/// Whether the UI can stop polling. Escrow settlement still awaits payout.
-bool _isTerminal(OrderStatus s) => const {
+/// Whether a trade in [s] has ended: nothing moves it out again, so the UI
+/// can stop polling. Escrow settlement still awaits payout.
+bool isTerminalTradeStatus(OrderStatus s) => const {
   OrderStatus.success,
   OrderStatus.settledByAdmin,
   OrderStatus.completedByAdmin,
@@ -121,6 +122,33 @@ bool _isTerminal(OrderStatus s) => const {
   OrderStatus.cooperativelyCanceled,
   OrderStatus.canceledByAdmin,
 }.contains(s);
+
+/// The status a trade shows: its [row]'s persisted one, or the [live] one
+/// from [tradeStatusProvider], which reads the order book first.
+///
+/// The row wins in two cases:
+/// * **It has ended** ([isTerminalTradeStatus]). Whatever the book says about
+///   the order later is no longer this trade. The one way such a row can be
+///   wrong is the cancel's optimistic write on an active trade: a cooperative
+///   cancel the peer never accepts, on a trade that then completes.
+/// * **It is a take ([isTake]) and the book says `pending`.** A public
+///   `pending` means nobody holds the order, so it is never a take's status.
+///   Older builds marked a take `Canceled` as soon as its cancel went out,
+///   even before it went active; once the daemon put the order back in the
+///   book, that `pending` read as the user's own order, with a Cancel the
+///   daemon refuses (`IsNotYourOrder`). A take parked at `WaitingTakerBond`
+///   is another: publicly its order is still `pending`.
+///
+/// Otherwise the live status, or the row's while there is none yet.
+OrderStatus shownTradeStatus({
+  required OrderStatus row,
+  required OrderStatus? live,
+  required bool isTake,
+}) {
+  if (live == null || isTerminalTradeStatus(row)) return row;
+  if (isTake && live == OrderStatus.pending) return row;
+  return live;
+}
 
 /// Loads the buyer/seller role for a trade from the persistent DB.
 ///
--- a/lib/features/trades/screens/trade_detail_screen.dart
+++ b/lib/features/trades/screens/trade_detail_screen.dart
@@ -36,6 +36,7 @@ import 'package:mostro/shared/widgets/mostro_reactive_button.dart';
 import 'package:mostro/src/rust/api/disputes.dart' as disputes_api;
 import 'package:mostro/src/rust/api/orders.dart' as orders_api;
 import 'package:mostro/src/rust/api/reputation.dart' as reputation_api;
+import 'package:mostro/src/rust/api/types.dart' show TradeInfo;
 
 export 'package:mostro/features/trades/models/trade_status.dart';
 
@@ -189,8 +190,22 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
   /// await: the seller's payment can land while the cancel dialog is open.
   TradeStatus _liveStatus(TradeStatus fallback) {
     final live = ref.read(tradeStatusProvider(widget.orderId)).valueOrNull;
-    return live == null ? fallback : tradeStatusFromOrderStatus(live);
+    if (live == null) return fallback;
+    final trade = ref.read(tradeInfoProvider(widget.orderId)).valueOrNull;
+    return tradeStatusFromOrderStatus(_shown(live, trade));
   }
+
+  /// What [live] shows for this trade once its [trade] row is read in
+  /// ([shownTradeStatus]); [live] itself while the row has not loaded, or
+  /// when there is none.
+  static OrderStatus _shown(OrderStatus live, TradeInfo? trade) =>
+      trade == null
+          ? live
+          : shownTradeStatus(
+            row: trade.order.status,
+            live: live,
+            isTake: !trade.order.isMine,
+          );
 
   Future<void> _cancelOrder(TradeStatus status) async {
     final l10n = AppLocalizations.of(context);
@@ -207,8 +222,14 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
                     dialogRef
                         .watch(tradeStatusProvider(widget.orderId))
                         .valueOrNull;
+                final trade =
+                    dialogRef
+                        .watch(tradeInfoProvider(widget.orderId))
+                        .valueOrNull;
                 final now =
-                    live == null ? status : tradeStatusFromOrderStatus(live);
+                    live == null
+                        ? status
+                        : tradeStatusFromOrderStatus(_shown(live, trade));
                 return Text(_cancelDialogContent(l10n, now));
               },
             ),
@@ -405,7 +426,8 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
       debugPrint('[TradeDetailScreen] trade status failed: ${live.error}');
     }
     if (!live.hasValue) return TradeStatus.loading;
-    final status = tradeStatusFromOrderStatus(live.value!);
+    final trade = ref.watch(tradeInfoProvider(widget.orderId)).valueOrNull;
+    final status = tradeStatusFromOrderStatus(_shown(live.value!, trade));
     if (status != TradeStatus.pendingRating) return status;
     final rating = ref.watch(tradeRatingProvider(widget.orderId));
     if (rating.isLoading && !rating.hasValue) return TradeStatus.loading;
@@ -439,13 +461,18 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
     // A step advanced by the relay: a nudge, the crossfades below, and a
     // fresh deadline — every step has its own expiration, owned by a
     // different party, so the clock loaded for the previous step is stale.
+    // Judged on what the screen shows: a book change the row outranks is
+    // not a step of this trade.
     ref.listen<AsyncValue<OrderStatus>>(tradeStatusProvider(widget.orderId), (
       previous,
       next,
     ) {
+      final trade = ref.read(tradeInfoProvider(widget.orderId)).valueOrNull;
       final before = previous?.valueOrNull;
       final after = next.valueOrNull;
-      if (before != null && after != null && before != after) {
+      if (before != null &&
+          after != null &&
+          _shown(before, trade) != _shown(after, trade)) {
         HapticFeedback.mediumImpact();
         _loadExpiresAt();
       }
--- a/test/features/order/screens/take_order_screen_test.dart
+++ b/test/features/order/screens/take_order_screen_test.dart
@@ -4,6 +4,8 @@ import 'package:clock/clock.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
+import 'package:go_router/go_router.dart';
+import 'package:mostro/core/app_routes.dart';
 import 'package:mostro/core/app_theme.dart';
 import 'package:mostro/core/automation/automation_id.dart';
 import 'package:mostro/core/automation/automation_ids.dart';
@@ -18,6 +20,7 @@ import 'package:mostro/src/rust/api/types.dart';
 import 'package:mostro/shared/utils/fiat_currencies.dart';
 
 import '../../../support/fake_orders.dart';
+import '../../../support/fake_trades.dart';
 import '../../../support/provider_harness.dart';
 
 const _id = '09150348-1a2b-4c3d-8e9f-0a1b2c3d99b5';
@@ -25,8 +28,9 @@ const _dark = OrderDetailPalette.dark;
 const _book = OrderBookPalette.dark;
 
 /// Pumps the take-order screen over a book fed by [books], with the node's
-/// rate and the (absent) trade role stubbed. 1 000 ARS is 1 000 sats at
-/// the stubbed rate, so the estimates are easy to read.
+/// rate stubbed and the user's trade rows being [trades] (none by default).
+/// 1 000 ARS is 1 000 sats at the stubbed rate, so the estimates are easy to
+/// read.
 typedef _Take = Future<TradeInfo> Function({
   required String orderId,
   required TradeRole role,
@@ -39,6 +43,7 @@ Future<StreamController<List<OrderItem>>> _pump(
   bool isBuying = true,
   double? rate = 100000000,
   _Take? take,
+  List<TradeInfo> trades = const [],
 }) async {
   tester.view.physicalSize = const Size(360, 760);
   tester.view.devicePixelRatio = 1.0;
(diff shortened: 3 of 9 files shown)


lib/features/trades/screens/trade_detail_screen.dart
```diff
@@ -405,7 +426,8 @@ class _TradeDetailScreenState extends ConsumerState<TradeDetailScreen> {
       debugPrint('[TradeDetailScreen] trade status failed: ${live.error}');
     }
     if (!live.hasValue) return TradeStatus.loading;
-    final status = tradeStatusFromOrderStatus(live.value!);
+    final trade = ref.watch(tradeInfoProvider(widget.orderId)).valueOrNull;
+    final status = tradeStatusFromOrderStatus(_shown(live.value!, trade));
```
grunch, line 430: _🎯 Functional Correctness_ | _🟠 Major_

**The leftover take can still flash `pending` with Cancel while the row is loading — the exact bug this PR fixes.**

`_status()` resolves as soon as `tradeStatusProvider` has a value, but `trade` comes from `tradeInfoProvider`, which awaits `rawTradesProvider` (a full `listTrades()` round trip). While that future is pending, `_shown(live, null)` returns the live status unchanged — the book's public `pending`.

Nothing else holds the screen in `loading` during that window: `build()` gates on `_isBuyer()`, which resolves from `tradeRoleProvider` or `tradeRoleFromDbProvider` (the indexed `getTradeRole`), independently of `rawTradesProvider`. For a leftover `Canceled` take opened cold (from a notification, the chat header, or after a restart), the role typically arrives first, so for one or more frames the screen renders `TradeStatus.pending` → `TradeView` offers `trade.cancel` → a tap publishes the `cancel` the daemon answers with `CantDo(IsNotYourOrder)` and the local cancel removes the live order from the book (#434, third bullet).

The same `null`-row fallback also applies to the haptic listener (L470): a `pending → in-progress` book change that lands before the row loads still nudges.

The tests don't catch this because `_pumpRoutedTradeDetail` resolves `loadTrades` immediately.

Suggested fix: don't let a live `pending` through until the row is known.

```dart
TradeStatus _status() {
  final live = ref.watch(tradeStatusProvider(widget.orderId));
  // …
  if (!live.hasValue) return TradeStatus.loading;
  final tradeAsync = ref.watch(tradeInfoProvider(widget.orderId));
  // A public `pending` may be a take's leftover (#434): only the row can
  // say, so hold `loading` rather than offer a Cancel the daemon refuses.
  if (live.value == OrderStatus.pending && tradeAsync.isLoading) {
    return TradeStatus.loading;
  }
  final status = tradeStatusFromOrderStatus(
    _shown(live.value!, tradeAsync.valueOrNull),
  );
  // …
}
```

and add a widget test where `loadTrades` completes after the status stream has emitted `pending`, asserting `_cancelButton()` is never found.

<details>

```
In lib/features/trades/screens/trade_detail_screen.dart, _status() (around line 429): when the
live status is OrderStatus.pending and tradeInfoProvider(widget.orderId) is still loading, return
TradeStatus.loading instead of passing a null row to _shown(). Apply the same guard to the haptic
listener (around line 470): skip the nudge while the row is unresolved. Add a widget test in
test/features/trades/trade_detail_screen_test.dart where the status stream emits pending before
loadTrades completes with a Canceled take row, and assert the Cancel button never appears.
```

</details>

lib/features/order/providers/trade_state_provider.dart
```diff
@@ -41,13 +41,47 @@ final tradeStatusLookupProvider =
...
-/// without the Rust side.
+/// The local user's role in an order they still take part in
+/// ([participatingRole]), read through the bridge; injectable so screens that
+/// must know whether they already participate can be tested without the Rust
+/// side.
 final tradeRoleLookupProvider = Provider<Future<TradeRole?> Function(String)>(
-  (ref) => (orderId) => orders_api.getTradeRole(orderId: orderId),
+  (ref) =>
+      (orderId) async =>
+          participatingRole(await orders_api.listTrades(), orderId),
```
grunch, line 51: _🎯 Functional Correctness_ | _🟠 Major_

**The role lookup now throws on a storage error; before it returned `null`. The take screen doesn't handle that.**

The old implementation, `orders_api.getTradeRole`, never failed: `get_trade_role` logs a DB error and returns `Ok(None)`. `orders_api.listTrades()` propagates it (`db.list_trades().await?`), so this closure can now complete with an error.

Both call sites in `take_order_screen.dart` assume it can't:

- `initState` calls `_redirectIfParticipant()` unawaited with no `catch`, so the error surfaces as an unhandled async exception (reported to the zone / Crashlytics in release).
- `_onTakeOrder` awaits it before `setState(() => _cta = TakeOrderCta.loading)` with no `try`. The future throws, `_cta` stays `idle`, and **the Take button silently does nothing** — no snackbar, no retry hint.

Suggested fix: keep the old contract at the seam (an unreadable store means "not a known participant", as `getTradeRole` did), and log it.

```dart
final tradeRoleLookupProvider = Provider<Future<TradeRole?> Function(String)>(
  (ref) => (orderId) async {
    try {
      return participatingRole(await orders_api.listTrades(), orderId);
    } catch (e, st) {
      debugPrint('[tradeRoleLookup] listTrades failed for $orderId: $e\n$st');
      return null;
    }
  },
);
```

Please add a take-screen test with a throwing lookup that asserts the take is still dispatched (or, if you prefer failing closed, that an error snackbar appears — but not a dead button).

<details>

```
In lib/features/order/providers/trade_state_provider.dart, tradeRoleLookupProvider (line ~48):
wrap the listTrades() call in try/catch, log the error, and return null so the lookup keeps the
never-throws contract getTradeRole had. Add a test in
test/features/order/screens/take_order_screen_test.dart overriding tradeRoleLookupProvider with a
throwing function, and assert tapping "Take order" still dispatches takeOrderActionProvider.
```

</details>

lib/features/order/providers/trade_state_provider.dart
```diff
@@ -41,13 +41,47 @@ final tradeStatusLookupProvider =
...
 );
 
+/// The role of the user's trade on [orderId] among [trades], or null when
+/// they no longer take part in it: no row at all, or only a take that has
+/// ended ([isEndedTake]).
+///
+/// Every row for the order is read, not just one: a database from before
+/// takes replaced their order's earlier row can hold two, and a live one
+/// among them still makes the user a participant.
+TradeRole? participatingRole(Iterable<TradeInfo> trades, String orderId) {
```
grunch, line 61: _⚡ Performance_ | _🟡 Minor_

**Every take-screen lookup now deserialises the user's whole trade history — twice per take.**

`tradeRoleLookupProvider` runs on `initState` and again on the Take tap. `list_trades` reads every `trades` row, `serde_json`-decodes each blob, sorts them, and FRB copies the full `Vec<TradeInfo>` into Dart, just to filter one `order.id`. `get_trade_role` hit the `idx_trades_order_id` expression index instead. For a long-lived account (hundreds of closed trades, each with `bond` / peer snapshots) that is a noticeable delay right when the user taps Take.

Not blocking for the fix, but since the rule lives in Dart only because Rust has no "all rows for an order" query, consider a narrow bridge call instead (e.g. `list_trades_for_order(order_id)` using the existing `json_extract(data, '$.order.id') = ?` index) and keep `participatingRole` as the pure rule over its result.

test/features/order/screens/take_order_screen_test.dart
```diff
@@ -51,7 +56,9 @@ Future<StreamController<List<OrderItem>>> _pump(
         yield [order];
         yield* books.stream;
       }),
-      tradeRoleLookupProvider.overrideWithValue((_) async => null),
+      tradeRoleLookupProvider.overrideWithValue(
+        (orderId) async => participatingRole(trades, orderId),
```
grunch, line 60: _🧪 Test Coverage_ | _🟡 Minor_

**This override re-implements the production provider, so its wiring is untested.**

The harness replaces `tradeRoleLookupProvider` with `participatingRole(trades, orderId)`. The production body (`participatingRole(await orders_api.listTrades(), orderId)`) is never executed by any test, so a regression back to `getTradeRole`, a wrong argument order, or the error path from the comment above would all keep these tests green.

Consider putting the trades read behind its own seam (e.g. a `tradeRowsLookupProvider = Provider<Future<List<TradeInfo>> Function()>`), have `tradeRoleLookupProvider` compose it, and override **that** here, so the real composition — including the error handling — is what the take-screen tests exercise.

grunch (changes requested):

> [!CAUTION]
> **Merge conflict with `main` — CI has not run on this PR.** `mergeable: CONFLICTING`. GitHub does not run `pull_request` workflows on a head it cannot merge, so the only check reported is CodeRabbit (skipped). The PR's `flutter analyze` / `flutter test` claims are local-only until this is rebased.
>
> The textual conflict is confined to `test/features/order/screens/take_order_screen_test.dart`: since this branch's base (`68ab52b`), `main` added bond-policy overrides to `_pump` (`mostroNodeProvider`, `bondEstimateProvider`, a `node` / `bondEstimate` parameter pair) and new bond-notice tests in the same group this PR extends. Keep both sides: the router harness and `trades` parameter from here, the node overrides from `main`. `trade_rows_provider.dart`, `trade_detail_screen.dart`, `trade_rows_provider_test.dart` and `fake_trades.dart` auto-merge, but please re-run the suites after rebasing — `main`'s rows provider now also reads the claim store, and `TradeDetailScreen._cancelOrder` gained a maker-bond branch keyed on the status this PR reshapes.

<details>
<summary>🧹 Nitpick / verified-OK notes (3)</summary><blockquote>

- **`isTerminalTradeStatus` vs the removed `_terminal`** — verified identical (the same 7 statuses), so the My Trades list's ended-row behaviour is unchanged by the refactor. ✅
- **Interaction with the bond flow now on `main`** — `shownTradeStatus` is correct for both bond windows: a taker at `WaitingTakerBond` whose order is publicly `pending` keeps the row's status (previously the trade screen showed `pending`, so this PR also fixes that); a maker at `WaitingMakerBond` has no book entry, so the live status already falls back to the row, and once the bond locks the book's `pending` (maker, `isTake: false`) wins as it should. ✅
- **`trade_action_listener.dart:59` still calls `orders_api.getTradeRole` directly** — that bridge call is `get_trade_by_order_id` (`LIMIT 1`, no `ORDER BY`), the same "one arbitrary row" lookup `participatingRole` was written to avoid. Not a regression of this PR (#419 makes duplicate rows unlikely going forward), but worth a follow-up so the two role lookups agree on pre-#419 databases.

</blockquote></details>

<details>

**Reviewed commit:** `c0aa804` (2 commits: `897bbba`, `c0aa804`)
**Files reviewed:** 9 — `trade_state_provider.dart`, `trade_rows_provider.dart`, `trade_detail_screen.dart`, `specs/004…/orders.md`, and the 5 test / support files.
**Profile:** strict

**Walkthrough**

| Area | Change |
|---|---|
| `trade_state_provider.dart` | New `shownTradeStatus` (row vs live), `participatingRole` / `isEndedTake`; `tradeRoleLookupProvider` now scans `listTrades()` instead of `getTradeRole`; `_isTerminal` → public `isTerminalTradeStatus`. |
| `trade_rows_provider.dart` | Drops its private `_terminal` set and delegates to `shownTradeStatus`. |
| `trade_detail_screen.dart` | Status, cancel-dialog copy, post-cancel decision and haptic nudge all go through `_shown(live, row)`. |
| `take_order_screen_test.dart` | Harness moves under a `GoRouter` so the participant redirect is observable; two new tests. |
| `specs/004…/orders.md` | Documents both rules. |

</details>

Catrya:
@grunch Thanks — all four are fixed, and the merge is done. Details below, including two places where I landed on something slightly different from the suggestion.

## Merge and CI

Merged `main` into the branch. The only textual conflicts were the ones you predicted, in the take-screen harness (`typedef _Take`, and the `node` / `bondEstimate` pair next to this PR's `trades`), resolved by keeping both sides, plus two in files `main` reshaped since: `tradeStatusProvider` (now `_followTrade`) and the cancel dialog (now with the peer-asked notice). The dialog already watched the row for `peerAsked`, so the merged version reads it once and uses it for both that and the shown status.

Re-ran the suites after merging: `flutter analyze` clean, `flutter test` green (1546). `_cancelOrder`'s maker-bond branch keys on `TradeStatus.waitingBond`, which this PR's rule only ever produces for a **take** (`isTake` is `!order.isMine`), so the maker path is untouched. CI will have its say on the push.

## 1. The leftover take can flash pending with Cancel while the row loads

Right, and thank you — I had weighed this as "one frame" and it is not: the row is a full trades read while the role comes from the indexed `get_trade_role`, so on a cold open the role wins the race by a wide margin.

Fixed as suggested, with one change: the guard is `isLoading && !hasValue`, not `isLoading`. `rawTradesProvider` is invalidated on every `TradeUpdate`, and a refreshing provider stays `isLoading` while keeping its previous value, so gating on `isLoading` alone would flash `loading` on every update. `_status()` uses the same shape the rating lookup already uses a few lines below.

The haptic listener now returns early for the same reason: without the row, a change into or out of a public `pending` says nothing yet.

Tests added: a take whose `loadTrades` completes after the status stream has emitted `pending` (the Cancel button is never found, and the screen reads cancelled once the row lands), and a book change under an unresolved row that must not nudge. Each fails without its guard.

## 2. The role lookup now throws where it used to answer null

Confirmed and fixed: `list_trades` propagates the query error, `get_trade_role` logged it and answered `Ok(None)`, and both call sites assume the latter.

It now catches, logs and answers `null`, keeping the old contract. Failing open is deliberate and documented: an unreadable store is no proof of participation, and a take the user really does hold is refused by the daemon, which the screen already reports through `_showTakeError`. A test with a throwing reader asserts the take is still dispatched rather than the button going dead.

## 3. The override re-implements the production provider

Agreed. The seam you describe already exists: `main` added `tradeListReaderProvider`, an injectable `listTrades`. `tradeRoleLookupProvider` now composes it, and the take-screen harness overrides **that**, so the tests exercise the real body — the rule and the error handling above — instead of a copy.

## 4. Deserialising the whole trade history, twice per take

The measurement is right: `get_trade_role` went through `idx_trades_order_id`, and this does not.

Some context on the size of it: the app already reads the full list routinely — `rawTradesProvider` for My Trades, `tradeInfoProvider` on the trade screen, and `_followTradeRow`'s polling — so this is not a new class of read, though the take tap is the one place where latency is felt.

I'd rather not add a bridge call in this PR: it means new Rust surface, regenerated bindings and a contract update, in a PR that is otherwise Dart-only and already carries a 296-commit merge. I'll open a follow-up for `list_trades_for_order` over the existing index, with `participatingRole` kept as the pure rule over its result. Happy to do it here instead if you prefer it not to wait.

grunch (approved):
All four points from the previous round are resolved on `f88ecdb`, and CI is green (Rust, Flutter, Web smoke, iOS).

**Verified**
- **Flash of `pending` + Cancel while the row loads** — `_status()` holds `loading` only on the *first* read (`isLoading && !hasValue`). That's better than what I suggested: `rawTradesProvider` is invalidated on every `TradeUpdate`, so gating on `isLoading` alone would flicker. The haptic listener returns early for the same reason. Both new widget tests fail without their guard.
- **Role lookup throwing** — `tradeRoleLookupProvider` catches, logs and returns `null`, which keeps the old `getTradeRole` contract. Failing open is the right call: the daemon still refuses a take the user already holds, and `_showTakeError` reports it. There's a test for the throwing reader.
- **Untested wiring** — the harness now overrides `tradeListReaderProvider`, so the take-screen tests run the real composition, error path included.
- **Full `listTrades` on the take tap** — fine as a follow-up (`list_trades_for_order` over `idx_trades_order_id`). Please open the issue.
- The merge with `main` kept both sides of the harness. The maker-bond branch of `_cancelOrder` is untouched, because `isTake` is `!order.isMine`.

**About CodeRabbit's "don't treat every terminal row as authoritative"**: I don't think it applies any more. On `main`, `apply_local_cancel` no longer writes an optimistic `Canceled` on an active trade: it records `cooperative_cancel_state = RequestedByMe` and leaves the status alone. A leftover `Canceled` row from an older build is also not frozen, because `wire_status_applies(Some(Canceled), Success)` is true (the wire status is terminal). So the row itself moves to `Success`, and `shownTradeStatus` shows that. No change needed.

**Nit (non-blocking)**: two texts still describe that optimistic write as current behaviour, and it's gone since the merge:
- the `shownTradeStatus` doc comment ("The one way such a row can be wrong is the cancel's optimistic write on an active trade…")
- the PR description's *Known trade-off* section and the first "Still open" bullet

It would help to reword them to "rows written by older builds", so nobody opens a follow-up for a write that no longer exists.

Approving.

PR #455, opened 2026-09-13
fix(home): the order book grid survives an order coming back

### LINKED ISSUE #453: Order book grid crashes when an order leaves the book and comes back
Image

With the order book in its grid layout (window 600 px or wider: 2 columns, 1200 px or wider: 3), the grid throws during layout as soon as an order that had left the book comes back. From then on it fails on every frame and the order area stays broken. In one session it logged 1,450 repeats, until the app was closed.

It happens to anyone watching the grid, not only to the user involved: an order leaves the book when someone takes it, and comes back when that take is cancelled or times out. The single-column list (phones, narrow windows) is not affected.

## Steps to reproduce

1. Run the app in debug with the window 600 px or wider, so the book shows the grid.
2. With several orders on screen, take one that is not the last card, and cancel it while it is still `waiting-buyer-invoice` / `waiting-payment`.
3. The order goes back to the book. The grid turns red, and the console shows:

```
══╡ EXCEPTION CAUGHT BY RENDERING LIBRARY ╞═════════════════════════
The following _TypeError was thrown during performLayout():
Null check operator used on a null value

The relevant error-causing widget was:
  SliverMasonryGrid:file:///…/lib/features/home/widgets/order_book_list.dart:90:19

When the exception was thrown, this was the stack:
#0      RenderSliverMasonryGrid.performLayout (package:flutter_staggered_grid_view/src/rendering/sliver_masonry_grid.dart:380:54)
…
  currently live children: 1 to 2
═════════════════════════════════════════════════════════════════════
Another exception was thrown: Null check operator used on a null value
Another exception was thrown: 'package:flutter/src/rendering/object.dart': Failed assertion: line 5493 pos 14: '!semantics.parentDataDirty': is not true.
… (repeated every frame)
```
## Cause

`OrderBookList` passes a `findChildIndexCallback` to its `SliverMasonryGrid` so that rows move instead of being rebuilt when the book re-sorts. The callback was added for the list in #357, and the grid reuses it since the order book redesign (#424). `flutter_staggered_grid_view`'s masonry grid does not handle children moved this way: after a card leaves and another is inserted before the cards on screen, `RenderSliverMasonryGrid.performLayout` reads a layout offset that was never set.

Reproduced in a widget test on `main`, with the book changing in place or while another route covers the home:

| Book change | 1 column (list) | 2 columns | 3 columns |
|---|---|---|---|
| A card in the middle leaves and comes back | ok | ❌ | ❌ |
| The first card leaves and comes back | ok | ❌ | ❌ |
| An older order is inserted in the middle | ok | ok | ❌ |
| A card leaves (nothing comes back) | ok | ok | ok |
| A new order arrives on top | ok | ok | ok |

The same grid **without** the index callback passes every row of the table. `flutter_staggered_grid_view` 0.7.0 (July 2023) is the latest release, so there is no upstream fix to update to.

## Proposed fix

Drop the index callback in the grid, and the `ValueKey` on its cards: as `OrderBookList` itself notes, a key without the callback is worse than no key. The single-column list keeps both.

The cost is that, in the grid only, cards shifted by an arriving order or a re-sort are rebuilt instead of moved, which undoes #357's optimisation there.

Tests:
- `order_book_list_reorder_test.dart`: the "keeps a row with its order" case covers the list only;
- a new regression test: with 2 and 3 columns, an order that leaves the book and comes back lays out without an exception.

Rejected alternatives:
- **A fixed-extent grid:** it overflows on long localized copy and large text, which is why the redesign review moved to masonry.
- **Rebuilding the grid whenever the set of orders changes:** a workaround that depends on exactly when the package breaks.

Catrya:
Closes #453 

With the order book in its grid layout (window 600 px or wider), every order that left the book and came back made `SliverMasonryGrid` throw in `performLayout` (`Null check operator used on a null value`) on every frame. The grid went red in debug and blank for the user for a few seconds, until a later update laid it out again. An order leaves and comes back whenever a take is cancelled or times out, so anyone watching the grid saw it, not only the user involved.

## Cause

`OrderBookList` passed the list's `findChildIndexCallback` (#357) to its masonry grid (#424), so that cards moved instead of being rebuilt. `flutter_staggered_grid_view` 0.7.0 (its latest release, July 2023) cannot lay out a child moved that way.

## Change

- **Grid** (2 or 3 columns): no index callback, and no `ValueKey` on its cards. As `OrderBookList` notes, a key without the callback is worse than none.
- **List** (1 column): unchanged. It keeps both, and `indexOfKey` now lives in its branch.

The cost: in the grid, a card shifted by an arriving order or a re-sort is rebuilt in place instead of moved.

## Tests

- `order_book_list_reorder_test.dart`: the "keeps a row with its order across a re-sort" case now covers the list only.
- New regression group: with 2 and 3 columns, the grid goes through an order in the middle leaving and coming back, the first order leaving and coming back, and an older order arriving in the middle. It lays out without an exception every time.
- With the callback restored in the grid, 5 of the 6 new cases fail. The sixth, 2 columns with an older order arriving in the middle, did not fail on `main` either: that case only breaks with 3 columns.
- `flutter analyze` is clean and `flutter test` passes (1029).

## Manual test

With a wide window, took and cancelled orders so each came back to the book, three times. No exceptions and no blank grid; before this change each of those threw on every frame.


lib/features/home/widgets/order_book_list.dart
```diff
@@ -94,21 +103,21 @@ class OrderBookList extends StatelessWidget {
             mainAxisSpacing: _gap,
             crossAxisSpacing: _gap,
             delegate: SliverChildBuilderDelegate(
-              (context, index) => _card(orders[index]),
+              (context, index) => _card(orders[index], keyed: false),
```
grunch, line 106: _🎯 Functional Correctness_ | _🟠 Major_

**Unkeyed grid cards let a press open the wrong order when the book shifts mid-gesture.**

Without a key, the element at grid index *i* is reused for whatever order lands at *i* after an update. Its `InkWell` keeps the in-progress gesture, and on release it calls the *new* widget's `onTap`, which is `() => onOrderTap(<the order now at i>)`. Book updates arrive live from relays, and the grid is exactly the layout used on desktop, web and tablets, where press-and-hold (and a slow click) is common.

Reproduced on this commit: press `order-20` (index 1), a newer order arrives and every card shifts one slot, release → **`onOrderTap('order-10')`**. The user lands on an order they never selected, possibly with a Take button in front of them. The same reuse moves hover and keyboard focus to a different order on web and desktop.

The PR removes the key on the premise that "without the callback a key would only make that worse". That holds for rebuild cost, not correctness. A keyed card whose key no longer matches the widget at its index fails `Widget.canUpdate`, so the element is torn down, which cancels the gesture. It does **not** reintroduce the crash, because nothing is moved by key. With `keyed: true` here and still no `findChildIndexCallback`:

- the same probe records **no tap** (the press is cancelled);
- all 6 new "grid lays out an order that comes back" tests still pass.

Suggested fix: keep the key in the grid, not the callback, and reword the comment above so the next reader doesn't "optimise" it away again.

```diff
-              (context, index) => _card(orders[index], keyed: false),
+              (context, index) => _card(orders[index], keyed: true),
```

```dart
    // Unlike the list, the grid does not move cards by key:
    // `SliverMasonryGrid` cannot lay out a child moved by
    // `findChildIndexCallback` (#453). Its cards stay keyed anyway. A card
    // whose order moved is torn down and re-inflated rather than reused for
    // another order, so a press, hover or focus in progress never lands on
    // an order the user did not pick. The cost is a re-inflate instead of a
    // rebuild for each shifted card.
```

<details>

```
In lib/features/home/widgets/order_book_list.dart, the SliverMasonryGrid builder (around line 106):
pass keyed: true to _card so grid cards keep ValueKey(order.id), while still NOT passing
findChildIndexCallback to the grid's SliverChildBuilderDelegate. Update the comment block above
the grid (around lines 81-89) to explain the key is kept for gesture/focus safety and that only
the index callback is omitted (#453). Then add the regression test described on
test/features/home/order_book_list_reorder_test.dart.
```

</details>

lib/features/home/widgets/order_book_list.dart
```diff
@@ -75,14 +75,23 @@ class OrderBookList extends StatelessWidget {
...
     }
 
     // Masonry, not a fixed-extent grid: a card's height depends on its
     // content and the text scale (wrapping chips, two lines of payment
     // methods), so a fixed tile ratio overflows on long localized copy or
-    // large text. The same index callback keeps rows moving on a re-sort.
+    // large text.
+    //
+    // Unlike the list, the grid neither keys its cards nor moves them by key:
+    // `SliverMasonryGrid` (flutter_staggered_grid_view 0.7.0, its latest
```
grunch, line 88: _📐 Maintainability_ | _🔵 Trivial_

**This comment will rot, and the dependency isn't pinned to the version it describes.**

"flutter_staggered_grid_view 0.7.0, its latest release" is true today, but `pubspec.yaml` declares `^0.7.0`, so a `0.7.x` release that changes `RenderSliverMasonryGrid` would land on the next `pub upgrade` with nothing prompting anyone to revisit this workaround.

Consider referencing the issue instead of a point-in-time claim (`… cannot lay out a child moved by findChildIndexCallback (#453, flutter_staggered_grid_view 0.7.0)`). Either pin the dependency exactly, or leave a `TODO(#453)` to re-enable the callback in the grid once upstream handles moved children. The regression group below would then show whether a future version fixed it.

test/features/home/order_book_list_reorder_test.dart
```diff
@@ -93,6 +92,46 @@ void main() {
...
+      for (final MapEntry(key: name, value: books) in changes.entries) {
+        testWidgets('with $columns columns: $name', (tester) async {
+          await withClock(Clock.fixed(kFakeNow), () async {
+            for (final book in books) {
+              await _pump(tester, _book(book), columns: columns);
+              await tester.pump();
+              expect(tester.takeException(), isNull, reason: 'book $book');
+            }
+
+            expect(find.byType(OrderListItem), findsNWidgets(4));
```
grunch, line 128: _🧪 Test Coverage_ | _🟡 Minor_

**The new group proves "no exception", but nothing guards what the user taps after a shift.**

Every case asserts `takeException() == null` and four cards. That is the right regression for #453, but it is equally green whether grid cards are keyed or not, so the wrong-order press from the other comment isn't covered, and would silently come back if someone later drops the key "for performance".

Suggested addition next to this group (verified: it fails on this commit and passes with `keyed: true`):

```dart
testWidgets('a press that spans a book shift never opens another order', (
  tester,
) async {
  tester.view.physicalSize = const Size(1000, 2400);
  tester.view.devicePixelRatio = 1.0;
  addTearDown(tester.view.reset);
  await withClock(Clock.fixed(kFakeNow), () async {
    final taps = <String>[];
    Future<void> pumpWith(List<int> book) => tester.pumpWidget(
      MaterialApp(
        theme: buildDarkTheme(),
        localizationsDelegates: AppLocalizations.localizationsDelegates,
        supportedLocales: AppLocalizations.supportedLocales,
        home: Scaffold(
          body: OrderBookList(
            orders: _book(book),
            currencyFlags: const {'USD': '🇺🇸'},
            reasons: const {},
            columns: 2,
            onOrderTap: taps.add,
          ),
        ),
      ),
    );

    await pumpWith([10, 20, 30, 40]);
    final gesture = await tester.startGesture(
      tester.getCenter(find.byType(OrderListItem).at(1)), // order-20
    );
    await tester.pump(const Duration(milliseconds: 50));
    await pumpWith([5, 10, 20, 30, 40]); // every card shifts one slot
    await gesture.up();
    await tester.pumpAndSettle();

    expect(taps, isNot(contains('order-10')));
  });
});
```

grunch (changes requested):
**Actionable comments posted: 3**

Root cause and fix direction are right: the crash is `childScrollOffset(child)!` in `RenderSliverMasonryGrid.performLayout` (`flutter_staggered_grid_view-0.7.0/lib/src/rendering/sliver_masonry_grid.dart:380`), reached once `findChildIndexCallback` moves a child the masonry layout never positioned, and dropping the callback in the grid removes it. The single-column path is unchanged. One part of the change is not neutral, though: **removing the `ValueKey` from grid cards lets an in-flight press open a different order than the one pressed.** Verified below, and the fix is a one-word change the PR's own regression tests already accept.

<details>
<summary>🧹 Verified-OK notes (4)</summary><blockquote>

- **Crash site** matches #453's trace (`sliver_masonry_grid.dart:380:54`, the `childScrollOffset(child)!` in the earliest-scroll-offset scan). ✅
- **No per-order state can bleed on an in-place rebuild:** `OrderListItem` and every sub-widget in `order_list_item.dart` (`_HeaderRow`, `_CurrencyChip`, `_Chip`, `_AmountRow`, `_FixedSatsCaption`, `_PaymentMethodsRow`, `_ReputationStrip`, `_Stat`) are `StatelessWidget`s, so the only state that crosses a shift is the `InkWell`'s gesture/splash — see the inline comment. ✅
- **List path** (`columns == 1`) keeps `ValueKey` + the doubled index callback; the "keeps a row with its order across a re-sort" test still guards it. ✅
- **CI** green on the merge ref (Flutter, Rust, Web smoke); `mergeable: MERGEABLE`. ✅

</blockquote></details>

<details>

**Reviewed commit:** `7255ac6`
**Files reviewed:** 2 — `lib/features/home/widgets/order_book_list.dart`, `test/features/home/order_book_list_reorder_test.dart`
**Profile:** strict

**How the finding below was verified** (throwaway worktree on `7255ac6`, not pushed): a probe widget test pumps the grid (2 columns) with orders `[10, 20, 30, 40]`, starts a press on the second card (`order-20`), pumps a book where a newer order arrives on top (`[5, 10, 20, 30, 40]`), releases, and records `onOrderTap`.

| Grid cards | Index callback | `onOrderTap` after the press | PR's 6 "grid lays out an order that comes back" tests |
|---|---|---|---|
| unkeyed (this PR) | none | **`order-10`** — a different order than the one pressed | pass |
| `ValueKey(order.id)` | none | **none** — the press is cancelled with the torn-down element | **pass** |

</details>

grunch (approved):
Strict re-review on `7d260b2`. The three points from the previous round are resolved, and I checked the fix locally, not just by reading it.

**Verified** (throwaway worktree: this branch merged with current `main` at `7e829b1`)
- `flutter test test/features/home/`: 96/96 pass. CI on the merge ref is green (Flutter, Rust, Web smoke, iOS).
- **Mutation A**, restoring `findChildIndexCallback` in the grid: **5 of the 6** "grid lays out an order that comes back" cases fail, matching the description. The 2-column "older order comes back in the middle" case doesn't tell the two apart, as stated. It's fine to keep it as coverage.
- **Mutation B**, unkeying the grid cards: *"a press that spans a book shift never opens another order"* fails, so the key is now guarded against a later "optimisation".
- **Mechanism**: with no index callback, `SliverMultiBoxAdaptorElement.performRebuild` keeps each child at its old index, and a key mismatch fails `Widget.canUpdate`. So the element is torn down, never moved, which is why the key can't bring the masonry crash back.
- **Rot**: the comment now cites `#453` + `0.7.0`, and `TODO(#453)` points at the regression group that will say when upstream is fixed. That's enough without pinning.
- Nothing on `main` since the branch's base (`68ab52b`) touches `order_book_list.dart`, `order_list_item.dart` or the test. Only `pubspec.yaml` version bumps, and the package is untouched.

**Before merging (please)**
- **The PR description contradicts the code since `7d260b2`.** *Change* still says "no index callback, **and no `ValueKey` on its cards**. As `OrderBookList` notes, a key without the callback is worse than none", and *The cost* says a shifted card "is rebuilt in place". Both are now false: the grid keeps the key, the class comment argues the opposite, and a shifted card is **re-inflated**. Anyone reading the PR later to learn why the grid is keyed will get the wrong answer. Please update the description to match the second commit.

**Nits (non-blocking)**
- `order_book_list.dart:10-14`: the class doc was re-wrapped with a hard break after "builds." (`/// builds.` / `/// Inline in the screen …`). Dartdoc joins them into one paragraph anyway, so either reflow it or add a blank `///`. "the grid's keys without one" reads awkwardly too; "the grid's keyed cards without the index callback" is clearer.
- `order_book_list_reorder_test.dart`: the press test sits in the group named "grid lays out an order that comes back", but it's about gesture safety, not layout. A separate group (e.g. "OrderBookList grid keeps a press with its order") would make a failure easier to read.
- **Keyboard focus, as an acknowledged cost**: a desktop or web user with keyboard focus on a grid card loses it whenever an order arrives above that card. On `main` (with the callback) it stayed with the order. That's the right trade over focus silently landing on another order, and much better than a crashing grid. It's worth one clause in the comment ("focus is dropped, not kept") so the TODO's value is clear.

Approving: the code is correct and well guarded. Please fix the description before merging.

PR #456, opened 2026-09-13
fix(order): the take screen keeps Taking… while its own take settles

### LINKED ISSUE #454: Take screen says "No longer available" while the user's own take is being confirmed
After tapping **Take order**, the button switches from *Taking…* to **"No longer available"** for a few seconds before the app moves on to the trade (and the add-invoice or pay-invoice screen). The take is working fine; the message is simply wrong, and it tells the user the order is gone right when they got it.

## Steps to reproduce

1. Open a pending order from the book.
2. Tap **Take order**.
3. The button briefly shows *Taking…*, then **No longer available**. A few seconds later the app navigates to the trade.

The Rust side confirms the take within the same second:

```
orders: take_order published order=a40d8018-… trade_index=85 — waiting for daemon
orders: take_order confirmed by daemon: order=a40d8018-… reply=AddInvoice
session: installed order=a40d8018 idx=85 role=Buyer
```

## Cause

Two things combine:

1. **`take_order` updates the book before it returns.** As soon as the daemon confirms, it updates the order's book entry with the trade's status (e.g. `WaitingBuyerInvoice`) "so the order doesn't linger as Pending". It then persists the trade, subscribes to the order's d-tag and installs the session, and only after that does it return to the screen.
2. **The take screen reads that change as the order becoming unavailable.** In `take_order_screen.dart`:

   ```dart
   final isUnavailable =
       _cta == TakeOrderCta.unavailable ||
       live == null ||
       live.status != OrderStatus.pending;
   ```

   Nothing here accounts for a take in flight. The `ref.listen` just above it does: it never marks the order unavailable while `_cta == TakeOrderCta.loading`. But this expression in `build` overrides that: once the book reflects the user's own take, the button shows *No longer available* until `take_order` returns and the screen navigates (after `refreshTrades` and, for a buyer, `getSettings`).

The order is fixed, so this happens on every take where the book update reaches the screen before `take_order` returns. It was introduced with the take screen redesign (35c1aaf).

## Proposed fix

While a take is in flight, keep the button on *Taking…* whatever the book says:

```dart
final isUnavailable =
    _cta == TakeOrderCta.unavailable ||
    (_cta != TakeOrderCta.loading &&
        (live == null || live.status != OrderStatus.pending));
```

Every outcome of the take still ends right:

Every outcome of the take still ends right:

- **Success:** the screen navigates to the trade without ever showing *No longer available*.
- **`OrderAlreadyTaken`:** `_showTakeError` sets `unavailable`, as today.
- **Any other error:** `_cta` goes back to `idle`, and if the book then says the order is no longer pending, *No longer available* shows, which is true in that case.

Test: a widget test where the book moves the order to `waitingBuyerInvoice` while the take is still pending, asserting that the button still reads *Taking…*.

Catrya:
Closes #454.

After tapping **Take order**, the button switched from *Taking…* to **"No longer available"** for a few seconds before the app moved on to the trade. The take itself was fine, but the screen told the user the order was gone right when they got it.

## Cause

`take_order` (Rust) updates the order's book entry with the trade's status as soon as the daemon confirms the take. It then persists the trade, subscribes to the order and installs the session, and only after that does it return. The take screen's `build` read that book change as the order leaving the pending book:

```dart
final isUnavailable =
    _cta == TakeOrderCta.unavailable ||
    live == null ||
    live.status != OrderStatus.pending;
```

Its own `ref.listen` just above already held back while a take was in flight (`_cta != TakeOrderCta.loading`), but this expression overrode it. The behaviour came with the take screen redesign (35c1aaf).

## Change

While the take is loading, the button stays on *Taking…* whatever the book says:

```dart
final isUnavailable =
    _cta == TakeOrderCta.unavailable ||
    (_cta != TakeOrderCta.loading &&
        (live == null || live.status != OrderStatus.pending));
```

Once the take settles, nothing else changes:

- **Success:** the screen navigates to the trade (or the bond, add-invoice or pay-invoice step), as before.
- **`OrderAlreadyTaken`:** marks the order unavailable, as before.
- **Any other failure:** lands back on `idle`, where a book that has moved on shows *No longer available*, which is true then.

A take that parks at `WaitingTakerBond` never hit this: Rust leaves its book entry `pending` during the bond window.

## Tests

In `take_order_screen_test.dart`, `TakeOrderScreen taking` group:

- **Keeps *Taking…* while its own take moves the order on:** the book moves the order to `waitingBuyerInvoice` while the take is pending, and the button still reads *Taking…*. This test fails without the change.
- **A failed take over an order that moved on reads unavailable:** the order goes `in-progress`, then the take fails with `NoDaemonResponse`, and the screen shows *No longer available*. This pins the behaviour kept after a failure.

`flutter analyze` is clean and `flutter test` passes (1114).

## Manual test

Took two orders against a local mostrod. Each went straight from *Taking…* to the trade, with no *No longer available* in between.

## Notes

- The order book grid still throws when a cancelled order comes back to the book (#453). That was seen during the manual test and is fixed separately in #455, which this branch does not include.
- #448 also edits `take_order_screen_test.dart` (the `_pump` harness and a new group at the end of the file). The tests here are added inside an existing group to keep the two apart; whichever lands second may still need a small merge.


lib/features/order/screens/take_order_screen.dart
```diff
@@ -282,10 +282,17 @@ class _TakeOrderScreenState extends ConsumerState<TakeOrderScreen> {
...
+    // the daemon confirms, then persists the trade and subscribes, and only
+    // then does `take_order` return). Read as the order going away, that
+    // showed "No longer available" for the order the user had just got
+    // (#454). The listener above holds back for the same reason; once the
+    // take settles, a failure lands back on `idle` and the book decides.
     final isUnavailable =
         _cta == TakeOrderCta.unavailable ||
-        live == null ||
-        live.status != OrderStatus.pending;
+        (_cta != TakeOrderCta.loading &&
```
grunch, line 294: _🎯 Functional Correctness_ | _🟠 Major_

**A successful take still shows "No longer available", now during the navigation instead of before it.**

The hold is keyed on `_cta == TakeOrderCta.loading`, but `_onTakeOrder`'s `finally` still runs after the navigation:

```dart
      context.go(AppRoute.tradeDetailPath(widget.orderId));
      context.push(AppRoute.payInvoicePath(widget.orderId));
    } catch (e) { … } finally {
      if (mounted && _cta == TakeOrderCta.loading) {
        setState(() => _cta = TakeOrderCta.idle);   // still mounted: the route is animating out
      }
    }
```

`context.go` doesn't unmount this screen synchronously; it stays in the tree while the next route animates in. The `finally` puts `_cta` back on `idle`, so this expression lets the book decide again. The book already reads `waitingPayment`/`waitingBuyerInvoice` from the user's own take, so the outgoing page renders **No longer available**, and the countdown switches to closed as well since `isClosed: isUnavailable`.

Reproduced on this commit: the book moves to `waitingPayment`, the take then succeeds, and **19 consecutive 16 ms frames** show *No longer available* with `TakeOrderScreen` still mounted before the app lands on the pay step. That is #454's symptom, shortened but not gone. The manual test in the description can easily miss it under the transition.

Suggested fix: once the take succeeded and the screen is navigating away, don't fall back to `idle`.

```diff
     setState(() => _cta = TakeOrderCta.loading);
+    var navigated = false;
     try {
       final trade = await ref.read(takeOrderActionProvider)(…);
       if (!mounted) return;
       …
       if (trade.order.status == OrderStatus.waitingTakerBond) {
+        navigated = true;
         context.go(AppRoute.tradeDetailPath(widget.orderId));
         context.push(AppRoute.payBondPath(widget.orderId));
       } else if (widget.isBuying) {
         final settings = await settings_api.getSettings();
         if (!mounted) return;
+        navigated = true;
         context.go(AppRoute.tradeDetailPath(widget.orderId));
         …
       } else {
+        navigated = true;
         context.go(AppRoute.tradeDetailPath(widget.orderId));
         context.push(AppRoute.payInvoicePath(widget.orderId));
       }
     } catch (e) { … } finally {
-      if (mounted && _cta == TakeOrderCta.loading) {
+      if (mounted && !navigated && _cta == TakeOrderCta.loading) {
         setState(() => _cta = TakeOrderCta.idle);
       }
     }
```

Set the flag **after** the buyer branch's `await settings_api.getSettings()`, as above. If that call throws, the screen never navigates and must still return to `idle`, or the button is stuck on *Taking…*. The variant I ran set the flag once before the branches; the success/expiry probes and all 19 tests in this file pass with it, and moving it past the only `await` in the buyer branch doesn't change the seller or bond paths they exercise.

<details>

```
In lib/features/order/screens/take_order_screen.dart, _onTakeOrder: add `var navigated = false;`
before the try; set `navigated = true;` immediately before the first context.go(...) of each
success branch (in the buyer branch, after `await settings_api.getSettings()` and its mounted
check); in `finally`, only reset `_cta` to idle when `!navigated`. Add a routed widget test where
the book moves the order to waitingPayment, the take then succeeds (seller path), and assert that
"No longer available" is never found while TakeOrderScreen is still in the tree.
```

</details>

lib/features/order/screens/take_order_screen.dart
```diff
@@ -282,10 +282,17 @@ class _TakeOrderScreenState extends ConsumerState<TakeOrderScreen> {
...
+            (live == null || live.status != OrderStatus.pending));
     final cta = isUnavailable ? TakeOrderCta.unavailable : _cta;
```
grunch, line 296: _🎯 Functional Correctness_ | _🟡 Minor_

**The countdown can still override an in-flight take.**

This line trusts `_cta == TakeOrderCta.unavailable`, and `_syncCountdown` sets exactly that when `expiresAt` passes, without checking for a take in flight:

```dart
    if (left <= Duration.zero) {
      _remaining.value = Duration.zero;
      if (_cta != TakeOrderCta.unavailable) {          // true while loading
        setState(() => _cta = TakeOrderCta.unavailable);
      }
```

An order whose expiry falls inside the few seconds `take_order` takes flips from *Taking…* to **No longer available** mid-take (reproduced on this commit). The `finally` then leaves it there, since `_cta` is no longer `loading`. If the daemon accepted the take, the user is told the order is gone right before being navigated to it: the case this PR sets out to remove, on another path.

Suggested fix: the countdown only retires an idle button; a take in flight is settled by the daemon's answer.

```diff
-      if (_cta != TakeOrderCta.unavailable) {
+      if (_cta == TakeOrderCta.idle) {
         setState(() => _cta = TakeOrderCta.unavailable);
       }
```

Verified: with this change the expiry probe keeps *Taking…*, and the existing expiry tests still pass. After a failed take, `finally` returns to `idle` and this expression reads the expired book entry, so the button still dies in place.

test/features/order/screens/take_order_screen_test.dart
```diff
@@ -273,6 +273,59 @@ void main() {
...
+        await tester.tap(find.text('Take order'));
+        await tester.pump();
+        books.add([
+          fakeOrder(id: _id, status: OrderStatus.waitingBuyerInvoice),
+        ]);
+        await tester.pump();
+        await tester.pump();
+
+        expect(find.text('Taking…'), findsOneWidget);
+        expect(find.text('No longer available'), findsNothing);
```
grunch, line 299: _🧪 Test Coverage_ | _🟡 Minor_

**Neither new test lets the take *succeed* after the book moved, which is where the symptom still shows.**

The first test leaves the take unanswered, so it only proves the hold while `loading`. The second covers the failure path. The success path, with the book already moved, the take resolving and `finally` running while the screen is still mounted, is untested, and so is expiry during the take. Both paths still regress (see the comments above).

Suggested additions (both fail on this commit and pass with the suggested fixes):

```dart
testWidgets('a take that succeeds never reads unavailable on the way out', (
  tester,
) async {
  // Seller path under a router, so the navigation and the exit frames are real.
  // Book → waitingPayment, then the take resolves successfully.
  // Pump ~60 × 16 ms and fail if "No longer available" is found while
  // `find.byType(TakeOrderScreen)` is still non-empty; end on the pay step.
});

testWidgets('an order expiring mid-take keeps Taking…', (tester) async {
  var now = kFakeNow;
  await withClock(Clock(() => now), () async {
    // expiresAt: kFakeNow + 2 s; tap Take with the reply left pending;
    // advance `now` past expiry and pump 3 s; expect 'Taking…', not
    // 'No longer available'.
  });
});
```

The first needs the router harness #448 introduces, or a local one. Coordinate with that PR to avoid a second harness.

grunch (changes requested):
**Actionable comments posted: 3**

The diagnosis is right and the new rule is the right one: `take_order` writes the daemon's status into the book entry before `persist_confirmed_take`, `subscribe_single_order` and `install_session`, so the in-flight take was reading its own book write as the order going away. The rule is not enforced on every path, though. **The same "No longer available" still reaches the user after a successful take and when the order expires mid-take.** Both are verified below, with a fix that keeps all 19 tests in this file green.

<details>
<summary>🧹 Verified-OK notes (4)</summary><blockquote>

- **Rust claims in the description** — confirmed on `2724772`: the book entry takes the reply's status (`order_book().upsert_order(info)`) before persistence, the per-order subscription and the session install; and `!waiting_bond` skips that write, so a take parked at `WaitingTakerBond` keeps its entry `pending`. ✅
- **Failure path** — `finally` returns `_cta` to `idle` after a non-`OrderAlreadyTaken` failure, and the book decides again; the new second test pins it. ✅
- **CI** green on the merge ref (Flutter, Rust, Web smoke); `mergeable: MERGEABLE`; no commit on `main` since this branch's base touches either file. ✅
- **Overlap with #448** — both edit `take_order_screen_test.dart`. The tests here sit inside an existing group, as the description says, but #448 replaces `_pump`'s `MaterialApp` with a router harness, so expect a small conflict in the harness whichever lands second.

</blockquote></details>

<details>

**Reviewed commit:** `2724772`
**Files reviewed:** 2 — `lib/features/order/screens/take_order_screen.dart`, `test/features/order/screens/take_order_screen_test.dart`
**Profile:** strict

**How the findings were verified** (throwaway worktree on `2724772`, not pushed). A probe test pumps `TakeOrderScreen` under a `GoRouter`, as a seller so the success path makes no Rust settings call. The book is fed the way the real flow feeds it.

| Probe | This PR | With the suggested fix |
|---|---|---|
| Take succeeds after the book moved to `waitingPayment`: frames showing *No longer available* while `TakeOrderScreen` is still mounted | **19 frames (~300 ms)**, then lands on the pay step | **0 frames**, lands on the pay step |
| Order's `expiresAt` passes while the take is in flight | *Taking…* → **No longer available** | stays *Taking…* |
| `take_order_screen_test.dart` (19 tests, including this PR's two) | pass | **pass** |

</details>

Catrya:
@grunch All three taken, and both probes reproduced here first. One of them turned up a case the suggested fix leaves open, so that part landed slightly differently — details below.

## 1. A successful take still shows "No longer available"

Reproduced exactly: seller path, book moved to `waitingPayment`, take resolves.

| | Frames showing "No longer available" with `TakeOrderScreen` still mounted |
|---|---|
| previous commit | **19** |
| with the `navigated` flag | **0** |

Fixed as suggested, with the flag set past the buyer branch's `await settings_api.getSettings()` for the reason you gave: a settings read that throws leaves the screen here, and the button has to come back rather than stay on *Taking…*.

## 2. The countdown can override an in-flight take

Also reproduced: an order expiring inside the take went from *Taking…* to *No longer available*.

Your fix is in — the countdown only retires an idle button. But I checked the part that says a failed take then reads the expired book entry, and on this screen it does not:

| After an expiry mid-take, when the take fails | previous commit | countdown fix alone | what shipped |
|---|---|---|---|
| "Take order" (live again) | 0 | **1** | 0 |
| "No longer available" | 1 | 0 | 1 |

The book still calls the order `pending` until the daemon publishes its end, so with the countdown holding its fire the expiry was simply lost: `_cta` went back to idle and the button came alive on an order whose clock had run out, countdown at 00:00. The timer will not fire again either, since it returns without rescheduling.

So the expiry is now remembered in an `_expired` marker, kept apart from `_cta`, and applied by `_onTakeOrder`'s `finally` when the take fails. The three paths end right: success navigates, a failure before expiry returns to idle, a failure after expiry dies in place.

## 3. The tests don't cover those paths

Added three, each failing without its own guard:

- a successful take that never reads unavailable while the screen is mounted (60 frames, ending on the pay step);
- an order expiring mid-take that keeps *Taking…*;
- an order that expired mid-take whose take then fails, where the button must not come back to life.

On the harness: when I wrote them #448 had not landed, so the success test builds its own router inline rather than adding a second `_pump`. #448 is in `main` now and merged here, so that test can move to the shared routed harness — say the word and I'll fold it in before you look again.

## Merge

Merged `main` (289 commits, no conflicts, #448 included). `flutter analyze` clean, `flutter test` green (1557). Note for anyone picking this branch up: it predates generated code being committed, so before the merge it did not build without regenerating first.

lib/features/order/screens/take_order_screen.dart
```diff
@@ -172,27 +188,37 @@ class _TakeOrderScreenState extends ConsumerState<TakeOrderScreen> {
...
         context.go(AppRoute.tradeDetailPath(widget.orderId));
         context.push(AppRoute.payBondPath(widget.orderId));
       } else if (widget.isBuying) {
         // With a default LN address Mostro pays it directly and the buyer
         // skips the add-invoice step.
         final settings = await settings_api.getSettings();
         if (!mounted) return;
+        // Set past that await: a settings read that throws leaves the screen
+        // here, and the button has to come back rather than stay on Taking…
+        navigated = true;
```
grunch: By this point `take_order` has returned, so the trade exists. If `getSettings()` throws, the error goes to the `catch` and `_showTakeError` shows it as a take failure, although that method's own doc says an error there means the trade was **not** created. The `finally` then sets `idle`, and since the book already moved the order out of `pending`, the user sees the error plus *No longer available* for an order they actually took, and stays on this screen instead of their trade.

Suggestion: don't let a settings failure undo a successful take. For example, catch around `getSettings()` and fall back to pushing the add-invoice step (the no-default-address path), so the user always lands on the trade.

grunch (commented):
The core change is right: holding `isUnavailable` off while `_cta == loading` matches what the `ref.listen` above already does, and it fixes #454. `_expired` and the `finally` cover the expiry cases, and the new tests exercise the real race between the take and the book update. Nothing here blocks the merge; three things are worth a look.

**1. A settings read that throws reports a successful take as a failure** (inline below).

**2. The description no longer matches the diff.** It says *"Once the take settles, nothing else changes"* and lists two tests. The diff also adds `_expired` (an expiry that falls inside a take no longer kills it and is applied only if the take fails), adds `navigated` (no *No longer available* while the screen animates out after a success), and has five tests. Worth updating so the change in expiry handling doesn't go unnoticed.

**3. Two paths the PR handles on purpose have no test:**
- the settings read that throws (point 1);
- the "never reads unavailable on the way out" test covers only the seller path; the buyer path has the extra `getSettings` await before navigating.

**Pre-existing, not introduced here:** `_onTakeOrder` checks `_cta != idle` at the top but only sets `loading` after two awaits (the role lookup and, for range orders, the amount modal). Two quick taps can both pass the guard and dispatch two takes. Probably worth its own issue.

grunch (approved):
Approving at `b46f46c`. Everything from the last two rounds is in:

- **A settings read that throws after a successful take** now has its own `try/catch`, so it never reaches `_showTakeError`. `navigated` is set before the `await`, and the buyer lands on the trade screen without the add-invoice push. That is right: once `take_order` has returned, the trade exists.
- **The description** now matches the diff: `_expired`, `navigated`, the unreadable setting, and six tests.
- **Coverage:** the success test runs for both seller and buyer, and the buyer case goes through the settings-read-throws path (no SnackBar, lands on the trade). Expiry is covered both mid-take and after a failed take.

**Checked against current `main`:** `4ad29a0` (the `_taking` flag, #551) touches the same file and merges cleanly with this branch. The combined code is consistent: `_taking` wraps `_takeOrder`, and `_takeOrder`'s `finally` still owns `_cta`. On the merged tree, `flutter analyze` is clean and `take_order_screen_test.dart` passes (28/28).

**Low, not blocking:** the buyer case covers the unreadable-setting path only because the Rust bridge throws in a widget test. If `settings_api` ever gets a fake in that harness, the test still passes but no longer covers that path. A one-line comment on the assertion, or a dedicated fake that throws, would make that dependency explicit.

PR #461, opened 2026-09-13
feat(bond): PR-5 — web store smoke, push gap, docs and dead l10n

### LINKED ISSUE #145: [Epic] Anti-abuse bond — full protocol support
**Milestone:** 3 — Protocol gaps
**Type:** protocol / epic
**Priority:** high

v1 implements the anti-abuse bond protocol across multiple phases (Phase 1.5 / 3.5 / 5 in the v1 CHANGELOG). This epic ports full client-side bond support to v2.

This is an **epic** — tracked via sub-issues, not landed as one PR.

## Starting point (verified 2026-07-18)

The original "zero bond support, no symbols" framing is out of date. Current v2 state:

- v2 **recognizes and cleanly rejects** bond-requiring nodes: `Action::PayBondInvoice` -> stable `BondRequired` marker (`rust/src/api/orders.rs:261-272`), localized in all 5 locales, with a unit test. Part of this rollout is *retiring* that rejection.
- `mostro-core 0.13.1` (already pinned) ships the full bond vocabulary — `AddBondInvoice`, `BondInvoiceAccepted`, `BondPayoutCompleted`, `BondSlashed`, payloads `BondResolution` / `BondPayoutRequest`, statuses `WaitingTakerBond` / `WaitingMakerBond`. **No crate upgrade is needed.**
- Kind 38385 is parsed (`lib/features/about/models/mostro_instance.dart:75-112`) but has **no bond-policy field** yet.
- `WaitingTakerBond | WaitingMakerBond` currently map to `None` (`rust/src/api/orders.rs:2069-2070`, "out of scope") — they need local representation.

Priority is **high** if nodes we care about will require bond (it blocks trading against them); **medium** if no production node requires it yet.

## Sub-issues

- #190 — Parse + validate bond policy from kind 38385; sanitize malformed tags.
- #191 — Phase 5: maker anti-abuse bond on order creation.
- #192 — Phase 1.5: recognize `add-bond-invoice`; publish the payout invoice on receipt.
- #193 — Phase 3.5: bond payout phase model + confirmation acks (`bond-invoice-accepted`, `bond-payout-completed`).
- #194 — Recognize `bond-slashed`; never overwrite the tracked order's real amount.
- #195 — UI: payout banner (Trade Detail) + PAYOUT PENDING badge (My Trades) + bond-slashed detail dialog.
- #196 — About screen: display the active node's bond policy.
- #197 — Session lifecycle: defer session deletion ~60s on cancel; cancel the defer on `bond-slashed`.

## Not yet scoped

The **taker pay-bond flow** (Flow 1 in the v1 doc / v1 PR #592 — recognize `pay-bond-invoice`, add a `WaitingTakerBond` representation, and retire the `BondRequired` rejection) is not covered by the sub-issues above. It is the entry point for the taker experience and where the current rejection is removed. Open it as a follow-up before the payout/slash flows can be exercised end-to-end against a take.

This epic blocks trading against any Mostro node that requires bond.

## Refs
- v1 bond PRs: #608 (maker, Phase 5), #592 (taker).
- v1 docs: `MostroP2P/mobile` `docs/architecture/ANTI_ABUSE_BOND.md` (client-side parity spec).
- Design mocks: PR #183 (`docs/design/145-*.png`).

grunch:
## Summary

Phase 5 of `docs/ANTI_ABUSE_BOND.md` (T5.1 + T5.2 + T5.3): test plumbing and docs, no protocol change. Closes #145.

### T5.1 — web smoke: bond rows through the real store
- The bridge probe proved a Rust call returned. It said nothing about IndexedDB. On web, bond rows reach the UI only after a serde decode in the wasm core and an FRB decode in Dart, and a build that breaks either one silently shows an empty My Trades.
- `smoke.mjs` gains an opt-in step, enabled only in the CI release-bundle run via `SMOKE_BOND_STORE=1`. It seeds a `Pending` claim plus trades at `WaitingTakerBond` and `WaitingMakerBond` into the database the first load created, then reloads the page. Finally it compares what the app publishes with the seed.
- The app publishes through `lib/core/web/store_probe.dart`. It runs on web only, is not awaited, and reads through `listBondClaims` and `listTrades`.
- A new Rust test, `web_smoke_seed_rows_decode_as_stored`, decodes `test/web/smoke/seed/bond_store.json`. A field rename then fails `cargo test` instead of a browser run.
- The self-test holds the new check down from both sides:
  - `store-probe` reads the rows back and passes.
  - `store-probe-empty` publishes nothing and fails.
  - `healthy` has no bond stores and fails.
- `pages_bundle_test.dart` guards the probe flag, the CI opt-in and the database name.

### T5.2 — push routing: documented gap
The push server sees kind 14 p-tagged to a registered trade pubkey and sends a content-free wake-up. It cannot decrypt the action, so no payload can say `add-bond-invoice` or `bond-payout-completed`.
- **Typed routing has no producer.** The `type`/`orderId` payloads that `push_notification_service.dart` routes on have no producer for any action today.
- **No token is registered.** `registerToken` has no caller.
- **Claims can target wiped keys.** A claim is addressed to the slashed attempt's key, which may belong to a wiped trade.

The findings are recorded in spec §8.5 and §13, plus a CLAUDE.md gotcha. No routing code was added.

### T5.3 — docs and dead l10n
- **`CLAUDE.md`:** new gotchas say bond statuses never reach the wire book, claims are independent of trades, and push cannot carry bond events. The smoke bullet mentions the store check.
- **`specs/004`:**
  - A new `contracts/bond.md` pins the bridge API.
  - `contracts/types.md` adds the bond statuses, `TradeInfo.bond` and the bond types.
  - `data-model.md` adds `Trade.bond`, bond windows, the `BondClaim` entity and the `bond_claim_retained_nodes` setting.
  - `contracts/orders.md` drops the stale "`BondRequired` … not supported yet", documents the maker and taker bond paths, and adds the five bond actions to the inbound table.
- **Spec status line:** it no longer says Draft, and a "What PR-5 found and did" note follows the Phase 5 table.
- **Stale Rust test comment:** it still claimed `pay-bond-invoice` maps to `BondRequired`.
- **Dead l10n:** 69 keys no Dart code references are removed from all five locales. None are bond-related. The scan was a word-boundary search over `lib/`, `test/`, `integration_test/` and `tool/`, excluding generated code. gen-l10n has no dynamic lookup, and `untranslated_messages.txt` stays empty.

## Test plan
- [x] `cargo test`: 591 passed, including the new seed-decode test. `cargo clippy` shows no new warnings; the 28 existing ones are in untouched files.
- [x] `flutter analyze` is clean.
- [x] `flutter test` passes all 1153 tests.
- [x] `node selftest.mjs` passes all 8 cases, including the 3 new store cases.
- [x] The pre-commit hook passed.
- [ ] CI web-build job: the release bundle smoke run with `SMOKE_BOND_STORE=1` prints `✓ bond rows read back (1 claim, 2 trades)`.

## Manual testing
1. Check out the branch, then run `flutter gen-l10n` and `flutter analyze`. Expect no issues, which confirms no removed l10n key is still used.
2. Build the web bundle with `./scripts/build-web.sh --release`, then run `flutter build web --release --base-href "/app/" --pwa-strategy=none`.
3. In `test/web/smoke`, run `npm ci && npx playwright install chromium`. Then run `BASE_PATH=/app/ SMOKE_BOND_STORE=1 node smoke.mjs`. Expect `✓ bond rows read back (1 claim, 2 trades)` and `web bundle smoke test passed.`
4. Break the decode on purpose. In `test/web/smoke/seed/bond_store.json`, change the claim's `"phase": "Pending"` to `"phase": "Bogus"`, rerun step 3, and expect it to fail with `seeded bond rows were not read back`. Run `cargo test web_smoke_seed_rows_decode_as_stored` too and expect it to fail. Revert the edit.
5. In `test/web/smoke`, run `node selftest.mjs` and expect `all 8 cases behaved as expected`.
6. Run `flutter run -d chrome` with the usual COOP/COEP `--web-header` flags. In DevTools, `window.mostroStoreProbe` should be a JSON string with `claims` and `trades` arrays, both empty on a fresh profile, and there should be no console errors.
7. Run the app on Android or Linux and confirm startup is unchanged. The probe is web-only.
8. Read `docs/ANTI_ABUSE_BOND.md` §8.5 and `specs/004-mostro-p2p-client/contracts/bond.md`, and check the descriptions match the bond flows you know from Phases 1–4.

## Summary by CodeRabbit

- **Documentation**
  - Updated anti-abuse bond documentation to reflect the implemented lifecycle, claim handling, bond states, and notification limitations.
  - Added contract and data-model documentation for bond policies, claims, events, and bonded order flows.

- **Tests**
  - Expanded web release checks to verify bond claims and bonded trades persist and reload correctly.
  - Added coverage for empty or unavailable bond storage scenarios.

- **Maintenance**
  - Removed obsolete localization strings across supported languages.


lib/core/app_bootstrap.dart
```diff
@@ -132,6 +134,10 @@ Future<void> bootstrapAndRun({List<String> seedRelays = const []}) async {
       debugPrint('[main] Mortsom build: orders expire after ${orderExpiry}s');
     }
     markBridgeReady();
+    // Web only, and not awaited: startup does not wait on it. With
+    // SMOKE_BOND_STORE=1 the smoke test seeds bond rows and checks they come
+    // back through the bridge — see lib/core/web/store_probe.dart.
+    if (kIsWeb) unawaited(publishStoreProbe());
```
grunch: _⚠️ Potential issue_ | _🟠 Major_

**The CI-only store probe runs on every production web launch and puts the user's bond rows on `window`.**

`publishStoreProbe()` is gated only on `kIsWeb`, so the deployed app at mostro.network/app decodes every trade and every claim at each startup. It then writes a summary to `window.mostroStoreProbe`: order ids, claim phases, trade statuses and bond states. Nothing in production reads it.

- **Cost:** `listTrades` decodes the whole trades store on each launch for no user-facing purpose.
- **Exposure:** the data already lives in same-origin IndexedDB, but a plain global is readable by anything that runs in the page. That includes injected content scripts, which cannot open the origin's IndexedDB as easily.

The bundle under test must stay the one that ships, so a `--dart-define` gate is the wrong fix. A runtime opt-in that only the smoke test sets keeps the bundle identical:

```diff
- if (kIsWeb) unawaited(publishStoreProbe());
+ if (kIsWeb && storeProbeRequested()) unawaited(publishStoreProbe());
```

```dart
// bridge_probe_web.dart
const kStoreProbeRequestFlag = 'mostroStoreProbeRequested';
bool storeProbeRequested() =>
    globalContext.getProperty<JSAny?>(kStoreProbeRequestFlag.toJS)?.dartify() == true;
// bridge_probe_stub.dart
bool storeProbeRequested() => false;
```

```js
// smoke.mjs, before page.goto, only when SMOKE_BOND_STORE === '1'
await page.addInitScript(() => { globalThis.mostroStoreProbeRequested = true; });
```

Init scripts run again on `page.reload()`, so the seeded load still gets the probe. The store fixtures ignore the flag, and `pages_bundle_test.dart` should assert the request flag on both sides.

test/web/smoke/smoke.mjs
```diff
@@ -324,6 +364,57 @@ async function main() {
...
+      await page.reload({ waitUntil: 'domcontentloaded', timeout: TIMEOUT_MS });
+      await page
+        .waitForFunction(
+          () =>
+            typeof globalThis.mostroStoreProbe === 'string' ||
+            typeof globalThis.mostroStoreProbeError === 'string',
+          undefined,
+          { timeout: TIMEOUT_MS },
+        )
+        .catch(() => fail('the app never published what it read from the store (mostroStoreProbe)'));
```
grunch, line 409: _⚠️ Potential issue_ | _🟡 Minor_

**A bridge failure on the reloaded page is reported as a probe timeout, after the full budget.**

The isolation, mount and bridge checks run only for the first load. After `page.reload()`, this waits only for `mostroStoreProbe` or `mostroStoreProbeError`. The app publishes neither when the second startup's bridge call fails, because `publishStoreProbe` sits inside the rehydrate `try`. The run then burns `TIMEOUT_MS` (120 s in CI) and blames the store probe, while the real cause, `mostroBridgeError`, is never printed.

Waiting on the bridge error too makes it fail fast, with the right reason:

```diff
         .waitForFunction(
           () =>
             typeof globalThis.mostroStoreProbe === 'string' ||
-            typeof globalThis.mostroStoreProbeError === 'string',
+            typeof globalThis.mostroStoreProbeError === 'string' ||
+            typeof globalThis.mostroBridgeError === 'string',
           undefined,
           { timeout: TIMEOUT_MS },
         )
         .catch(() => fail('the app never published what it read from the store (mostroStoreProbe)'));
+      const reloadBridgeError = await page.evaluate(() => globalThis.mostroBridgeError);
+      if (reloadBridgeError) await fail(`Rust bridge call failed after the reload: ${reloadBridgeError}`);
```

CLAUDE.md
```diff
@@ -182,5 +184,19 @@ bridged by flutter_rust_bridge.
...
   daemon rejects with `CantDo` (#203).
+- **Bond statuses never reach the wire book.** `WaitingTakerBond` publishes as `pending` (the
+  order stays takeable by others until a bond locks) and `WaitingMakerBond` publishes nothing
+  (the order is invisible until the maker's bond locks). Both exist only on the local trade row,
+  learned from daemon messages. Never derive them from Kind 38383, and never read a `pending`
+  book entry as "nobody is paying a bond on it" (`docs/ANTI_ABUSE_BOND.md` §2.7–2.8).
+- **Bond payout claims are independent of trades.** A `BondClaim` lives in its own
+  `bond_claims` store keyed `node:order`. It outlives the trade row (completed, canceled or
+  wiped) and is always submitted to the node that issued it, even after a node switch: the
+  kind-14 filter keeps that node as an author while a claim is open, and for its claim window
```
grunch, line 195: _⚠️ Potential issue_ | _🟡 Minor_

**The retention period is understated.**

This says a node the user switched away from stays on the filter "for its claim window". The code keeps it for the claim window **plus** `RETAIN_MARGIN_DAYS` (15), through `bond_claims::retain_until`. `contracts/bond.md` in this same PR states it correctly. CLAUDE.md is the file agents read first, so it should not be the less accurate of the two:

```diff
- kind-14 filter keeps that node as an author while a claim is open, and for its claim window
- after the user switches away.
+ kind-14 filter keeps that node as an author while a claim is open, and for its claim window
+ plus a 15-day margin after the user switches away.
```

lib/core/web/bridge_probe_web.dart
```diff
@@ -23,6 +23,12 @@ const kBridgeReadyFlag = 'mostroBridgeReady';
 /// Set to the error string when that call threw instead.
 const kBridgeErrorFlag = 'mostroBridgeError';
 
+/// Set to a JSON summary of the bond rows read back through the bridge.
+const kStoreProbeFlag = 'mostroStoreProbe';
```
grunch: _🧹 Nitpick_ | _🔵 Trivial_

**The store probe lives in the bridge probe's library, whose docs no longer describe it.**

`bridge_probe.dart`'s library comment still says its only job is proving the Rust bridge answered. It now exports two unrelated signals, and the store flags are documented only in `store_probe.dart`. Either extend the library comment in `bridge_probe.dart` to name both signals, or give the store probe its own conditional export (`store_probe_stub.dart` / `store_probe_web.dart`). The second keeps "rename the flag on one side only" contained to one pair of files.

specs/004-mostro-p2p-client/contracts/bond.md
```diff
@@ -0,0 +1,98 @@
...
+**Delivery coverage.** The global kind-14 filter's authors are the active node,
+every node with a non-terminal claim, and every node the user switched away
+from, kept for its claim window plus a 15-day margin. The retained set is
+persisted in the `bond_claim_retained_nodes` setting. Without it, a slash that
+lands after a switch would be asked for on a node nothing listens to.
+
+### list_bond_claims() → Vec<BondClaim>
+Every claim, most recently changed first. **Errors**: `StorageUnavailable`.
+
+### get_bond_claim(order_id: String) → BondClaim?
```
grunch, line 66: _🧹 Nitpick_ | _🔵 Trivial_

**Missing error.** Both `get_bond_claim` and `get_bond_claim_from` fail with `StorageUnavailable` when the database is not initialised: `claim_for_order` and `get_bond_claim_from` both `ok_or_else` on `app_db::db()`. `list_bond_claims` documents it, but these two do not.

```diff
 The claim for one order: the open one when several nodes issued one, otherwise
-the most recently changed.
+the most recently changed. **Errors**: `StorageUnavailable`.
```

The same line belongs under `get_bond_claim_from`.

specs/004-mostro-p2p-client/data-model.md
```diff
@@ -215,6 +222,39 @@ under different ids.
...
+| submitted_invoice | String? | The bolt11 sent |
+| fiat_code | String | Display only |
+| fiat_amount | f64? | Display only |
+| payment_method | String | Display only |
+| updated_at | Timestamp | Last change; list order |
+
+**Validation rules**:
+- A cadence retry of `add-bond-invoice` never re-arms a `Submitted` claim; a
+  re-prompt after `Acknowledged` does.
+- `Completed` and `Expired` are terminal.
```
grunch: _🧹 Nitpick_ | _🔵 Trivial_

**The rules omit two upsert cases that change what a stored claim means.** `bond_claims::upsert_claim` does more than this list says:

- **Expired claims can reopen.** An `Expired` claim re-prompted while `now <= deadline_at` goes back to `Pending`, so "terminal" is only true once the frozen deadline has passed.
- **A new slash replaces the claim.** An `add-bond-invoice` with a different `slashed_at` for the same order replaces the stored claim outright, with a new deadline.

```diff
-- `Completed` and `Expired` are terminal.
+- `Completed` is terminal; `Expired` is terminal once `deadline_at` has passed
+  (a re-prompt inside the frozen deadline reopens it as `Pending`).
+- A request with a different `slashed_at` for the same (node, order) replaces
+  the claim, with a deadline computed afresh.
```

rust/src/api/bond.rs
```diff
@@ -467,6 +467,80 @@ pub fn on_bond_slashed() -> BondSlashedStream {
...
     use super::*;
 
+    /// The web smoke test seeds these rows into IndexedDB and expects the
+    /// release build to read them back (docs/ANTI_ABUSE_BOND.md T5.1).
+    /// Decoding the file here keeps it in step with the stored types: a
+    /// renamed field fails `cargo test` instead of a CI browser run.
+    #[test]
+    fn web_smoke_seed_rows_decode_as_stored() {
+        let seed: serde_json::Value =
+            serde_json::from_str(include_str!("../../../test/web/smoke/seed/bond_store.json"))
```
grunch, line 479: _🧹 Nitpick_ | _🔵 Trivial_

**The test compiles a file from outside the crate.**

`include_str!("../../../test/web/smoke/seed/bond_store.json")` resolves outside `rust/`. It is `#[cfg(test)]` and CI checks out the whole repo, so nothing breaks today. Any context that builds the crate's tests from `rust/` alone would fail to compile the whole test target, not just this test, for example a `cargo package` verify step or a container that copies only `rust/`.

Acceptable as is, because the coupling to the smoke seed is the point. A one-line comment saying the relative path is intentional would stop someone from "fixing" it into a crate-local copy that silently drifts.

grunch (commented):
## Strict review — PR-5 (web store smoke, push gap, docs, dead l10n)

**Actionable comments posted: 7** (1 major, 2 minor, 4 nitpicks)

### Verified
- **CI actually exercises the new check.** The web job log shows `SMOKE_BOND_STORE: 1` and `✓ bond rows read back (1 claim, 2 trades)` on the release bundle. The self-test reports all 8 cases, including the 3 store cases.
- **Seed decode:** `web_smoke_seed_rows_decode_as_stored` decodes every seeded document into `BondClaim` / `TradeInfo`, checks storage keys, and requires both bond statuses to be present.
- **Contract claims match the code:**
  - `pay-bond-invoice` on a take is an acceptance at `WaitingTakerBond` (`classify_take_reply` test).
  - A `canceled` in the taker window wipes the row, with `BondLostRace`, `MakerCanceled` or `UserCanceled` (`bond_cancel_reason`, `a_canceled_during_the_bond_window_reads_as_a_lost_race`).
  - `BondSlashed` overwrites a provisional `Released`.
  - The claim upsert table matches `upsert_claim`, apart from the two omissions noted inline.
  - A maker's local deadline is the earlier of the bolt11 expiry and the order's `expires_at` (`bond_deadline`).
- **Push gap:** `registerToken` has no caller, and the push contract (`contracts/nostr.md`) is content-free. Documenting rather than routing is the right call.
- **l10n removal:** the key order and values of every remaining entry are unchanged in all five ARBs. `flutter analyze` and `flutter gen-l10n` are clean, and `untranslated_messages.txt` stays empty.

### Findings
| Severity | File | Issue |
|---|---|---|
| 🟠 Major | `lib/core/app_bootstrap.dart` | The CI-only probe runs in production and publishes bond rows on `window` |
| 🟡 Minor | `test/web/smoke/smoke.mjs` | A bridge failure after the reload surfaces as a 120 s probe timeout |
| 🟡 Minor | `CLAUDE.md` | Retention after a node switch omits the 15-day margin |
| 🔵 Nit | `lib/core/web/bridge_probe_web.dart` | The store probe lives under the bridge probe's docs |
| 🔵 Nit | `contracts/bond.md` | `StorageUnavailable` is missing on the two claim getters |
| 🔵 Nit | `data-model.md` | The claim rules omit Expired reopening and replacement by a new `slashed_at` |
| 🔵 Nit | `rust/src/api/bond.rs` | `include_str!` reaches outside the crate |

### Recommendation
Address the major one before merging: gate the probe on a flag only the smoke test sets. That keeps the shipped bundle identical and the check just as strong. The two minors are small doc and script fixes, and the nitpicks are optional.

docs/ANTI_ABUSE_BOND.md
```diff
@@ -1,6 +1,6 @@
 # Anti-Abuse Bond — Client Implementation Spec & Phased Plan
 
-**Status:** Draft — planning document for epic [#145](https://github.com/MostroP2P/app/issues/145)
+**Status:** Implemented — every phase in §10 has landed (PR-5 closes epic [#145](https://github.com/MostroP2P/app/issues/145)); kept as the design reference for bond support in this client
```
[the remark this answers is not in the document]
grunch: Fixed in f34a851. Section 13 now marks items 1, 2, 3, 4 and 6 as resolved, each with the phase and the code or test that settled it:

- **1. Restart coverage:** the verification is recorded in §9.
- **2. Restore:** `persist_restored_bond_rows` builds a row for every restored order parked on a bond.
- **3. Decoder:** `lightning-invoice` 0.34 is in the tree and builds for wasm in CI.
- **4. Amount seeding:** the bond amount never seeds the order's sats, and `classify_take_reply` plus the dispatch tests assert it.
- **6. Lost race:** `a_canceled_during_the_bond_window_reads_as_a_lost_race` covers it.

Item 5 was already resolved, and item 7 stays as non-blocking upstream follow-ups, unchanged.

lib/core/app_bootstrap.dart
grunch: Fixed in f34a851. The store read-back now lives in its own signal files (`store_probe_signal{,_stub,_web}.dart`). It runs only when `window.mostroStoreProbeRequested` is `true`.

- **Only the smoke test asks.** `smoke.mjs` sets that flag through `page.addInitScript` when `SMOKE_BOND_STORE=1`, so the request is in place on the first load and again on the reload.
- **Production is untouched.** A normal launch reads nothing and publishes nothing. The shipped bundle is still the one under test.
- **The fixture proves the request.** `store-probe` now publishes only when asked, so its passing self-test case also proves the request reaches the page across the reload.
- **The flag is guarded.** `pages_bundle_test.dart` asserts the request flag on both the Dart and JS sides.

test/web/smoke/smoke.mjs
grunch: Fixed in f34a851. After the reload, the smoke test now also waits on `mostroBridgeError` and fails straight away with `Rust bridge call failed after the reload: …` instead of timing out on the probe.

CLAUDE.md
grunch: Fixed in f34a851. CLAUDE.md now says the node is kept for its claim window plus a 15-day margin, matching `retain_until` and `contracts/bond.md`.

lib/core/web/bridge_probe_web.dart
grunch: Fixed in f34a851. I took the second option. The store signals moved out of `bridge_probe_*` into their own conditional export, so the bridge probe files are back to exactly what `main` has. A flag renamed on one side stays contained to the store probe pair.

specs/004-mostro-p2p-client/contracts/bond.md
grunch: Fixed in f34a851. `StorageUnavailable` is now documented on both `get_bond_claim` and `get_bond_claim_from`.

specs/004-mostro-p2p-client/data-model.md
grunch: Fixed in f34a851. The rules now say `Expired` is terminal only once `deadline_at` has passed, since a re-prompt inside the deadline reopens it as `Pending`. They also say a different `slashed_at` replaces the claim with a fresh deadline, which matches `upsert_claim`.

rust/src/api/bond.rs
grunch: Fixed in f34a851. Added a comment above the `include_str!` saying the path leaves the crate on purpose: it is the file the smoke test seeds, and a crate-local copy would drift unnoticed.

grunch:
@coderabbitai On the **Out of Scope Changes** warning about the 69 removed l10n keys: this is in scope, and I'm keeping it.

Task T5.3 in `docs/ANTI_ABUSE_BOND.md` explicitly includes "remove dead l10n". That wording was in the Phase 5 table on `main` before this PR was opened. The epic closes with this PR, so the cleanup was planned here rather than as a separate chore.

The removal is limited to keys no Dart code references. The scan was a word-boundary search over `lib/`, `test/`, `integration_test/` and `tool/`, excluding generated code, and gen-l10n has no dynamic key lookup. Every remaining entry is unchanged in all five locales. `flutter analyze` and `flutter gen-l10n` are clean, and `untranslated_messages.txt` stays empty.

Review round 1 is in f34a851. It addresses your Section 13 comment and the 7 comments from the strict review.

PR #493, opened 2026-09-17
perf(relay): PR 3.6 + the 2.5 gap — batch key derivation, coalesced DM filter and Online sync

grunch:
## Summary

`docs/OPTIMIZATION_PLAN.md` **PR 3.6**, plus the *remaining gap* recorded under **PR 2.5**. Three independent commits, Rust only, no change to the public `rust/src/api/` surface (no FRB regeneration).

### 1. `perf(keys)` — one seed derivation for all trade keys

`derive_trade_key` turns the mnemonic into a seed **on every call** (2048 rounds of PBKDF2) and re-walks the four hardened levels. `build_trade_key_map` called it once per index, so startup and every node switch cost one PBKDF2 per trade the user ever made.

`derive_trade_keys(words, up_to)` derives the seed and the `m/44'/1237'/38383'/0` parent once, then one non-hardened child step per index. The plan proposed deriving "concurrently"; this makes the work disappear instead, and needs no runtime (wasm has none).

| 100 keys | per index | batch |
|---|---|---|
| release | 211 ms | **9.6 ms** |
| debug | 2.17 s | **0.12 s** |

A test holds both paths equal for indexes 1..=5; the reviewer also confirmed against `bip32` 0.5.3 that `derive_from_path` is a fold of `derive_child`, so the two are identical by construction. The derivation path constant is untouched.

### 2. `perf(relay)` — coalesce the bulk DM filter refresh, off the user's path

`ensure_global_dm_coverage` re-issued `mostro-dm` for every new key: CLOSE + a REQ **without `since`** on every relay, each replaying the whole stored kind-14 history — and `create_order` / `take_order` **awaited it before publishing**.

Nothing needs that wait. The key joins `GLOBAL_DM_KEYS` synchronously (decryption works at once), and the reply to the request about to go out arrives on the per-trade subscription, which every caller does await before publishing — verified for all four call sites (create, take, range remainder, restore). The refresh now runs through `nostr::coalesce::Coalesced`: once per burst, in the background.

**Coverage is still never pruned**, as the plan's revised fix and `contracts/orders.md` require. Only how often and where the REQ is issued changes.

### 3. `fix(relay)` — the 2.5 gap: Online transitions no longer queue up runs

The connection watcher ran the Online sequence inline (10 s capability fetch, outbox flush, order/chat resubscribes). Transitions arriving meanwhile queued in the channel and each re-ran all of it, back to back. Coalesced, a burst costs one run plus at most one more for whatever arrived mid-run, and the watcher keeps observing instead of blocking. The window is 100 ms, not more: the first `Online` of a cold start gates the order-book subscription.

Known and documented in code: an `Offline` during the window or a run does not cancel it (the inline version could not either). The plan's other 2.5 leftovers — scheduling outbox retries and retrying the capability fetch — are **not** in this PR.

### `Coalesced`

Armed flag; `request` spawns sleep → disarm → run. Disarming **before** the run is the point: a request landing mid-run arms another, because the run read its input at start. Four tests: burst → 1 run, caller does not wait, runs again after the window, request-during-run → 2 runs. Works on wasm through `rt::MaybeSend`.

## Test plan

- [x] `cargo test` — 673 passed (3 new in `crypto::keys`, 4 in `nostr::coalesce`)
- [x] `cargo check --target wasm32-unknown-unknown`
- [x] Independent Rust review: no CRITICAL/HIGH
- [ ] CI green

## Manual testing

1. Build and run this branch with an identity that already has several trades (Rust changed → full rebuild). Log level **debug**.
2. Cold start: the log line `trade key map: N keys derived for daemon-message decryption` must show the same `N` as on `main`, and appear noticeably sooner on a debug build with many trades.
3. Create an order. Expect `daemon-message subscription active for trade=…` **before** the publish lines, the order confirmed as usual, and — about half a second later, not before the publish — a single `sub replaced id=mostro-dm kinds=[14] p_count=N+1`.
4. Take an order on a second instance: same expectation, and the take must not feel slower than on `main` (it no longer waits for the relay round trip).
5. Create a **range** order and let it be partially taken, then release: the remainder's key and the release arrive back to back — expect **one** `sub replaced id=mostro-dm` for the burst, not two.
6. Restart the app mid-trade and have the counterparty advance the trade while it is closed: on start the missed status must still arrive (the bulk filter replays history exactly as before).
7. Toggle the network off and on (airplane mode) a few times quickly: expect one `relay pool ONLINE — fetching node capabilities…` line per settled reconnection, not one per toggle, and queued messages flushed once back online.

## Summary by CodeRabbit

* **Performance**
  * Improved trade-key loading by deriving keys in batches, reducing startup work.
  * Relay synchronization and message-filter updates now coalesce rapid requests to avoid redundant operations.

* **Reliability**
  * Improved handling of online state changes and synchronization requests during active operations.
  * Added coverage for batched key derivation and request coalescing behavior.


grunch:
@coderabbitai review

PR #495, opened 2026-09-17
feat(core): PR 3.1 — map-backed order book with revisioned deltas

grunch:
## Summary

`docs/OPTIMIZATION_PLAN.md` **PR 3.1**, first of the 3.1 → 3.2 → 3.3 chain (the next two are stacked on this branch).

The order book was `Arc<RwLock<Vec<OrderInfo>>>`: every upsert, status change and lookup was a linear scan, and the only thing it could broadcast was a clone of the whole vector. Now:

- **`BookState { orders: HashMap<String, OrderInfo>, revision: u64 }`** — O(1) upsert / lookup / removal.
- **`OrderBookDelta`** — `Upserted { revision, order }`, `Removed { revision, order_id }`, `Reset { revision }` (the book was replaced or cleared wholesale; what vanished cannot be told order by order, so the subscriber starts over).
- **Nothing downstream changes yet.** The FRB snapshot stream keeps emitting whole-book snapshots, now built from the map and **ordered by id** so they are deterministic (a map has no order; display order was always Dart's).

### The contract this establishes (what 3.2 will expose)

1. A delta is sent **while the book's write lock is held** → subscribers receive them in revision order, one per change. That includes `upsert_order_deferred` / `upsert_order_coalesced`: their batching was only ever about snapshots being the whole book; a delta is one order.
2. **Resync needs a boundary, not just a refetch** (the plan's own warning). `snapshot_with_revision()` reads book + revision under one lock; a subscriber that lagged applies only deltas **newer** than it. A delta at or below is already inside the snapshot, and replaying it could undo a later change — the test interleaves a removal *before* the snapshot read with a re-insert and a status change *after* it, and checks the mirror converges on the book.
3. An order re-announced **unchanged** costs no revision and no delta (`OrderInfo` gains `PartialEq`). That is the common case on the wire and, from 3.2 on, a bridge message saved.

The delta API has no production caller until 3.2, so it carries `#[allow(dead_code)]` with that reason; 3.2 removes the allowances.

## Test plan

- [x] `cargo test` — 674 passed (8 new: upsert/revision, deferred+coalesced emit at once, removal only when something was removed, status change = upsert, replace/clear = Reset, unchanged upsert emits nothing, interleaved resync converges, snapshot stream still whole-book)
- [x] No existing test depended on insertion order of snapshots
- [x] `cargo clippy --all-targets` — no new warnings; `cargo check --target wasm32-unknown-unknown`
- [x] FRB regenerated (comment-only diff in generated Dart) and committed; `flutter analyze` clean
- [ ] CI green

## Manual testing

1. Build and run this branch (Rust changed → full rebuild). Home must list the same orders as `main` against the same node, in the same display order (sorting is Dart-side and unchanged).
2. Pull-to-refresh: the list reloads without flicker or duplicates.
3. Create an order → it appears in **My orders** and in the book; cancel it → it disappears from both.
4. Take an order from a second instance: on the maker, the order leaves the public book and the trade shows its private status; on the taker, the trade detail opens with the right status (exercises `update_order_status` and `get_order` on the map).
5. Switch Mostro node in Settings: the previous node's orders vanish at once (`clear` → `Reset`) and the new node's book loads.
6. Let a taken order time out (or cancel the take before paying): the order returns to the public book as `pending` for the ex-taker (`settle_after_lost_take`).


grunch:
@coderabbitai review

PR #496, opened 2026-09-17
feat(bridge): PR 3.2 — order-book delta stream over FRB

grunch:
> **Stacked on #495** (PR 3.1). Review that first; this PR's diff is against its branch and will retarget to `main` once it merges.

## Summary

`docs/OPTIMIZATION_PLAN.md` **PR 3.2**: the per-order deltas of #495 cross the bridge.

- **`on_order_deltas() -> OrderDeltaStream`** — `next()` yields `OrderDelta`:
  - `Upserted { revision, order }`
  - `Removed { revision, order_id }`
  - `Resync`
- **`get_order_book_snapshot() -> OrderBookSnapshot { revision, orders }`** — the whole book (every status, exactly what the snapshot stream carries) read under one lock with its revision.

### The consuming rule (documented on `OrderDelta`, so it shows up in the generated Dart docs)

1. Subscribe **first**, then read the snapshot — no change can fall between the two.
2. Apply only deltas with `revision` greater than the snapshot's / the last applied. A delta at or below it is already inside the snapshot; applying it could resurrect a removed order.
3. On `Resync`, read a fresh snapshot and continue with the same rule.

`Resync` covers the two things a consumer cannot patch: the book was **replaced or cleared** (node switch), and the subscriber **lagged**. The snapshot stream can ignore a dropped message because the next one is complete; a dropped delta is a hole, so it is never swallowed.

Revisions cross as **`u32`** — a plain `int` in Dart on every target, where `u64` is a `BigInt`. Past its range (four billion changes in one process) the stream degrades to `Resync` per change: slower, never wrong.

### Not in this PR

Nothing in Dart consumes the stream yet — that is PR 3.3, stacked on this. The **snapshot stream stays**, as the plan asks ("keep one release for fallback, then remove").

### Housekeeping from 3.1

With a production caller, the `#[allow(dead_code)]` markers of #495 are gone. `OrderBookDelta::Reset` loses its `revision` field: nothing read it, since a consumer takes the revision from the snapshot it has to read anyway.

## Test plan

- [x] `cargo test` — 678 passed. 4 new, run against a **local** `OrderBook` through `OrderDeltaStream::over` (the global book is shared by parallel tests), with a `Mirror` that does in Rust what Dart will do:
  - snapshot + deltas after it ≡ book (add, status change, removal)
  - deltas already inside the snapshot are skipped (subscribe → add → remove → snapshot → empty mirror)
  - a cleared book reaches the consumer as `Resync`
  - a subscriber flooded past the channel capacity gets `Resync`, re-reads, and converges again
- [x] `cargo clippy --all-targets` — no new warnings; `cargo check --target wasm32-unknown-unknown`
- [x] FRB regenerated and committed in the same commit; `flutter analyze` clean
- [ ] CI green (incl. the generated-code check and web build)

The plan's "Dart integration test: snapshot + deltas ≡ Rust book" cannot run under `flutter test` (no Rust library there); the Rust `Mirror` tests pin the same contract at the API boundary, and 3.3 tests the Dart side of it against a fake stream.

## Manual testing

This PR adds API without a consumer, so the check is that nothing regressed:

1. Build and run this branch (Rust + bindings changed → full rebuild) on native **and** web (`./scripts/build-web.sh` first). The app must start and Home must list the node's orders as before — a bindings mismatch would fail at startup.
2. Create an order, cancel it; take an order from a second instance. Book and trade screens behave as on #495.
3. Switch Mostro node in Settings and back: the book empties and refills each time.
4. Web only: reload the page mid-session; the book loads again (the wasm bridge with the regenerated bindings works end to end).


grunch:
Added `OrderDelta::Loaded` (commit `7194a2c`): found while building the Dart consumer for 3.3. An empty book produces no per-order delta, so a delta consumer needs the same EOSE confirmation the snapshot stream already gets from `publish_on_stored_events_end`, or the Home screen stays in its loading state forever against a quiet node. Two tests: the pending feed's EOSE reaches the consumer as `Loaded`; another feed's EOSE sends nothing.

grunch:
Review round 1: added `OrderBookSnapshot.loaded` — see #498 for the scenario (a consumer created after the EOSE never hears `Loaded`). Set together with the event under the book's write lock, so *subscribe → read `loaded == false`* guarantees the event is still to come; reset by `clear()`.

PR #498, opened 2026-09-17
perf(ui): PR 3.3 — the order book follows deltas; one mapping per changed order

grunch:
> **Stacked on #496** (PR 3.2), which is stacked on #495 (PR 3.1). This PR's diff is against #496's branch and retargets to `main` as they merge.

## Summary

`docs/OPTIMIZATION_PLAN.md` **PR 3.3** — the Dart end of the delta pipeline, and the step where the whole chain starts paying off.

**Before:** every change to the book crossed the bridge as the *entire book*, and `orderBookProvider` re-mapped every `OrderInfo` into a fresh `OrderItem`. One relay event cost O(orders) in Rust, on the bridge and in Dart — and every screen watching a single order was handed a "new" object each time.

**After:** a change crosses as the one order it concerns.

### `OrderBookFeed` (`order_book_feed.dart`, pure Dart, generic, no Riverpod)

- Follows the rule documented on `OrderDelta`: **subscribe first**, read the revisioned snapshot, apply only deltas newer than it, start over on `Resync`.
- **One mapping per changed order.** Untouched orders keep their *identity* across emissions, so `orderByIdProvider`'s `select` sees nothing move while the rest of the book does.
- **Emits at most once per 50 ms.** A delta is one order; a cold start delivers thousands back to back. Emitting on each would rebuild the O(N²) start-up on the Dart side (every emission walks the book in the filter, the sort and the index).
- **Loading-state semantics preserved.** An empty book is not shown until the relay's EOSE confirms it (`OrderDelta.loaded`), so "no orders" never flashes before the orders arrive — and a quiet node still leaves loading. A `Resync` to an empty book *is* shown if orders were on screen (node switch).
- The bridge sits behind `orderDeltaSourceProvider`, so all of this is tested against a scripted source.

### Filter hot path

`filteredOrdersProvider` split, trimmed and lower-cased **every order's** payment methods — and rebuilt the selected set — **per order, per pass**. `OrderItem.paymentTokens` is now computed once per order (orders persist across emissions now, so it sticks), and the selected set once per pass.

### Deliberately not done (recorded in the plan)

- **No incrementally maintained sorted list.** With mapping and per-order allocations gone, the filter is a pointer walk plus a sort, at most 20×/s. Three sort orders with ties and five filters make an incremental structure real complexity for microseconds — to be measured in PR 5.2 first.
- **Rust `OrderFilters` stays dead code.** Delete-or-wire-up is the PR 3.8 decision; nothing here forces it.
- **The snapshot stream is not removed** — the plan keeps it one release as fallback.

`orderBookProvider` keeps its type (`StreamProvider.autoDispose<List<OrderItem>>`): none of its consumers or the ten test overrides change.

## Test plan

- [x] `flutter test` — all passing. New: 10 `OrderBookFeed` tests (subscribe-before-snapshot, non-empty snapshot shown at once, empty book silent until `loaded`, upsert/remove, stale delta skipped, 300-delta burst → **1** emission, only the changed order is mapped and the other keeps identity, resync re-reads, resync-to-empty empties the list, closes with the bridge) + 2 filter tests (case/padding, tokens computed once)
- [x] `flutter analyze` — clean; formatter kept to touched files
- [x] No Rust change in this PR
- [ ] CI green (incl. web build + smoke — Home mounting is what the smoke test asserts)

## Manual testing

1. Build and run the top of the stack (Rust from #495/#496 changed → full rebuild) on native **and** web.
2. Cold start against the default node: Home shows the skeleton, then the orders — no "no orders" flash in between, and the list does not visibly rebuild several times while it fills.
3. Point the app at a node with **no pending orders** (or filter to nothing is *not* the same — use a quiet node): Home must leave the skeleton and show the empty state within a moment of connecting, not spin forever.
4. With Home open, create an order from a second instance: it appears within a moment, without the list jumping; cancel it: it disappears.
5. Open an order's detail (take screen), and meanwhile create/cancel *other* orders from the second instance: the open screen must not flicker or rebuild (its order did not change).
6. Filters: select a payment method written in different case/padding than the orders use (e.g. orders say `Wire, Revolut`, select `revolut`); then combine with currency, rating and premium filters, and each of the three sort orders. Results must match `main`.
7. Switch Mostro node in Settings: the previous node's orders vanish at once and the new node's appear; switch back.
8. Pull-to-refresh: reloads without duplicates. Background the app for a few minutes on a busy node, then resume: the book is current (a lagged stream resyncs from a snapshot).
9. Leave Home for Trades/Chat and come back: the book is there immediately (provider re-created from a non-empty snapshot).


grunch:
Review round 1 (independent review of the whole 3.1→3.3 stack: no CRITICAL/HIGH). Its one MEDIUM was a gap inherited from the snapshot pipeline: Home **re-created** over a genuinely empty book waits forever, because the EOSE behind `loaded` comes once per subscription, not once per screen. Closed across the stack: #496 adds `OrderBookSnapshot.loaded` (set with the event under the write lock, cleared on a node switch) and this PR makes `OrderBookFeed` show a confirmed-empty snapshot at once. Two new Dart tests, two new Rust tests.

PR #501, opened 2026-09-17
ci(release): attach Linux, Windows, macOS and iOS builds to a release

grunch:
> Phase 2 of #500 (merged).

## Summary

Phase 2 of the tag-triggered release: every release now also carries

| Asset | |
| --- | --- |
| `mostro-vX.Y.Z-linux-x64.tar.gz` | Flutter bundle, built on Ubuntu 22.04 → runs on glibc 2.35+ |
| `mostro-vX.Y.Z-windows-x64.zip` | `mostro.exe` + DLLs |
| `mostro-vX.Y.Z-macos-universal.zip` | `mostro.app`, Apple Silicon + Intel |
| `mostro-vX.Y.Z-ios-unsigned.ipa` | unsigned, for sideloading |

### How it is put together

- **`release-builds.yml`** — reusable workflow, one job per platform. Called by `release.yml` *and* by the new **`release-dry-run.yml`**, which runs the same four builds on PRs that touch the release workflows, `linux/`, `windows/`, `macos/`, `rust_builder/` or a lockfile, and publishes nothing. Windows, macOS and iOS cannot be compiled on a Linux dev machine, so this is the only place a broken build shows up before a release does. **This PR triggers it — see the "Release dry run" checks below.** It is path-filtered, so it must never become a required check.
- **The APKs are the release.** `publish` waits for the desktop/iOS builds but only *requires* `android`. If one fails, the release goes out with what was built, and the notes — rendered again from the files that actually exist — name the missing platform instead of linking to a 404. The run ends red; *Re-run failed jobs* adds the missing asset to the same release.
- **Release notes** gained a Desktop and an iOS section (`tool/release/downloads.dart`), including what each OS does with a build no vendor certificate signed: Gatekeeper (`xattr -dr com.apple.quarantine`), SmartScreen (*More info → Run anyway*), and that the `.ipa` only works through a re-signing sideloader — without push.

### Bug found on the way

**The macOS app could not reach any relay.** It is sandboxed, and neither `.entitlements` file had `com.apple.security.network.client`. It builds and launches fine, which is why nothing caught it. Added, with a test.

## What this does *not* do (needs accounts/certificates the project does not have)

- macOS: no Developer ID signature, **not notarized**
- Windows: no Authenticode signature
- iOS: no TestFlight / signed build — that needs an Apple Developer account, a distribution certificate and a provisioning profile as secrets
- Linux: plain bundle, no `.deb` / AppImage / Flatpak

`docs/RELEASING.md` lists what closing each gap takes.

## Test plan

- [x] `flutter test test/ci` — 52 tests (new: asset-name contract between workflows and notes, single build definition for release and dry run, dry run has no write access or secrets, `publish` requires only `android`, `ditto` packaging, macOS network entitlement)
- [x] `flutter analyze` — no issues
- [x] Local `flutter build linux --release …` + the workflow's `tar` step: 33 MB archive that unpacks to `mostro`, `lib/`, `data/`
- [x] Generator run end to end with a partial asset set (only APKs + Linux): Desktop table lists Linux only, notice names "Windows, macOS and iOS"
- [x] **Release dry run on this PR**: Linux / Windows / macOS / iOS jobs green — after three Windows fixes it surfaced (CMake 4 vs the Firebase C++ SDK, the cargokit manifest path through the plugin junction, `permission_handler_windows` vs MSVC 14.51 coroutines)
- [ ] Not verifiable in CI: that the macOS app *runs* (keychain access under an ad-hoc signature is the thing to check on a real Mac) and that the `.ipa` sideloads

https://claude.ai/code/session_01DUT7HFjXvsLcBV58EMhCC1


grunch:
@coderabbitai review

PR #505, opened 2026-09-18
fix(orders): tell both sides about a cooperative cancel request

grunch:
## Summary

Once a trade is `active`, a cancel is a **request** the counterparty must agree to (protocol `cancel.md`, "Cancel cooperatively"). Mostro confirms the requester with `cooperative-cancel-initiated-by-you` and tells the counterparty with `cooperative-cancel-initiated-by-peer`. This client received both messages and dropped them as "no status change": Alice was never told her request was pending, and Bob never learned anyone had asked. If Bob happened to cancel too, Mostro still settled the trade, so the flow "worked" only by accident.

A second bug hid behind the first: `cancel_order` marked an active trade's local row `Canceled` right away. Alice saw a cancelled trade the daemon still ran, and that terminal status then made the daemon's `cooperative-cancel-accepted` look like a replay over a finished trade, so it was dropped too (`status_sync_blocked_by_terminal`).

### Rust
- `apply_local_cancel`: from `active` on, keep the status and record `cooperative_cancel_state = RequestedByMe` on the row. Never-active rows are still left for the daemon's `Canceled`, and an `in-progress` row that turns out to be a never-active take is settled by that `Canceled` as before.
- New `Storage::set_cooperative_cancel_state` (SQLite `json_set`, IndexedDB via `trade_json`). The field already existed on `TradeInfo` and was never written.
- `cooperative-cancel-initiated-by-{you,peer}` get their own dispatch arm (same gates as a status sync): persist who asked and emit a `TradeUpdate` with the row's **current** status (`Active` or `FiatSent`) and a new reason, `TradeUpdateReason::CooperativeCancelRequestedByMe` / `ByPeer`.

### Dart
- `CancelRequestNotice` on the trade screen while the request is open (`active` / `fiatSent`); hidden once a dispute or an outcome takes over.
- Requester: the action bar drops `Cancel` (asking twice is not an action); Dispute and the step's primary action stay. Scoped to `active` / `fiatSent` in `TradeView.of`, so the flag the row keeps after a dispute does not remove the seller's own cancel there.
- Counterparty: `Cancel` reads **Accept cancel**, and the confirmation dialog says the cancel ends the trade for both. The dialog watches the trade row, so a request landing while it is open updates the copy.
- Notification cards for both reasons, keyed on status + reason so each request is one card.
- Strings in en/es/fr/de/it. `docs/automation-contract.md` gains `trade.cancelRequest`; `specs/004` contract and spec updated.

### Design notes
- The protocol has no "request withdrawn" signal, so `cooperative_cancel_state` is never cleared. While the trade stays `active`/`fiatSent` the notice stays, which matches what the daemon believes. v1 behaves the same.
- Cards are raised for the requester's own confirmation as well: the trade looks untouched otherwise, and the card is what says it now waits on the other side.

## Tests
- Rust: `cancel_of_an_active_trade_is_a_request_the_daemon_settles` (replaces `cancel_of_an_active_trade_still_marks_it_canceled`, which locked the old behaviour in) and `a_peers_cancel_request_is_remembered_and_announced` in `orders.rs`; SQLite round trip; `trade_json` unit test. `cargo test` (709) and `cargo clippy -D warnings` pass.
- Dart: `trade_view_test` (drop rule and its bounds), `notification_model_trade_cards_test`, `event_cards_test`, new `cancel_request_notice_test`, and three `trade_detail_screen_test` cases. `flutter analyze` clean, `flutter test` 1394 passed.

## Manual testing

Two devices (or two profiles), one Mostro node, a trade taken to `active` between Alice and Bob.

1. On Alice's device open the trade and tap **Cancel**, then **Yes, cancel**. Expected: snackbar "Cancel request sent"; the trade stays on the trade screen with the notice "You asked to cancel this trade…"; the **Cancel** button is gone, **Open dispute** stays; the buyer's **Fiat sent** (or the seller's release) stays. In Notifications, a card "Cancel requested".
2. On Bob's device, without touching anything. Expected: a card "The other party wants to cancel"; opening the trade shows the notice "The other party asked to cancel this trade…" and the cancel button now reads **Accept cancel**.
3. On Bob's device tap **Accept cancel**. Expected: the dialog reads "The other party asked to cancel. Cancelling now ends the trade for both of you…". Confirm.
4. Both devices. Expected: the trade reads cancelled by agreement and the card "Trade canceled by agreement" appears. Before this PR Alice's side stayed at the stale local `Canceled` and never received the acceptance.
5. Repeat step 1 after Bob has marked **Fiat sent**. Expected: same notices on both sides, on the `fiat-sent` step.
6. Regression, never-active cancel: create an order, take it, and cancel from the pay-invoice or add-invoice screen before the trade goes active. Expected: unchanged, the trade leaves My Trades once the daemon's `Canceled` lands.
7. Regression, dispute after an unanswered request: as Alice, after step 1, open a dispute. Expected: the notice disappears; the seller's action bar in the dispute still offers its cancel.

https://claude.ai/code/session_01HoK5CZrdXRuVfSjnjZMwJT

## Summary by CodeRabbit

- **New Features**
  - Added cooperative cancellation requests for active and fiat-sent trades.
  - Trade screens now show who requested cancellation, remove the requester’s cancel action, and let the other party accept.
  - Added notifications and confirmation messaging for pending requests.
  - Added localized support in English, German, Spanish, French, and Italian.

- **Bug Fixes**
  - Cancellation requests no longer immediately mark eligible trades as canceled.

- **Documentation**
  - Updated trade automation and cancellation behavior documentation.


grunch:
@coderabbitai review

Issue #523, opened 2026-09-19
One chat REQ per trade trips nos.lol's 'too many concurrent REQs'


grunch:
## Problem
`resubscribe_active_chats` (`rust/src/api/messages.rs` ~l. 1569) opens one subscription per trade that can still chat (`chat_still_relevant`: any non-terminal status with a counterparty), id `mostro-chat-<order>` (`chat_subscription_id`). With a few dozen such trades plus `mostro-dm`, `mostro-orders`, `mostro-orders-recent`, `mostro-relay-list` and the per-trade `mostro-daemon-*`, the app exceeds nos.lol's concurrent-REQ limit.

## Observed
After a seed import, a log report shows ~30 `mostro-chat-*` subscriptions and, on `wss://nos.lol`, repeated:

```
[WARNING] relay: notice relay=wss://nos.lol msg=ERROR: too many concurrent REQs
```

Any REQ the relay refused does not exist there — peer chat, dispute chat or daemon messages for those trades are silently not received from that relay (`live_subs` re-issues on *connect*, not on a NOTICE refusal; see docs/RELAYS.md).

## Contributing factor
Rows recovered from history that never reach a terminal status keep a chat REQ open forever. A seller's completion only arrives through the public Kind 38383 `success` (the daemon sends `PurchaseCompleted` to the buyer only), so restored seller trades parked at `SettledHoldInvoice` / `Pending` qualify as "live" — tracked separately (restored orders shown as in progress).

## Options
- Merge the chat filters into one (or a few) subscriptions: a single REQ whose `#p` lists every `pub(K_conv)`, replaced through `live_subs::replace` as trades come and go — the same pattern as the bulk kind-14 DM filter.
- Treat a `too many concurrent REQs` NOTICE as a failed REQ and back off / consolidate, rather than assuming it exists.

Related: #182 (subscription lifecycle: leaks and deterministic ids — not the count).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

grunch:
More evidence, and it now breaks restores.

Log of a seed import whose restore got no reply within 10 s (`NoDaemonResponse`), while the next import worked:

```
10:59:25 orders: daemon-message watcher already live for trade=ef9f72ca — re-arm refreshed its lease
10:59:25 restore: RestoreSession published trade_index=1 — waiting for daemon
10:59:32 relay: notice relay=wss://nos.lol msg=ERROR: too many concurrent REQs
[...]

Issue #522, opened 2026-09-19
Log report is flooded by relay debug lines: after a reconnect it covers ~2 seconds


grunch:
## Problem
The in-app log report (Settings → Log report) keeps the last **1000** entries (`BUFFER_CAPACITY`, `rust/src/api/logging.rs`). Every kind-14 frame a relay delivers is logged at DEBUG as

```
relay: raw ev=<id> kind=14 sub=<sub> relay=<url>
```

(`rust/src/api/orders.rs`, the `RelayMessage::Event` arm, ~l. 7463). A reconnect — or the kind-14 re-subscription after a restore — replays the whole DM history, once per relay, so these lines alone fill the buffer in seconds and evict everything else.

## Observed
A report shared right after a seed import (#519) spans **10:25:05 → 10:25:07**: ~1000 `raw ev=` lines from two relays and nothing from the `orders` / `restore` / `daemon-msg` subsystems. The lines needed to diagnose the import (`adopted range remainder order=…`, `rebuilt trade row from DM …`) were already gone. The *Orders* filter does not help: it filters the same evicted buffer.

## Expected
A shared report still carries the order/restore/payment history of the last minutes, whatever the relays replayed.

## Options
- Do not let `raw ev=` frames compete for the shared buffer: a separate small ring for relay traffic, or per-subsystem quotas.
- Or drop per-frame logging to TRACE (not buffered) and log one summary per EOSE (`sub=… relay=… events=N`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #524, opened 2026-09-19
fix(restore): settle replayed history instead of listing it as in progress

grunch:
## Problem
After #519 a seed import recovers correctly, but **My Trades lists old orders as in progress**: 7 range-sell remainders as *Published*, all dated "now".

The restore itself was right — the log shows `RestoreData … (0 orders, 3 disputes)`, and mostrod returns **every non-finished order, pending included** (`mostro/src/db.rs`, `EXCLUDED_ORDER_STATUSES`). The rows come from what follows: the kind-14 filter is replaced to cover all recovered keys (`p_count=97`), the relays replay the full history, and every old `new-order` is adopted as a Pending maker row (`adopt_range_remainder`), every old take rebuilt mid-trade (`rebuild_trade_from_dm`). Nothing settles them afterwards:
- a **seller** learns of completion only from the public Kind 38383 `success` (the daemon sends `PurchaseCompleted` to the buyer only);
- the stale sweep never examines `Pending` rows, and the book only carries pending + the last 48 h.

They also date themselves "now": the list shows `started_at`, which a rebuilt row set to the replay time.

## Fix
- **Restore snapshot** (`mostro::restore_history`): `restore_session` stores the resync floor and the order ids the daemon returned (orders + open disputes) in settings (`restore_snapshot`). A row whose trade index is **≤ floor and not listed** is history. Every trade started after the restore uses a higher index, so this needs no timestamps and also catches rows already stored wrong on a device.
- **History pass** (`reconcile_restored_history`): settles such rows against the order's public status (`fetch_public_order_status`, an unwindowed `d`-tag lookup):
  - `success` → the row becomes a completed trade;
  - any other ending (canceled, expired, or the book still showing it live while the daemon says otherwise) → wiped through `wipe_never_active_trade`, leaving the #394 tombstone so the next replay doesn't rebuild it;
  - no answer → left for the next pass.

  Runs 15 s and 60 s after a restore (the replay takes a while) and at the start of every stale sweep. It only rings `touch_trade` — filing history must not raise trade notices. Completed rows are protected from later replayed messages by the existing hard-terminal guard.
- **Date**: `trade_row_from_small_order` sets `started_at` from the order's `created_at`.

Not changed: the 3 open disputes are in the live set, so their replayed messages rebuild them normally.

Related: #522 (log buffer flooded by relay lines), #523 (one chat REQ per trade trips nos.lol) — found while diagnosing this; stuck history rows also kept chat REQs open.

## Tests
- `mostro::restore_history` (8): history = ≤ floor and not returned; returned or above-floor never; snapshot JSON round-trip; `success` → completed, other statuses → wipe, none → retry; which statuses still need settling.
- `restored_history_is_settled_against_the_public_status`: five rows on the real store — public `success` → Success; `canceled` → row gone + tombstone; no answer → untouched; a live-set dispute and an above-floor trade → untouched and **never looked up**.
- `a_rebuilt_row_starts_when_its_order_was_created`.
- `cargo test` 758 passed; clippy adds nothing (the one non-MutexGuard warning is in `db/sqlite.rs`, already on main).

## Test plan
- [x] `cd rust && cargo test && cargo clippy`
- [ ] Device: import the seed → within ~1 min the old *Published* rows disappear from *In progress*; completed ones show under history with their real dates; the 3 disputes remain

https://claude.ai/code/session_015RQArZ4KZ3x1QwgMVpYALB

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trade history recovery after restoring data, correctly distinguishing completed historical trades from trades still in progress.
  * Restored trades now retain their original order creation time instead of being dated at the time of replay.
  * Historical trades are settled according to their current order status, preventing ended trades from being incorrectly recreated.


grunch:
Re: [review 5255923088](https://github.com/MostroP2P/app/pull/524#pullrequestreview-5255923088) (nitpick on `reconcile_history_with`, lines 6493-6500) — Valid (the duplicate lookup) — fixed in 7f55b23, but not by skipping `SettledHoldInvoice` in the history pass: the sweep's payout branch only resolves that status on a public `success` (`sweep_action` keeps it otherwise), so a history row that ended any other way would read as in progress forever — the bug this PR fixes. Instead `reconcile_history_with` now returns the order ids it looked up, and `run_stale_sweep_once` skips them in its public-status branch for that pass (bond-expiry handling still runs for every row). The restore test asserts the looked-up set covers the three history rows and excludes the live dispute and the post-restore trade. Not applied: concurrent lookups — the pass runs in the background at most twice after a restore and then every 30 min, so sequential lookups cost latency no one waits on, while `join_all` over a large history would burst dozens of REQs at relays that already refuse ours for exceeding their limits (#523).

PR #525, opened 2026-09-19
fix(restore): retry a silent restore once and ignore replayed replies

grunch:
## Problem
Importing a seed sometimes ends with *"Account imported, but Mostro did not answer…"*; importing again works.

Log of a failing attempt (logs are cleared on import, so this is that attempt alone):
```
10:59:25 orders: daemon-message watcher already live for trade=ef9f72ca — re-arm refreshed its lease
10:59:25 restore: RestoreSession published trade_index=1 — waiting for daemon
10:59:25 publish … OK ×3 (nos.lol, shadowbip, mostro-p2p.tech)
10:59:32 relay: notice relay=wss://nos.lol msg=ERROR: too many concurrent REQs
10:59:35 publish … relay.mostro.network FAIL: timeout
10:59:43 relay: closed sub=… relay=wss://relay.mostro.network msg=Number of subscriptions exceeds limit
```
No `daemon-msg` line and no rejection: **no reply reached the app within the 10 s window** (`NoDaemonResponse`). Meanwhile relays refuse or close our subscriptions, and the app ignores `CLOSED` (only logs it, `orders.rs` `RelayMessage::Closed`), so the re-armed watcher is trusted as live where it may not exist. The root cause — subscription count and `CLOSED` handling — is #523.

A second, latent flaw found on the way: a restore is correlated **by trade pubkey only** (`take_matching_restore`), and after a re-import the counter restarts, so the restore's key (index 1) is the one every earlier import already restored with. The global kind-14 feed replays those old `RestoreData` / `CantDo` replies, and the first to arrive answered the new request — a stale snapshot or a false rejection.

## Fix
- **Retry once**: `recover_trades` runs `restore_session` through `retry_once_on_no_response`. Each attempt derives a fresh trade key and opens its own subscriptions — what the user was doing by hand. Any other error, or a second silence, is returned as is.
- **Age gate**: `PendingRequestKind::Restore { sent_at }`; `take_matching_restore(pubkey, reply_ts)` only accepts a reply no more than `RESTORE_REPLY_SKEW_SECS` (30 s) older than the request. Both call sites (the `RestoreSession` and `CantDo` arms) pass the event's `created_at`.

## Tests
- `take_matching_restore_ignores_replies_older_than_the_request`: a reply 31 s older than the request leaves the record; one exactly 30 s older (daemon clock behind) resolves it.
- `a_restore_without_reply_is_retried_once`, `a_restore_is_never_retried_twice_nor_after_a_real_error`.
- Existing restore-matching tests updated to the new signature.
- `cargo test` 751 passed; clippy adds nothing; FRB regeneration produces no diff.

## Test plan
- [x] `cd rust && cargo test && cargo clippy`
- [ ] Device: reinstall → import the seed a few times → no "Mostro did not answer" (or, if the relays refuse both attempts, the log shows `retrying once on a fresh trade key`)

https://claude.ai/code/session_015RQArZ4KZ3x1QwgMVpYALB

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session restoration by retrying once when no response is received.
  * Prevented outdated restoration responses from being incorrectly matched after a session re-import.
  * Improved restoration response matching by accounting for response timing.


grunch:
Re: [review 5258856802](https://github.com/MostroP2P/app/pull/525#pullrequestreview-5258856802) — valid, fixed in b1a7cf0. All seven paths that raise the marker (`create_order`, `take_order`, `send_invoice`, `restore_session`, the bond and dispute requests) now use `crate::mostro::pending::NO_DAEMON_RESPONSE` instead of the literal, matching how `INVALID_TRADE_INDEX` is referenced. Added `nothing_spells_the_marker_out_a_second_time`: it scans `rust/src` and fails on the literal outside the constant's module and the retry's own tests — verified failing with one literal put back. `cargo test` 752 passed.

